PR #7, opened 2024-06-12
perf: Dory's compute_T_vec_prime should be able to use Blitzar's fixed MSM (PROOF-876)

jacobtrombetta:
# Rationale for this change
When creating a Dory proof the `compute_T_vec_prime` function requires multiple MSMs. The current implementation of this function only supports the use of the Arkworks library, which only runs on the CPU. In an effort to improve the performance of Dory, this PR introduces the Blitzar fixed MSM to `compute_T_vec_prime` when appropriate to perform the MSM computations.

The commit focusing only on performance changes is available [here](https://github.com/spaceandtimelabs/sxt-proof-of-sql/commit/ba506b0b5c4be809ca424e07423c67f15fee6c15).

- `compute_T_vec_prime` now supports Blitzar's fixed MSM if the feature flag `"blitzar"` is turned on. The CPU version using Arkworks is still supported.
- The `transpose_column` and related functions appearing in the `dory_commitment_helper_gpu` class are moved into a separate class, `transpose`.
- Added a reference to the `blitzar_handle` to the `ProverSetup`.
  - Removed `par_iter()` in `eval_vmv_re.rs` because `*mut blitzar_sys::sxt_multiexp_handle` cannot be shared between threads safely.

# Are these changes tested?
Yes, on both CPU and GPU.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dory_vmv_helper.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dory_vmv_helper.rs
@@ -1,6 +1,13 @@
-use super::{G1Affine, G1Projective, ProverSetup, F};
+#[cfg(not(feature = "blitzar"))]
+use super::G1Projective;
+use super::{transpose, G1Affine, ProverSetup, F};
 use crate::base::polynomial::compute_evaluation_vector;
-use ark_ec::{AffineRepr, VariableBaseMSM};
+use ark_ec::AffineRepr;
+#[cfg(not(feature = "blitzar"))]
+use ark_ec::VariableBaseMSM;
+use ark_ff::{BigInt, MontBackend};
+#[cfg(feature = "blitzar")]
+use blitzar::compute::ElementP2;
 use num_traits::{One, Zero};
 
 /// Compute the evaluations of the columns of the matrix M that is derived from `a`.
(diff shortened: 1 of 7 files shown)


crates/proof-of-sql/src/proof_primitive/dory/dory_vmv_helper.rs
```diff
@@ -13,8 +20,56 @@ pub(super) fn compute_v_vec(a: &[F], L_vec: &[F], sigma: usize, nu: usize) -> Ve
...
+) -> Vec<G1Affine> {
+    let num_columns = 1 << sigma;
+    let data_size = std::mem::size_of::<F>();
+    let mut blitzar_commit = vec![ElementP2::<ark_bls12_381::g1::Config>::default(); 1];
+    let gs: Vec<_> = prover_setup.Gamma_1[nu]
+        .iter()
+        .copied()
+        .map(Into::into)
+        .collect();
+    let blitzar_handle = blitzar::compute::MsmHandle::new(&gs);
```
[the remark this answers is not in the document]
jacobtrombetta: Great idea. Updated.

crates/proof-of-sql/src/proof_primitive/dory/dory_vmv_helper.rs
```diff
@@ -13,8 +21,50 @@ pub(super) fn compute_v_vec(a: &[F], L_vec: &[F], sigma: usize, nu: usize) -> Ve
...
+    sigma: usize,
+    nu: usize,
+    prover_setup: &ProverSetup,
+) -> Vec<G1Affine> {
+    let num_columns = 1 << sigma;
+    let num_outputs = 1 << nu;
+    let data_size = std::mem::size_of::<F>();
+
+    let a_array = convert_scalar_to_padded_array(a, num_columns, num_outputs);
+    let a_transpose = transpose::transpose_for_fixed_msm(&a_array, 0, num_columns, data_size);
```
JayWhite2357, line 53: Does this work?
```suggestion
    let a_transpose = transpose::transpose_for_fixed_msm(&a, 0, num_columns, data_size);
```

jacobtrombetta: No, passing `a` does not work because the trait bound `ark_ff::Fp<MontBackend<ark_bls12_381::FrConfig, 4>, 4>: AsBytes` is not satisfied.

The `transpose_for_fixed_msm` will pad to get to to the closest multiple of the `num_columns` above `offset + a.len()`. It does not know about the `num_outputs`. In this case we need pre-pad the array by the `num_outputs` in order to correctly do the transpose.

It does seem like the logic should be inside the `transpose_for_fixed_msm` function. Then it can just consider `rows` and `cols`. I can update the `transpose_for_fixed_msm` to do the padding an additional `num_outputs` parameter.

crates/proof-of-sql/src/proof_primitive/dory/setup.rs
```diff
@@ -85,7 +85,7 @@ impl<'a> From<&'a PublicParameters> for ProverSetup<'a> {
             value.Gamma_2_fin,
             value.max_nu,
             #[cfg(feature = "blitzar")]
-            &value.blitzar_handle,
+            value.get_blitzar_handle(),
```
JayWhite2357, line 88: Well, I guess since this is needed, the handle can't be truly hidden from `super`.

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -72,4 +72,21 @@ impl PublicParameters {
             blitzar_handle,
         }
     }
+
+    /// Get a reference to the blitzar handle.
+    #[cfg(feature = "blitzar")]
+    pub fn get_blitzar_handle(&self) -> &MsmHandle<ElementP2<ark_bls12_381::g1::Config>> {
```
JayWhite2357, line 78: `get_` is not typical Rust
```suggestion
    pub(super) fn blitzar_handle(&self) -> &MsmHandle<ElementP2<ark_bls12_381::g1::Config>> {
```

JayWhite2357 (changes requested):

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
jacobtrombetta: Removed, thank you.

JayWhite2357 (approved):

PR #9, opened 2024-06-13
docs: prep rustdocs for docs.rs

JayWhite2357:
# Rationale for this change

When pushing to crates.io, we want the docs to look good.

* The README is embeded in the `proof-of-sql` lib docs.
    * The link in the README are turned into absolute, rather than relative, link.
* A link to the `proof-of-sql` crate is added to the `proof-of-sql-parser` crate.

# Are these changes tested?

Yes. The docs look good locally. It is possible that pushing to `docs.rs` will result in some other issues, but this is hard to test without pushing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ cargo run --example hello_world
 
 #### Output
 
-```
+```bash
 Warming up GPU... 520.959485ms
 Loading data... 3.229767ms
 Parsing Query... 1.870256ms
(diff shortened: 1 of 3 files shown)


iajoiner (approved):
Thanks @JayWhite2357 !

iajoiner:
Hmm please address the doctest issue. `#[cfg(not(doctest))]` should help.

iajoiner (approved):

PR #12, opened 2024-06-13
feat!: initial support for timestamp

drcapybara:
# Rationale for this change

Provides the initial typing support for a postgresql-like ```TimeStamp``` type. The proofs ```TimeStampTZ``` is typed over a custom ```TimeUnit``` and ```TimeZone``` type.

The choice is made to type out TimeStamp in this way because the arrow::datatypes::TimeStamp has a required TimeUnit field and an Option TimeZone  field. Typing out our own TimeStamp with these fields gives us greater control over the arrow type if we want. We can also jut default to seconds and UTC if needed, which would simplify this design a bit.

## Typing Rationale

The ```arrow::datatypes::timezone``` is typed over a ```TimeUnit``` and an optional timezone ```Option<Arc<str>>```. Thus in our application it makes sense to have a mapping of this metadata:

example:
```rust
// arrow datatype mapping to our new timestamp type
DataType::Timestamp(time_unit, timezone_option) => Ok(ColumnType::TimestampTZ(
    PoSQLTimeUnit::from(time_unit),
    PoSQLTimeZone::try_from(timezone_option)?,
)),
```
If this becomes burdensome, we could just as easily remove the timezone type and simply default to UTC, and handle any timezone conversion in DML and DDL. We will align with postgresql and store all times as UTC by default.

Finally, the ```PoSQLTimeUnit``` type gives us the flexibility to store times in either seconds, milliseconds, nanoseconds, or microseconds for high precision. This type maps directly to ```TimeUnit``` which we alias as ```ArrowTimeUnit``` in this PR.

## Typing updates:

- [x] Column
- [x] OwnedColumn
- [x] CommittableColumn
- [x] ColumnBounds
- [x] Typed TimeZone
- [x] Typed TimeUnit
- [x] ```impl ArrayRefExt for ArrayRef -> to_curve_25519_scalar & to_column```
- [x] LiteralValue
- [x] owned_and_arrow_conversions
- [x]  ```impl<S: Scalar> FromIterator<i64> for OwnedColumn<S>```
- [x] ```impl<CP: CommitmentEvaluationProof> DataAccessor<CP::Scalar> for OwnedTableTestAccessor<CP>```
- [x] owned_table_utility
- [x] test accessor_utility
- [x] multi-linear-extension
- [x] Scalar trait bounds
- [x] compute_dory_commitment
- [x] filter_column_by_index
- [x] prover_evaluate
- [x] sum_aggregate_column_by_index_counts
- [x] compare_indexes_by_columns
- [x] impl ProvableQueryResult
- [x] to_owned_table
- [x] trait ProvableResultColumn 
- [x] make_empty_query_result
- [x] record_batch_dataframe_conversion
- [x] impl ToArrow for RecordBatch 

# Are these changes tested?

## Tests:

- [x] TimeUnit Conversions
- [x] ColumnBounds
- [x] ColumnCommitmentMetadata
- [x] arrow_array_to_column_conversion
- [x] column
- [x] owned_table 

# Split:

- lalrpop grammar update and token parsing
- timestamp.now()
- timestamp.current_time()

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -3,6 +3,7 @@ use crate::base::{
     math::decimal::Precision,
     ref_into::RefInto,
     scalar::Scalar,
+    time::timestamp::{ProofsTimeUnit, ProofsTimeZone},
 };
 #[cfg(feature = "blitzar")]
 use blitzar::sequence::Sequence;
@@ -38,7 +39,7 @@ pub enum CommittableColumn<'a> {
     /// Column of limbs for committing to scalars, hashed from a VarChar column.
     VarChar(Vec<[u64; 4]>),
     /// Borrowed Timestamp column, mapped to `u64`.
-    Timestamp(&'a [u64]),
+    Timestamp(ProofsTimeUnit, ProofsTimeZone, &'a [i64]),
 }
 
 impl<'a> CommittableColumn<'a> {
@@ -53,7 +54,7 @@ impl<'a> CommittableColumn<'a> {
             CommittableColumn::Scalar(col) => col.len(),
             CommittableColumn::VarChar(col) => col.len(),
             CommittableColumn::Boolean(col) => col.len(),
-            CommittableColumn::Timestamp(col) => col.len(),
+            CommittableColumn::Timestamp(_, _, col) => col.len(),
         }
     }
 
@@ -81,7 +82,7 @@ impl<'a> From<&CommittableColumn<'a>> for ColumnType {
             CommittableColumn::Scalar(_) => ColumnType::Scalar,
             CommittableColumn::VarChar(_) => ColumnType::VarChar,
             CommittableColumn::Boolean(_) => ColumnType::Boolean,
-            CommittableColumn::Timestamp(_) => ColumnType::Timestamp,
+            CommittableColumn::Timestamp(tu, tz, _) => ColumnType::Timestamp(*tu, *tz),
         }
     }
 }
@@ -103,7 +104,7 @@ impl<'a, S: Scalar> From<&Column<'a, S>> for CommittableColumn<'a> {
                 let as_limbs: Vec<_> = scalars.iter().map(RefInto::<[u64; 4]>::ref_into).collect();
                 CommittableColumn::VarChar(as_limbs)
             }
-            Column::Timestamp(times) => CommittableColumn::Timestamp(times),
+            Column::Timestamp(tu, tz, times) => CommittableColumn::Timestamp(*tu, *tz, times),
         }
     }
 }
@@ -133,7 +134,7 @@ impl<'a, S: Scalar> From<&'a OwnedColumn<S>> for CommittableColumn<'a> {
                     .map(Into::<[u64; 4]>::into)
                     .collect(),
             ),
-            OwnedColumn::Timestamp(times) => (times as &[_]).into(),
+            OwnedColumn::Timestamp(_, _, times) => (times as &[_]).into(),
         }
     }
 }
@@ -168,11 +169,6 @@ impl<'a> From<&'a [bool]> for CommittableColumn<'a> {
         CommittableColumn::Boolean(value)
     }
 }
-impl<'a> From<&'a [u64]> for CommittableColumn<'a> {
-    fn from(value: &'a [u64]) -> Self {
-        CommittableColumn::Timestamp(value)
-    }
-}
 
 #[cfg(feature = "blitzar")]
 impl<'a, 'b> From<&'a CommittableColumn<'b>> for Sequence<'a> {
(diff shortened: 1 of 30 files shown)


crates/proof-of-sql/src/base/database/owned_column.rs
```diff
@@ -85,6 +94,7 @@ impl<S: Scalar> FromIterator<i32> for OwnedColumn<S> {
         Self::Int(Vec::from_iter(iter))
     }
 }
+// TODO: does this conflict with TimeStamp?
```
drcapybara, line 97: Should we have a custom impl for timestamp here?

crates/proof-of-sql/src/base/database/column.rs
```diff
@@ -274,6 +298,22 @@ impl TryFrom<DataType> for ColumnType {
...
+            DataType::Timestamp(time_unit, timezone_option) => {
+                let custom_time_unit = ProofsTimeUnit::from(time_unit);
+
+                let timezone = match timezone_option {
+                    Some(tz_arc) => {
+                        let tz_str = &*tz_arc; // Dereference Arc<str> to &str
+                        chrono_tz::Tz::from_str(tz_str)
+                            .map_err(|_| format!("Invalid timezone string: {}", tz_str))?
+                    }
+                    None => chrono_tz::Tz::UTC, // Default to UTC if None
```
drcapybara, line 310: this aligns with postgreSQL

crates/proof-of-sql/src/base/database/column.rs
```diff
@@ -206,15 +228,15 @@ impl ColumnType {
                 | ColumnType::BigInt
                 | ColumnType::Int128
                 | ColumnType::Scalar
-                | ColumnType::Decimal75(_, _)
+                | ColumnType::Decimal75(_, _) // TODO: is a timestamp numeric?
```
iajoiner, line 231: Can you add timestamps? If not no.

crates/proof-of-sql/src/base/database/column.rs
```diff
@@ -206,15 +228,15 @@ impl ColumnType {
...
         )
     }
 
     /// Returns true if this column is an integer and false otherwise
     pub fn is_integer(&self) -> bool {
         matches!(
             self,
-            ColumnType::SmallInt | ColumnType::Int | ColumnType::BigInt | ColumnType::Int128
+            ColumnType::SmallInt | ColumnType::Int | ColumnType::BigInt | ColumnType::Int128 // TODO: is a timestamp an integer?
```
iajoiner, line 239: No since I'm not sure whether it makes any sense to do integer operation on it. If we add Duration type though it can become useful since Timestamp + Duration and Timestamp - Duration have meanings.

JayWhite2357 (changes requested):
Looks mostly good to me.
This should be marked as a breaking change since it will require an update of upstream crates.

crates/proof-of-sql/src/base/database/record_batch_dataframe_conversion.rs
```diff
@@ -81,6 +83,42 @@ pub fn record_batch_to_dataframe(record_batch: RecordBatch) -> Option<DataFrame>
                         // Note: we make this unchecked because if record batch has values that overflow 38 digits, so should the data frame.
                         .into_series()
                 }
+                arrow::datatypes::DataType::Timestamp(time_unit, _timezone_option) => {
```
JayWhite2357, line 86: NIT: it's probably a bit cleaner to do something like this:
```rust
arrow::datatypes::DataType::Timestamp(arrow::datatypes::TimeUnit::Second, _) => {
    let data = col
                                .as_any()
                                .downcast_ref::<arrow::array::TimestampSecondArray>()
                                .map(|array| array.values())
                                .unwrap();
                            Series::new(f.name(), data)
},
arrow::datatypes::DataType::Timestamp(arrow::datatypes::TimeUnit::Millisecond, _) => {
    let data = col
                                .as_any()
                                .downcast_ref::<arrow::array::TimestampMillisecondArray>()
                                .map(|array| array.values())
                                .unwrap();
                            Series::new(f.name(), data)
},
```
Sometimes nested matches are more efficient, but each branch is duplicated anyway, so in this case, unnesting them is probably a bit more concise.

You could do this in a bunch of different places in this PR.

JayWhite2357 (approved):

crates/proof-of-sql/src/base/commitment/committable_column.rs
```diff
@@ -208,6 +220,31 @@ mod tests {
         assert_eq!(res_committable_column, test_committable_column)
     }
 
+    #[test]
+    fn we_can_get_type_and_length_of_timestamp_column() {
+        // empty case
+        let smallint_committable_column =
```
iajoiner, line 226: Naming

crates/proof-of-sql/src/sql/ast/group_by_util.rs
```diff
@@ -114,6 +114,9 @@ pub(super) fn sum_aggregate_column_by_index_counts<'a, S: Scalar>(
         }
         Column::Scalar(col) => sum_aggregate_slice_by_index_counts(alloc, col, counts, indexes),
         Column::VarChar(_) => unimplemented!("Cannot sum varchar columns"),
+        Column::TimestampTZ(_, _, col) => {
+            sum_aggregate_slice_by_index_counts(alloc, col, counts, indexes)
+        }
```
iajoiner, line 119: We can't sum timestamps

drcapybara: what about for Decimal75?

iajoiner (commented):
Can we add some tests in `query_expr_tests.rs` and `integration_tests.rs` as well?

PR #14, opened 2024-06-14
feat: add `AddSubtractExpr`

iajoiner:
# Rationale for this change
We need to add support for arithmetic. Let's start from + and -.

- Add support for + and - in `proof-of-sql` crate
- generalize `scale_and_subtract_eval` to `scale_and_add_subtract_eval`
- make sure we pass in scales in any operation that can involve decimals

# Are these changes tested?
Yes.

### DIFF
--- a/crates/proof-of-sql/src/base/database/column.rs
+++ b/crates/proof-of-sql/src/base/database/column.rs
@@ -207,15 +207,15 @@ pub enum ColumnType {
     /// Mapped to String
     #[serde(alias = "VARCHAR", alias = "varchar")]
     VarChar,
-    /// Mapped to Curve25519Scalar
-    #[serde(alias = "SCALAR", alias = "scalar")]
-    Scalar,
     /// Mapped to i256
     #[serde(rename = "Decimal75", alias = "DECIMAL75", alias = "decimal75")]
     Decimal75(Precision, i8),
     /// Mapped to i64
     #[serde(alias = "TIMESTAMP", alias = "timestamp")]
     TimestampTZ(PoSQLTimeUnit, PoSQLTimeZone),
+    /// Mapped to Curve25519Scalar
+    #[serde(alias = "SCALAR", alias = "scalar")]
+    Scalar,
 }
 
 impl ColumnType {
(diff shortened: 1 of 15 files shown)


JayWhite2357 (commented):
Please add some tests that deal with overflow. I'm concerned that it isn't handled properly.

crates/proof-of-sql/src/base/database/column.rs
```diff
@@ -240,6 +240,44 @@ impl ColumnType {
         )
     }
 
+    /// Returns the number of bits in the integer type if it is an integer type. Otherwise, return None.
+    pub fn to_integer_bits(&self) -> Option<usize> {
```
JayWhite2357, line 244: Not a big fan of this function. At the very least it should be private.

My preference would be for the `max_integer_type` function to just have a (admittedly tedious) match.
```suggestion
    fn to_integer_bits(&self) -> Option<usize> {
```

crates/proof-of-sql/src/sql/ast/numerical_util.rs
```diff
@@ -0,0 +1,98 @@
...
+    lhs: Column<'a, S>,
+    rhs: Column<'a, S>,
+    alloc: &'a Bump,
+    is_subtract: bool,
+) -> ConversionResult<Column<'a, S>> {
+    let lhs_len = lhs.len();
+    let rhs_len = rhs.len();
+    if lhs_len != rhs_len {
+        return Err(ConversionError::DifferentColumnLength(lhs_len, rhs_len));
+    }
```
JayWhite2357, line 65: I think this can safely be an `assert!`, in which case this doesn't need to return a `Result`.

The main reason is that this method should never fail since it is within the prover.

crates/proof-of-sql/src/sql/ast/numerical_util.rs
```diff
@@ -0,0 +1,98 @@
...
+/// The counterpart of `try_add_subtract_columns` for evaluating decimal expressions.
+pub(crate) fn scale_and_add_subtract_eval<S: Scalar>(
+    lhs_eval: S,
+    rhs_eval: S,
+    lhs_scale: i8,
+    rhs_scale: i8,
+    is_subtract: bool,
+) -> ConversionResult<S> {
+    let max_scale = lhs_scale.max(rhs_scale);
+    let scaled_lhs_eval = scale_scalar(lhs_eval, max_scale - lhs_scale)?;
```
JayWhite2357, line 91: We should be able to remove the `Result` from this function as well.
```suggestion
    let scaled_lhs_eval = scale_scalar(lhs_eval, max_scale - lhs_scale).expect("scaling factor should not be negative");
```

crates/proof-of-sql/src/sql/ast/add_subtract_expr_test.rs
JayWhite2357, line 1: I would like to see extensive overflow tests in this file.
Something to the effect of `SELECT A+B, A-B FROM table` where table looks like this:
A | B
---|---
MAX | MAX
MIN | MIN
MAX | MIN
MIN | MAX

JayWhite2357 (changes requested):

crates/proof-of-sql/src/base/database/column.rs
iajoiner: I just made it private. I do think the match is not going to be great since we plan to add even more integer types.

JayWhite2357:
See https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/20 for how I think we should have the tests look going forward.

crates/proof-of-sql/src/sql/ast/provable_expr_plan.rs
```diff
@@ -34,6 +35,10 @@ pub enum ProvableExprPlan<C: Commitment> {
     Equals(EqualsExpr<C>),
     /// Provable AST expression for an inequality expression
     Inequality(InequalityExpr<C>),
+    /// Provable numeric `+` expression
+    Add(AddSubtractExpr<C>),
+    /// Provable numeric `-` expression
+    Subtract(AddSubtractExpr<C>),
```
JayWhite2357, line 41: NIT:
I don't think this needs to be separate variants.
If you want separate variants, I would suggest making `is_subtract` a const generic instead and then putting the appropriate const in the variant.
As it is, it's possible to have an Add variant that acts as a Subtract.

JayWhite2357 (approved):
Looks good. One NIT.

JayWhite2357 (dismissed):

JayWhite2357 (approved):

PR #16, opened 2024-06-19
fix!: separate element decoding error from verification error

iajoiner:
# Rationale for this change
For arithmetic (such as #14) to work it is necessary to separate element decoding error from verification error. There are already `QueryError::Overflow` and `QueryError::InvalidString` that handle some decoding errors. However currently in practice if overflows do take place we get `VerificationError`. This PR will fix this issue.

- add `QueryError::MiscellaneousDecodingError`, `QueryError::InvalidIndexes` and `QueryError::MiscellaneousEvaluationError` to account for decoding errors beyond overflows and invalid strings
- if decoding errors happen we will use them instead of `VerificationError`
- add overflow test in `crates/proof-of-sql/src/sql/proof/provable_query_result_test.rs`

# Are these changes tested?
Yes.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof/provable_query_result.rs
+++ b/crates/proof-of-sql/src/sql/proof/provable_query_result.rs
@@ -83,11 +83,11 @@ impl ProvableQueryResult {
         evaluation_point: &[S],
         table_length: usize,
         column_result_fields: &[ColumnField],
-    ) -> Option<Vec<S>> {
+    ) -> Result<Vec<S>, QueryError> {
         assert_eq!(self.num_columns as usize, column_result_fields.len());
 
         if !self.indexes.valid(table_length) {
-            return None;
+            return Err(QueryError::InvalidIndexes);
         }
 
         let evaluation_vec_len = self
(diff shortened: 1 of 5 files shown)


crates/proof-of-sql/src/sql/proof/provable_query_result.rs
```diff
@@ -83,7 +83,7 @@ impl ProvableQueryResult {
         evaluation_point: &[S],
         table_length: usize,
         column_result_fields: &[ColumnField],
-    ) -> Option<Vec<S>> {
+    ) -> Option<Result<Vec<S>, QueryError>> {
```
JayWhite2357, line 86: No reason to have it be both `Result` and `Option`
```suggestion
    ) -> Result<Vec<S>, QueryError> {
```

JayWhite2357 (changes requested):

drcapybara (approved):
LGTM - not sure if this is a breaking change or not with the switch to Result from Option

iajoiner:
@drcapybara You are right! It is a breaking change since it changes error messages when certain non-verification errors occur.

JayWhite2357 (approved):
LGTM

PR #18, opened 2024-06-20
feat!: add `MultiplyExpr`

iajoiner:
# Rationale for this change
We need to add support for arithmetic. Now let's continue with `*`.

- Add support for `*` in `proof-of-sql` crate

# Are these changes tested?
Yes.

### DIFF
--- a/crates/proof-of-sql/src/sql/ast/multiply_expr_test.rs
+++ b/crates/proof-of-sql/src/sql/ast/multiply_expr_test.rs
@@ -1,0 +1,346 @@
+use crate::{
+    base::{
+        commitment::InnerProductProof,
+        database::{owned_table_utility::*, Column, OwnedTableTestAccessor},
+        scalar::Curve25519Scalar,
+    },
+    sql::{
+        ast::{test_utility::*, ProofPlan, ProvableExpr, ProvableExprPlan},
+        parse::ConversionError,
+        proof::{exercise_verification, QueryError, VerifiableQueryResult},
+    },
+};
+use bumpalo::Bump;
+use curve25519_dalek::ristretto::RistrettoPoint;
+use itertools::{multizip, MultiUnzip};
+use rand::{
+    distributions::{Distribution, Uniform},
+    rngs::StdRng,
+};
+use rand_core::SeedableRng;
+
+// select a * 2 as a, c, b * 4.5 as b, d * 3  + 4.7 as d, e from sxt.t where d * 3.9 = 8.19
+#[test]
+fn we_can_prove_a_typical_multiply_query() {
+    let data = owned_table([
+        smallint("a", [1_i16, 2, 3, 4]),
+        int("b", [0_i32, 1, 2, 1]),
+        varchar("e", ["ab", "t", "efg", "g"]),
+        bigint("c", [0_i64, 2, 2, 0]),
+        decimal75("d", 2, 1, [21_i64, 4, 21, -7]),
+    ]);
+    let t = "sxt.t".parse().unwrap();
+    let accessor = OwnedTableTestAccessor::<InnerProductProof>::new_from_table(t, data, 0, ());
+    let ast = dense_filter(
+        vec![
+            (
+                multiply(column(t, "a", &accessor), const_int(2)),
+                "a".parse().unwrap(),
+            ),
+            col_expr_plan(t, "c", &accessor),
+            (
+                multiply(column(t, "b", &accessor), const_decimal75(2, 1, 45)),
+                "b".parse().unwrap(),
+            ),
+            (
+                add(
+                    multiply(column(t, "d", &accessor), const_smallint(3)),
+                    const_decimal75(2, 1, 47),
+                ),
+                "d".parse().unwrap(),
+            ),
+            col_expr_plan(t, "e", &accessor),
+        ],
+        tab(t),
+        equal(
+            multiply(column(t, "d", &accessor), const_decimal75(2, 1, 39)),
+            const_decimal75(3, 2, 819),
+        ),
+    );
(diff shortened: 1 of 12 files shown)


crates/proof-of-sql/src/sql/ast/multiply_expr_test.rs
```diff
@@ -0,0 +1,344 @@
...
+    let expected_res = owned_table([bigint("a", [i64::MAX]), smallint("b", [2_i16])]);
+    assert_eq!(res, expected_res);
+}
+
+// select a * b as c from sxt.t
+#[test]
+fn result_expr_can_overflow_more() {
+    let data = owned_table([
+        bigint("a", [i64::MAX, i64::MIN, i64::MAX, i64::MIN]),
+        bigint("b", [i64::MAX, i64::MAX, i64::MIN, i64::MIN]),
```
drcapybara, line 169: this is nice

drcapybara, line 169: I'll make sure to have tests for boundary conditions like this for new features as well

crates/proof-of-sql/tests/integration_tests.rs
```diff
@@ -323,7 +323,7 @@ fn we_can_prove_a_complex_query_with_curve25519() {
         "sxt.table".parse().unwrap(),
         owned_table([
             smallint("a", [1_i16, 2, 3]),
-            int("b", [1_i32, 0, 1]),
+            int("b", [1_i32, 4, 3]),
             bigint("c", [3_i64, 3, -3]),
             bigint("d", [1_i64, 2, 3]),
             varchar("e", ["d", "e", "f"]),
```
drcapybara, line 329: could we test boundaries here as well

iajoiner: @drcapybara I mean we can but I think we don't need to. The purpose of this test is to check a bunch of random features in a single complex query.

drcapybara (approved):
LGTM - Up to you if you want to enhance tests here with boundary conditions. We might need the blitzar upgrade first.

crates/proof-of-sql/src/sql/parse/provable_expr_plan_builder.rs
```diff
@@ -144,9 +144,15 @@ impl ProvableExprPlanBuilder<'_> {
...
-            BinaryOperator::Multiply | BinaryOperator::Division => Err(
-                ConversionError::Unprovable(format!("Binary operator {:?} is not supported", op)),
-            ),
+            BinaryOperator::Multiply => {
+                let left = self.visit_expr(left);
+                let right = self.visit_expr(right);
+                ProvableExprPlan::try_new_multiply(left?, right?)
+            }
+            BinaryOperator::Division => Err(ConversionError::Unprovable(format!(
+                "Binary operator {:?} is not supported",
```
JayWhite2357, line 153:
```suggestion
                "Binary operator {:?} is not supported at this location",
```

crates/proof-of-sql/src/sql/parse/query_context_builder.rs
```diff
@@ -313,9 +313,8 @@ pub(crate) fn type_check_binary_operation(
         BinaryOperator::Add | BinaryOperator::Subtract => {
             try_add_subtract_column_types(*left_dtype, *right_dtype).is_ok()
         }
-        BinaryOperator::Multiply | BinaryOperator::Division => {
-            left_dtype.is_numeric() && right_dtype.is_numeric()
-        }
+        BinaryOperator::Multiply => try_multiply_column_types(*left_dtype, *right_dtype).is_ok(),
+        BinaryOperator::Division => left_dtype.is_numeric() && right_dtype.is_numeric(),
```
JayWhite2357, line 317: I believe this is a breaking change.
Did `SELECT a*b from table` work before when `a` and `b` were both `decimal(38,0)`?
Does it work now?
To be clear, I support this change, but it should be marked as breaking if it is.

JayWhite2357 (changes requested):

crates/proof-of-sql/src/sql/parse/query_context_builder.rs
iajoiner: It would have worked in polars. It doesn't work now. Yup.

crates/proof-of-sql/src/sql/ast/numerical_util.rs
```diff
@@ -82,6 +121,23 @@ pub(crate) fn add_subtract_columns<'a, S: Scalar>(
...
+    alloc: &'a Bump,
+) -> &'a [S] {
+    let lhs_len = lhs.len();
+    let rhs_len = rhs.len();
+    assert!(
+        lhs_len == rhs_len,
+        "lhs and rhs should have the same length"
+    );
+    alloc.alloc_slice_fill_with(lhs_len, |i| {
+        lhs.scalar_at(i).unwrap() * rhs.scalar_at(i).unwrap()
```
JayWhite2357, line 137: NOTE: This is fine for now, but I think we need to revise things like this eventually.
I don't like blanket converting to scalars here. Aside from type safety, which is perhaps unavoidable, this is an efficiency concern.
For example:
```
let a = i64::MIN
let b = i64::MAX
let prod = S::from(a) * S::from(b)
```
is less efficient than
```
let a = i64::MIN
let b = i64::MAX
let prod = S::from((a as i128) * (b as i128))
```

This certainly would lead to larger `match` branches, but in general, I like the philosophy of "only convert to Scalar when necessary".

This is a very minor performance change in this case, so we can do the optimization later, but it is something that might eventually give very real performance gains with some of the optimizations I have in mind.

JayWhite2357 (approved):

PR #19, opened 2024-06-20
fix: add check for group by result uniqueness

JayWhite2357:
# Rationale for this change

The `GroupByExpr` needs to have the result checked for uniqueness of the groups in order to prevent a malicious prover from returning a result that is only partially aggregated.

* `ProofExpr::verifier_evaluate` now accepts an optional result as a parameter.
* `GroupByExpr` uses this result to check that the groups are strictly increasing. This guarantees uniqueness.

# Are these changes tested?

Yes.

### DIFF
--- a/crates/proof-of-sql/src/sql/ast/group_by_expr.rs
+++ b/crates/proof-of-sql/src/sql/ast/group_by_expr.rs
@@ -1,13 +1,15 @@
 use super::{
-    aggregate_columns, fold_columns, fold_vals, group_by_util::AggregatedColumns,
-    provable_expr_plan::ProvableExprPlan, ColumnExpr, ProvableExpr, TableExpr,
+    aggregate_columns, fold_columns, fold_vals,
+    group_by_util::{compare_indexes_by_owned_columns, AggregatedColumns},
+    provable_expr_plan::ProvableExprPlan,
+    ColumnExpr, ProvableExpr, TableExpr,
 };
 use crate::{
     base::{
         commitment::Commitment,
         database::{
             Column, ColumnField, ColumnRef, ColumnType, CommitmentAccessor, DataAccessor,
-            MetadataAccessor,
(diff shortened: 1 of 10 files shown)


crates/proof-of-sql/src/sql/ast/group_by_util.rs
```diff
@@ -183,3 +186,29 @@ pub(super) fn compare_indexes_by_columns<S: Scalar>(
         .find(|&ord| ord != Ordering::Equal)
         .unwrap_or(Ordering::Equal)
 }
+
+/// Compares the tuples (group_by[0][i], group_by[1][i], ...) and
+/// (group_by[0][j], group_by[1][j], ...) in lexicographic order.
+///
+/// Identical in functionality to [compare_indexes_by_columns]
+pub(super) fn compare_indexes_by_owned_columns<S: Scalar>(
```
iajoiner, line 194: Not that we need this now but I wonder whether we should have a trait for common functionalities of different kinds of columns.

crates/proof-of-sql/src/sql/ast/group_by_expr.rs
```diff
@@ -141,11 +144,28 @@ impl<C: Commitment> ProofExpr<C> for GroupByExpr<C> {
...
+        match result {
+            Some(table) => {
+                let cols = self
+                    .group_by_exprs
+                    .iter()
+                    .map(|col| table.inner_table().get(&col.column_id()))
+                    .collect::<Option<Vec<_>>>()
+                    .ok_or(ProofError::VerificationError(
+                        "Result does not all correct group by columns.",
+                    ))?;
```
iajoiner, line 157: Not that we need to fix this now but we really need to split out `VerificationError` into numerous variants based on what the actual error is.

iajoiner (approved):
Thanks @JayWhite2357 ! We can discuss the 2 comments later.

PR #20, opened 2024-06-20
refactor: update and expr tests

JayWhite2357:
# Rationale for this change

Many of our ast node tests are outdated and somewhat inflexible. This PR improves test ergonomics and makes the `AndExpr` test more explicit. Other nodes tests will follow in later PRs.

* `filter_test_and_expr` (which relies on `polars`) is replaced by a more flexible/explicit `multizip().filter_map().multiunzip()` chain. This is made more ergonomic by:
    * `"type"_iter` methods are added to `OwnedColumn`.
    * `impl Index<&str> for OwnedTable` is added.
* `OwnedTableTestAccessor::new_from_table` is added for convenience.

Finally, `and_expr_test.rs` is refactored (using the above changes) such that:
* `polars` and `arrow` are unused.
* Each test is self-contained.
* DenseFilter is used instead of Filter.

While the net result is +23 LOC in `and_expr_test.rs`, I think this extra code is worth it because it simplifies the tests, even though there is a bit more duplicated code.

# Are these changes tested?

Yes

### DIFF
--- a/crates/proof-of-sql/src/base/database/mod.rs
+++ b/crates/proof-of-sql/src/base/database/mod.rs
@@ -38,10 +38,7 @@ mod record_batch_test_accessor_test;
 #[cfg(any(test, feature = "test"))]
 mod test_accessor_utility;
 #[cfg(any(test, feature = "test"))]
-pub use test_accessor_utility::{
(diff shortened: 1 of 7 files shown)


iajoiner (approved):
Thanks @JayWhite2357 !

PR #23, opened 2024-06-21
fix: make `QueryCommitments::from_accessor_with_max_bounds` mirror the underlying accessor schema

JayWhite2357:
# Rationale for this change

A bug was uncovered that resulted in because the prover and verifier were using schema accessors with different orderings of the columns when using a wildcard query.

* `QueryCommitments::from_accessor_with_max_bounds` now mirrors the underlying accessor schema.

# Are these changes tested?

Yes


iajoiner:
Hmm sorting of columns in a table should be positional, not alphabetical.

iajoiner (approved):
Thanks @JayWhite2357 ! It will be great to have a test with wildcards though.

PR #24, opened 2024-06-21
ci: add semantic-release

dalbertom:
Since we want to use 0.x.y versions for now, we modify the behavior via
releaseRules, e.g.
* breaking is set to minor release (by default it's major)
* feat is set to patch (by default it's minor)

Once we release v1.0.0 we should reset these to their intended behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,56 @@
+{
+    "name": "proof_of_sql",
+    "version": "0.0.0-development",
+    "devDependencies": {
+        "conventional-changelog-conventionalcommits": "^5.0.0",
+        "semantic-release": "^21.0.5"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -0,0 +1,72 @@
...
+                        },
+                        {
+                            "path": "dist/*.so*"
+                        },
+                        {
+                            "path": "dist/*.zip"
+                        },
+                        {
+                            "path": "dist/*.tar.gz"
+                        }
```
JayWhite2357, line 63: I'm not clear on what this is for.

JayWhite2357 (approved):

PR #26, opened 2024-06-24
refactor: conversion error decimal variant

drcapybara:
# Rationale for this change

The ```ConversionError``` error type has grown quite large with the addition of new features and types. This PR categorizes some timestamp and decimal errors into different modules and reduces the size and complexity of ```ConversionError```.

IntermediateDecimal, Decimal, IntermediateTimeStamp, Timestamp errors are moved to respective modules. Native thiserr ```#from``` is used in place of manual ```From``` conversions.

# Are these changes tested?

yes

### DIFF
--- a/crates/proof-of-sql/src/base/math/decimal.rs
+++ b/crates/proof-of-sql/src/base/math/decimal.rs
@@ -5,6 +5,27 @@ use crate::{
 };
 use proof_of_sql_parser::intermediate_decimal::IntermediateDecimal;
 use serde::{Deserialize, Deserializer, Serialize};
+use thiserror::Error;
+
+/// Errors related to decimal operations.
+#[derive(Error, Debug, Eq, PartialEq)]
+pub enum DecimalError {
+    #[error("Invalid decimal format or value: {0}")]
+    /// Error when a decimal format or value is incorrect
+    InvalidDecimal(String),
+
+    #[error("Unsupported operation: cannot round decimal: {0}")]
+    /// Decimal rounding is not supported
+    DecimalRoundingError(String),
+
+    #[error("Error while parsing precision from query: {0}")]
+    /// Error in parsing precision in a query
+    PrecisionParseError(String),
+
+    #[error("Decimal precision is not valid: {0}")]
+    /// Decimal precision exceeds the allowed limit
+    InvalidPrecision(String),
+}
 
 #[derive(Eq, PartialEq, Debug, Clone, Hash, Serialize, Copy)]
 /// limit-enforced precision
@@ -15,9 +36,11 @@ impl Precision {
     /// Constructor for creating a Precision instance
     pub fn new(value: u8) -> Result<Self, ConversionError> {
         if value > MAX_SUPPORTED_PRECISION || value == 0 {
-            Err(ConversionError::PrecisionParseError(format!(
-                "Failed to parse precision. Value of {} exceeds max supported precision of {}",
-                value, MAX_SUPPORTED_PRECISION
+            Err(ConversionError::Decimal(DecimalError::PrecisionParseError(
+                format!(
+                    "Failed to parse precision. Value of {} exceeds max supported precision of {}",
+                    value, MAX_SUPPORTED_PRECISION
+                ),
             )))
         } else {
             Ok(Precision(value))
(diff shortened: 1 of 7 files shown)


crates/proof-of-sql/src/base/math/decimal.rs
```diff
@@ -1,10 +1,37 @@
...
+    InvalidPrecision(String),
+
+    #[error("Error while parsing precision from query: {0}")]
+    /// Error in parsing precision in a query
+    PrecisionParseError(String),
+
+    #[error("Unsupported operation: cannot round decimal: {0}")]
+    /// Decimal rounding is not supported
+    RoundingError(String),
+}
```
iajoiner, line 34: What exactly are the use cases of each variant?
I think there are several classes of decimal errors here:
1. The string isn't even a decimal e.g. "notastring", "-21.233.122" etc aka `InvalidDecimal`
2. Invalid precision e.g. precision above 75/76/whatever set by `Scalar` or non-positive aka `InvalidPrecision`
3. Invalid scale e.g. operations that cause scale to be outside i8 range aka `InvalidScale`
Unless explicit rounding happens what we consider to be `RoundingError` is in reality an `InvalidPrecision` since in order not to round the precision will hit the upper bound.

iajoiner (changes requested):
Please address the issue above

crates/proof-of-sql/src/sql/parse/error.rs
```diff
@@ -50,50 +50,25 @@ pub enum ConversionError {
...
-    /// Error in parsing precision in a query
-    PrecisionParseError(String),
-
-    #[error("Decimal precision is not valid: {0}")]
-    /// Decimal precision is an integer but exceeds the allowed limit. We use i16 here to include all kinds of invalid precision values.
-    InvalidPrecision(i16),
-
     #[error("Encountered parsing error: {0}")]
     /// General parsing error
     ParseError(String),
```
iajoiner, line 55: Why does this still exist?

crates/proof-of-sql/src/sql/parse/error.rs
```diff
@@ -50,50 +50,25 @@ pub enum ConversionError {
...
 
+    #[error(transparent)]
+    /// Errors related to decimal operations
+    DecimalConversionError(#[from] DecimalError),
+
+    #[error(transparent)]
+    /// Errors related to decimal operations
+    IntermediateDecimalConversionError(#[from] IntermediateDecimalError),
```
iajoiner, line 63: Since these are different we need to separate them..or maybe `IntermediateDecimalError` should be included in `DecimalError`?

crates/proof-of-sql/src/sql/parse/error.rs
```diff
@@ -50,50 +50,25 @@ pub enum ConversionError {
...
+
     #[error("Unsupported operation: cannot round literal: {0}")]
     /// Error when a rounding operation is not supported
     LiteralRoundDownError(String),
```
iajoiner, line 67: Why does this still exist?

crates/proof-of-sql/src/base/math/decimal.rs
```diff
@@ -1,10 +1,41 @@
...
+    /// e.g. precision above 75/76/whatever set by Scalar
+    /// or non-positive aka InvalidPrecision
+    InvalidPrecision(String),
+
+    #[error("Unsupported operation: cannot round decimal: {0}")]
+    /// Unless explicit rounding happens, what we consider to be
+    /// RoundingError is in reality an InvalidPrecision since in
+    /// order not to round the precision will hit the upper bound.
+    /// Regardless, this error occurs when attempting to scale a
+    /// decimal in such a way that a loss of precision occurs.
```
iajoiner, line 36: Hmm if that's a separate enum variant we need a different description

crates/proof-of-sql/src/base/math/decimal.rs
```diff
@@ -1,10 +1,51 @@
...
+    /// into a posql decimal
+    #[error("Intermediate decimal conversion error: {0}")]
+    IntermediateDecimalConversionError(IntermediateDecimalError),
+}
+
+impl From<IntermediateDecimalError> for ConversionError {
+    fn from(err: IntermediateDecimalError) -> ConversionError {
+        DecimalConversionError(IntermediateDecimalConversionError(err))
+    }
+}
```
iajoiner, line 48: Please use #[from]

drcapybara: we cannot do that without putting the variant back into ```ConversionError```

iajoiner (approved):
Thanks @drcapybara !

PR #32, opened 2024-06-27
feat: add aggregation-related intermediate AST conversions

iajoiner:
# Rationale for this change
This is the first PR in generalization of `GroupByExpr`. Following conventions from [datafusion](https://docs.rs/datafusion/latest/datafusion/logical_expr/enum.Expr.html) we allow `AggregateExpr` so that provable aggregations can be expressed in `ProvableExprPlan` and provide relevant conversion.

- add `AggregateExpr` which allows expressions such as `sum(a+b)` to become `ProvableExprPlan`
- allow `Expression` -> `ProvableExprPlan` for aggregations

# Are these changes tested?
Existing tests should pass.

### DIFF
--- a/crates/proof-of-sql/src/sql/ast/aggregate_expr.rs
+++ b/crates/proof-of-sql/src/sql/ast/aggregate_expr.rs
@@ -1,0 +1,75 @@
+use super::{ProvableExpr, ProvableExprPlan};
+use crate::{
+    base::{
+        commitment::Commitment,
+        database::{Column, ColumnRef, ColumnType, CommitmentAccessor, DataAccessor},
+        proof::ProofError,
+    },
+    sql::proof::{CountBuilder, ProofBuilder, VerificationBuilder},
+};
+use bumpalo::Bump;
+use proof_of_sql_parser::intermediate_ast::AggregationOperator;
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+
+/// Provable aggregate expression
+///
+/// Currently it doesn't do much since aggregation logic is implemented elsewhere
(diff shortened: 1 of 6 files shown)


crates/proof-of-sql/src/sql/ast/aggregate_expr.rs
```diff
@@ -0,0 +1,78 @@
...
+};
+use bumpalo::Bump;
+use serde::{Deserialize, Serialize};
+use std::collections::HashSet;
+
+/// Provable aggregate expression
+///
+/// Currently it doesn't do much since aggregation logic is implemented elsewhere
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct AggregateExpr<C: Commitment> {
```
JayWhite2357, line 22: Might be nice to name this to indicate that it doesn't actually do the aggregation. I don't really like doing `...able` and I can't think of anything better, so maybe just keep it as it is.
```suggestion
pub struct AggregatableExpr<C: Commitment> {
```

JayWhite2357 (commented):
Looks pretty good.

crates/proof-of-sql/src/sql/ast/aggregate_expr.rs
iajoiner: @JayWhite2357 Datafusion names it [AggregateFunction](https://docs.rs/datafusion/latest/datafusion/logical_expr/expr/struct.AggregateFunction.html) which is also a bit confusing in its own way.

JayWhite2357: Up to you what you want to name it.

JayWhite2357 (approved):

PR #34, opened 2024-06-27
ci: fix dry-run

JayWhite2357:
# Rationale for this change

Semantic release is failing because `cargo publish -p proof-of-sql --dry-run` fails since it needs `proof-of-sql-parser`, which doesn't publish because it is a dry run.

Commented out the dry-run for `proof-of-sql` publishing.

# Are these changes tested?

No


iajoiner (commented):
I can approve this if necessary but why does this fail?

JayWhite2357:
> I can approve this if necessary but why does this fail?

Updated the description to explain it more.

iajoiner (approved):
Approved because for `proof-of-sql` to be published even with dry-run it is necessary to have `proof-of-sql-parser` published already

PR #36, opened 2024-06-27
feat: generalize `sum_expr` in `GroupByExpr`

iajoiner:
# Rationale for this change
This PR generalizes `GroupByExpr` by relaxing the requirement of `expr` in `SUM(expr)` to arbitrary `ProvableExprPlan` outside aggregation.

Generalize `sum_expr` in `GroupByExpr`

# Are these changes tested?
Old tests all pass. We need to get new ones with nontrivial `sum_expr` to pass as well.

### DIFF
--- a/crates/proof-of-sql/src/sql/ast/aggregate_expr.rs
+++ b/crates/proof-of-sql/src/sql/ast/aggregate_expr.rs
@@ -29,8 +29,8 @@ impl<C: Commitment> AggregateExpr<C> {
 }
 
 impl<C: Commitment> ProvableExpr<C> for AggregateExpr<C> {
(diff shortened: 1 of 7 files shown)


JayWhite2357 (commented):
Looks good. I believe changing the following should fix the failing tests
https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/24b3592c0a89311817ca86a2cec6758e219de1dd/crates/proof-of-sql/src/sql/ast/aggregate_expr.rs#L32-L33
It should be
```rust
    fn count(&self, builder: &mut CountBuilder) -> Result<(), ProofError> {
        self.expr.count(builder)
```

JayWhite2357 (approved):

JayWhite2357 (approved):

PR #40, opened 2024-07-01
feat: generalize `count_exprs` and `result_exprs` in `GroupByExpr`

iajoiner:
# Rationale for this change
This PR further generalizes `GroupByExpr`.
Note that we should do a final generalization that allows queries such as `select a + b + count(c) + sum(d) as res from tab where f > 3 group by a, b`

1. There can be multiple or none `count` in a provable GROUP BY clause
2. Result expressions can be arbitrary combinations of columns in the GROUP BY clause.

# Are these changes tested?
Yes


iajoiner:
Closed in favor of composing `GroupByExpr` with `DenseFilterExpr` or `ProjectionExpr`

PR #41, opened 2024-07-02
feat: add `ProjectionExpr`

iajoiner:
# Rationale for this change
We need to add `ProjectionExpr` both as a faster and simpler alternative to `DenseFilterExpr` when there is no `WHERE` clause and to provide one mechanism to further generalize `GROUP BY` i.e. adding a `ProjectionExpr` after `GroupByExpr` without changing `GroupByExpr` itself. 

Add `ProjectionExpr`.

# Are these changes tested?
Yes.

### DIFF
--- a/crates/proof-of-sql/src/sql/ast/projection_expr_test.rs
+++ b/crates/proof-of-sql/src/sql/ast/projection_expr_test.rs
@@ -1,0 +1,350 @@
+use crate::{
+    base::{
+        database::{
+            owned_table_utility::*, ColumnField, ColumnRef, ColumnType, OwnedTable,
+            OwnedTableTestAccessor, TableRef, TestAccessor,
+        },
+        math::decimal::Precision,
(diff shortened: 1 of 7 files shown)


crates/proof-of-sql/src/sql/ast/projection_expr_test.rs
```diff
@@ -0,0 +1,357 @@
+use crate::{
+    base::{
+        database::{
+            owned_table_utility::*, ColumnField, ColumnRef, ColumnType, OwnedTable,
+            OwnedTableTestAccessor, RecordBatchTestAccessor, TableRef, TestAccessor,
```
JayWhite2357, line 5: NIT: Let's used `OwnedTableTestAccessor` instead of `RecordBatchTestAccessor` and avoid any arrow within these tests.

JayWhite2357 (approved):

JayWhite2357 (approved):

PR #45, opened 2024-07-10
feat: time unit try from string

drcapybara:
# Rationale for this change

In DDL, a Timestamp column can be instantiated with the following syntax:

```sql
TIMESTAMP(6)
```

where 6 represents millisecond precision. It appears that the gateway by default supports only nanosecond precision on DDL, however if this changes in the future, we will need a way to parse the time unit from a DDL message in proofs-service. This PR contains a simple string parser that allows for selection of a supported precision for timestamp columns. 

- [x] Implements TryFrom &str for TimeUnit
- [x] updates Display implementation for TImestamp column type to fix a backwards print that is occuring:

```
column type TIMESTAMP(TIMEUNIT: UTC, TIMEZONE: NANOSECONDS (PRECISION: 9)) 
``` 

# Are these changes tested?

yes

### DIFF
--- a/crates/proof-of-sql-parser/src/posql_time/unit.rs
+++ b/crates/proof-of-sql-parser/src/posql_time/unit.rs
@@ -1,3 +1,4 @@
+use crate::error::PoSQLTimestampError;
 use arrow::datatypes::TimeUnit as ArrowTimeUnit;
 use core::fmt;
 use serde::{Deserialize, Serialize};
@@ -13,6 +14,19 @@ pub enum PoSQLTimeUnit {
     Microsecond,
     /// Represents nanoseconds with precision 9: ex "2024-06-20 12:34:56.123456789"
     Nanosecond,
+}
+
+impl TryFrom<&str> for PoSQLTimeUnit {
+    type Error = PoSQLTimestampError;
+    fn try_from(value: &str) -> Result<Self, PoSQLTimestampError> {
+        match value {
+            "0" => Ok(PoSQLTimeUnit::Second),
+            "3" => Ok(PoSQLTimeUnit::Millisecond),
+            "6" => Ok(PoSQLTimeUnit::Microsecond),
+            "9" => Ok(PoSQLTimeUnit::Nanosecond),
+            _ => Err(PoSQLTimestampError::UnsupportedPrecision(value.into())),
+        }
+    }
 }
 
 impl From<PoSQLTimeUnit> for ArrowTimeUnit {
@@ -54,8 +68,24 @@ impl fmt::Display for PoSQLTimeUnit {
 #[allow(deprecated)]
 mod time_unit_tests {
 
-    use crate::posql_time::{timestamp::PoSQLTimestamp, unit::PoSQLTimeUnit};
+    use crate::{
+        error::PoSQLTimestampError,
+        posql_time::{timestamp::PoSQLTimestamp, unit::PoSQLTimeUnit},
+    };
     use chrono::{TimeZone, Utc};
+
+    #[test]
+    fn test_invalid_precision() {
+        let invalid_precisions = ["1", "2", "4", "5", "7", "8", "10"]; // Testing various invalid inputs
+        for &value in invalid_precisions.iter() {
+            let result = PoSQLTimeUnit::try_from(value);
+            assert!(
+                matches!(result, Err(PoSQLTimestampError::UnsupportedPrecision(err)) if err == *value),
+                "Failed with value: {}",
+                value
+            );
+        }
+    }
 
     #[test]
     fn test_rfc3339_timestamp_with_milliseconds() {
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql-parser/src/posql_time/unit.rs
```diff
@@ -53,10 +67,75 @@ impl fmt::Display for PoSQLTimeUnit {
...
+        assert!(
+            matches!(PoSQLTimeUnit::try_from("10"), Err(PoSQLTimestampError::UnsupportedPrecision(x)) if x == "10")
+        );
+        // Non-numeric strings
+        assert!(
+            matches!(PoSQLTimeUnit::try_from("abc"), Err(PoSQLTimestampError::UnsupportedPrecision(x)) if x == "abc")
+        );
+        // Empty string
+        assert!(
+            matches!(PoSQLTimeUnit::try_from(""), Err(PoSQLTimestampError::UnsupportedPrecision(x)) if x.is_empty())
```
iajoiner, line 112: Please try one negative number as well

iajoiner:
This mostly looks good though we don't have any feature allowing it to be used though.

crates/proof-of-sql-parser/src/posql_time/unit.rs
```diff
@@ -53,10 +67,78 @@ impl fmt::Display for PoSQLTimeUnit {
...
+        assert_eq!(PoSQLTimeUnit::try_from("6"), Ok(PoSQLTimeUnit::Microsecond));
+        assert_eq!(PoSQLTimeUnit::try_from("9"), Ok(PoSQLTimeUnit::Nanosecond));
+    }
+
+    #[test]
+    fn test_invalid_precisions() {
+        // Test some random incorrect values
+        assert!(
+            matches!(PoSQLTimeUnit::try_from("-1"), Err(PoSQLTimestampError::UnsupportedPrecision(x)) if x == "-1")
+        );
```
iajoiner, line 87: You don't really need the `if x == "-1"` clause. Usually we do not do assert_eq on error since error message changes shouldn't normally trigger test changes. However since `x` in `PoSQLTimestampError::UnsupportedPrecision(x)` is entirely dependent on the invalid precision we may as well have `assert_eq!` instead.

crates/proof-of-sql-parser/src/posql_time/unit.rs
```diff
@@ -53,10 +67,78 @@ impl fmt::Display for PoSQLTimeUnit {
...
+        assert!(
+            matches!(PoSQLTimeUnit::try_from("zero"), Err(PoSQLTimestampError::UnsupportedPrecision(x)) if x == "zero")
+        );
+        assert!(
+            matches!(PoSQLTimeUnit::try_from("three"), Err(PoSQLTimestampError::UnsupportedPrecision(x)) if x == "three")
+        );
+    }
+
+    #[test]
+    fn test_invalid_precision() {
```
iajoiner, line 130: Hmm are these three tests really testing the exact same failures? If so we only need the last one with more invalid inputs.

drcapybara:
> This mostly looks good though we don't have any feature allowing it to be used though.

I need it now for string parsing in sxt-db

crates/proof-of-sql-parser/src/posql_time/unit.rs
drcapybara: you asked me to add a negative number case below

crates/proof-of-sql-parser/src/posql_time/unit.rs
drcapybara: I did but your feedback above seems to conflict with this

crates/proof-of-sql-parser/src/posql_time/unit.rs
iajoiner: We don't need multiple tests for the same failures.
`test_invalid_precisions`
`test_edge_cases_with_non_numeric_inputs`
`test_invalid_precision`
should be consolidated into one test since they all test for invalid precisions

iajoiner:
```rust
#[test]
    fn test_invalid_precision() {
drcapybara marked this conversation as resolved.
        let invalid_precisions = ["1", "2", "4", "5", "7", "8", "10", "zero", "three", "cat", "-1"]; // Testing various invalid inputs
        for &value in invalid_precisions.iter() {
            let result = PoSQLTimeUnit::try_from(value);
            assert!(
                matches!(result, Err(PoSQLTimestampError::UnsupportedPrecision(_)))
            );
        }
    }
```

crates/proof-of-sql-parser/src/posql_time/unit.rs
iajoiner: Yes. Please see below.

crates/proof-of-sql-parser/src/posql_time/unit.rs
drcapybara: understood, thanks

iajoiner (approved):

PR #47, opened 2024-07-11
bench: add Jaeger benchmarks for Dory proof evaluation (PROOF-892)

jacobtrombetta:
# Rationale for this change
In order to run benchmarks easily on Dory proof evaluation performance, this PR adds a new benchmark type to `jaeger_benches`. An argument is added to the command line interface to define the benchmark type. 

Previously, the Jaeger benchmarks were only run on the inner product proof with the command:
`cargo bench -p proof-of-sql --bench jaeger_benches`
which has been updated to
`cargo bench -p proof-of-sql --bench jaeger_benches InnerProductProof`.

The Dory proof evaluation benchmark can be run with
`cargo bench -p proof-of-sql --bench jaeger_benches Dory features="test"`.

- A Jaeger benchmark is added for Dory.
- A command line argument is added to specify which benchmark to run: `InnerProductProof` or `Dory --features="test"`

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/benches/README.md
+++ b/crates/proof-of-sql/benches/README.md
@@ -8,9 +8,10 @@ To run benchmarks with Jaeger, you need to do the following
     ```bash
     docker run --rm -d --name jaeger -p 6831:6831/udp -p 16686:16686 jaegertracing/all-in-one:latest
     ```
-2. Run the benchmark.
+2. Run a benchmark.
     ```bash
-    cargo bench -p proof-of-sql --bench jaeger_benches
+    cargo bench -p proof-of-sql --bench jaeger_benches InnerProductProof
+    cargo bench -p proof-of-sql --bench jaeger_benches Dory --features="test"
     ```
 3. Navigate to http://localhost:16686/ to see the results.
 4. To end the Jaeger service, run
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql/benches/README.md
```diff
@@ -8,9 +8,10 @@ To run benchmarks with Jaeger, you need to do the following
     ```bash
     docker run --rm -d --name jaeger -p 6831:6831/udp -p 16686:16686 jaegertracing/all-in-one:latest
     ```
-2. Run the benchmark.
+2. Run a benchmark.
     ```bash
-    cargo bench -p proof-of-sql --bench jaeger_benches
+    cargo bench -p proof-of-sql --bench jaeger_benches InnerProductProof
+    cargo bench -p proof-of-sql --bench jaeger_benches Dory --features="test"
```
iajoiner, line 14: Why the discrepancy?

jacobtrombetta: Dory's public parameters are generated with `proof_of_sql::proof_primitive::dory::test_rng()`. This function is defined in the `rand_util` module which is only added with the `--features="test"` flag.

Perhaps there is a more optimal way to do this?

iajoiner (approved):

crates/proof-of-sql/benches/README.md
iajoiner: @jacobtrombetta I wonder whether we can remove https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/crates/proof-of-sql/src/proof_primitive/dory/mod.rs#L30 though. Unless there is some good reason why the entire module is behind that flag.

jacobtrombetta: Good idea, I'll check with @JayWhite2357 to see if he has any hesitations removing that flag.

PR #50, opened 2024-07-15
fix: fix postprocessing projection in provable `GROUP BY`s & allow `WHERE` clause to be omitted

iajoiner:
# Rationale for this change
Recently we found weird postprocessing bugs for group bys. What we found is that
1. Currently provable `GROUP BY`s without a `WHERE` clause are not supported even though to support it is trivial.
2. Currently if a `GROUP BY` clause is provable the `ResultExpr` still incorrectly contains aggregate functions.

- fix postprocessing projection with provable `GROUP BY`
- allow provable group by without `WHERE` clause
- enable 3 ignored tests

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/parse/query_context.rs
+++ b/crates/proof-of-sql/src/sql/parse/query_context.rs
@@ -224,10 +224,6 @@ impl<C: Commitment> TryFrom<&QueryContext> for Option<GroupByExpr<C>> {
     type Error = ConversionError;
 
     fn try_from(value: &QueryContext) -> Result<Option<GroupByExpr<C>>, Self::Error> {
-        // Currently if there is no where clause, we can't prove the query
-        if value.where_expr.is_none() {
-            return Ok(None);
-        }
         let where_clause = WhereExprBuilder::new(&value.column_mapping)
             .build(value.where_expr.clone())?
             .unwrap_or_else(|| ProvableExprPlan::new_literal(LiteralValue::Boolean(true)));
@@ -283,6 +279,7 @@ impl<C: Commitment> TryFrom<&QueryContext> for Option<GroupByExpr<C>> {
                     false
                 }
             });
+
         // Check sums
         let sum_expr = sum_expr_columns
             .iter()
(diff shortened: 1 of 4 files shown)


crates/proof-of-sql/src/sql/parse/query_context.rs
```diff
@@ -224,10 +224,6 @@ impl<C: Commitment> TryFrom<&QueryContext> for Option<GroupByExpr<C>> {
     type Error = ConversionError;
 
     fn try_from(value: &QueryContext) -> Result<Option<GroupByExpr<C>>, Self::Error> {
-        // Currently if there is no where clause, we can't prove the query
-        if value.where_expr.is_none() {
-            return Ok(None);
-        }
```
tlovell-sxt, line 230: I'm a bit confused by this removal, why was this code here in the first place - why are we removing it? Could you add an explanation to the PR description

crates/proof-of-sql/src/sql/parse/query_expr.rs
```diff
@@ -62,10 +65,20 @@ impl<C: Commitment> QueryExpr<C> {
         let group_by = context.get_group_by_exprs();
         if !group_by.is_empty() {
             if let Some(group_by_expr) = Option::<GroupByExpr<C>>::try_from(&context)? {
+                // If the group by expression is provable the projection step is just identity.
```
tlovell-sxt, line 68: I do worry about this a little bit - we're assuming this is the case because of the `Some` result of the `GroupByExpr::try_from(context)` right? But will this implication hold true in the future? Is it possible for the group by expression to change in the future to allow for queries w/ aliases?

It would be nice if, in the event of such a change, something about this code would let us know we need to update it too, like a compilation failure or test failure. Maybe this should have a `assert!(result_aliased_exprs.iter().all())` to check for this guarantee. Not necessary if we don't expect this assumption to change though, I suppose

crates/proof-of-sql/src/sql/parse/query_context.rs
iajoiner: @tlovell-sxt Sure!

crates/proof-of-sql/src/sql/parse/query_expr.rs
iajoiner: We already allow queries with aliases. However in the future we will have possible filters and projections after group bys. So yes I will add the assertion.

tlovell-sxt (approved):
LGTM, thanks for the new tests and improved description

PR #51, opened 2024-07-16
fix: fix decimal comparison and aggregation

iajoiner:
# Rationale for this change
It is necessary for us to be able to do comparison and aggregation on decimal columns.

- add `signed_cmp` to `Scalar` definition to do signed Scalar comparison (i.e. `S::MAX_SIGNED` is the largest number and everything "beyond" it wraps around into negative numbers)
- allow decimal comparison
- allow `SUM` of decimals

# Are these changes tested?
Yes.

### DIFF
--- a/crates/proof-of-sql/src/base/scalar/mod.rs
+++ b/crates/proof-of-sql/src/base/scalar/mod.rs
@@ -5,7 +5,7 @@ mod mont_scalar;
 #[cfg(test)]
 mod mont_scalar_test;
 use crate::sql::parse::ConversionError;
-use core::ops::Sub;
+use core::{cmp::Ordering, ops::Sub};
 pub use mont_scalar::Curve25519Scalar;
 pub(crate) use mont_scalar::MontScalar;
 mod mont_scalar_from;
@@ -79,6 +79,14 @@ pub trait Scalar:
     const ONE: Self;
     /// 1 + 1
     const TWO: Self;
+    /// Compare two `Scalar`s as signed numbers.
+    fn signed_cmp(&self, other: &Self) -> Ordering {
+        match *self - *other {
+            x if x.is_zero() => Ordering::Equal,
+            x if x > Self::MAX_SIGNED => Ordering::Less,
+            _ => Ordering::Greater,
+        }
+    }
 }
 
 macro_rules! scalar_conversion_to_int {
@@ -207,4 +215,5 @@ macro_rules! scalar_conversion_to_int {
         }
     };
 }
+
 pub(crate) use scalar_conversion_to_int;
--- a/crates/proof-of-sql/src/sql/ast/group_by_util.rs
+++ b/crates/proof-of-sql/src/sql/ast/group_by_util.rs
@@ -107,18 +107,17 @@ pub(super) fn sum_aggregate_column_by_index_counts<'a, S: Scalar>(
     indexes: &[usize],
 ) -> &'a [S] {
     match column {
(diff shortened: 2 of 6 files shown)


crates/proof-of-sql/src/base/scalar/mod.rs
```diff
@@ -20,6 +20,16 @@ mod commitment_utility;
 pub use commitment_utility::compute_commitment_for_testing;
 use num_bigint::BigInt;
 
+/// Alternative ordering on finite sets considered to be signed.
+pub trait SignedOrd: Ord {
```
tlovell-sxt, line 24: I'm trying to think if a new trait is the right way to go here, or if it's a new type. For example

```rust
pub struct SignedScalar<S>(S) where S: Scalar;

// implement various num_traits and operations for `SignedScalar`, including `Ord`

// use `SignedScalar` in Decimal75 columns instead of plain Scalar
```

I think something like that is a bigger lift, but I think it has a couple benefits. Namely,
1. The new type makes it very clear what the meaning of the underlying data is, whereas a new trait is sort of agnostic about this (it could be either)
2. Allows for using the `SignedScalar` in generic code w/ trait bounds like `T: Ord` without unexpected results (the type is very clear about which order implementation is the right one)

The drawbacks would be, again, it's a bigger lift. Another one I can think of is that it might be hard to understand which one is the correct one to use elsewhere in the code base, or as we develop new code. I think for this purpose it would be good if `SignedScalar` didn't actually implement `Scalar`, so that we don't make the mistake of using it in the proving protocols, but it could implement `Into<S> where S: Scalar`. I don't know off the top of my head if this thought is compatible with all of our current code.

iajoiner: I think the essence of the problem is that any `Scalar` just like bytes and vectors of bytes have both signed and unsigned forms of comparison.

crates/proof-of-sql/src/sql/ast/group_by_util.rs
```diff
@@ -107,19 +107,16 @@ pub(super) fn sum_aggregate_column_by_index_counts<'a, S: Scalar>(
...
+        Column::Decimal75(_, _, col) => {
+            sum_aggregate_slice_by_index_counts(alloc, col, counts, indexes)
         }
         Column::Scalar(col) => sum_aggregate_slice_by_index_counts(alloc, col, counts, indexes),
-        Column::VarChar(_) => panic!("Cannot sum varchar columns"),
-        Column::TimestampTZ(_, _, _) => {
-            panic!("Cannot sum varchar columns")
-        }
+        // The following should never be reached because the `SUM` function can only be applied to numeric types.
+        _ => unreachable!("SUM can not be applied to non-numeric types"),
```
tlovell-sxt, line 119: Could you actually lay out the cases for this? Just so when we add new types it'll cause a compiler error until we update this code?

```suggestion
        Column::VarChar(_) | Column::TimestampTZ(_) | Column::Boolean(_) => unreachable!("SUM can not be applied to non-numeric types"),
```

tlovell-sxt (commented):
Just one thing from me

crates/proof-of-sql/src/sql/ast/group_by_util.rs
iajoiner: Sure!

tlovell-sxt (approved):
LGTM, idk if you want an additional review from Jay or anything

tlovell-sxt (approved):

PR #52, opened 2024-07-17
feat: timestamp inequality support

drcapybara:
# Rationale for this change

Quick PR to add support for inequality expressions between timestamps. Since theyre represented internally as i64, they can be easily compared against one another.

Adds support for statements with the following syntax:

```
SELECT * FROM table WHERE times <= timestamp '1970-01-01T00:00:00Z
SELECT * FROM table WHERE times >= timestamp '1970-01-01T00:00:00Z
SELECT * FROM table WHERE times > timestamp '1970-01-01T00:00:00Z
SELECT * FROM table WHERE times < timestamp '1970-01-01T00:00:00Z
```

# Are these changes tested?

yes

### DIFF
--- a/crates/proof-of-sql/tests/timestamp_integration_tests.rs
+++ b/crates/proof-of-sql/tests/timestamp_integration_tests.rs
@@ -374,7 +374,7 @@ mod tests {
 
 #[test]
 #[cfg(feature = "blitzar")]
-fn we_can_prove_a_query_with_arithmetic_in_where_clause_with_curve25519() {
+fn we_can_prove_timestamp_inequality_queries_with_multiple_columns() {
     use curve25519_dalek::RistrettoPoint;
     use proof_of_sql::base::{database::OwnedTable, scalar::Curve25519Scalar};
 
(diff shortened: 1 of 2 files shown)


iajoiner (commented):
Hmm where did you actually compare timestamps? Moreover how do you do it with different timezones?

crates/proof-of-sql/tests/timestamp_integration_tests.rs
```diff
@@ -134,6 +134,35 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_basic_timestamp_inequality_query() {
+        let test_timestamps = vec![i64::MIN, -1, 0, 1, i64::MAX];
+
+        run_timestamp_query_test(
+            "SELECT * FROM table WHERE times < timestamp '1970-01-01T00:00:00Z';",
```
iajoiner, line 142: We need filters in tests more complex than `column < literal`. Please at least try to compare two columns in one of the tests.

drcapybara:
> Hmm where did you actually compare timestamps? Moreover how do you do it with different timezones?

I would point you to this test [here](https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/80ae847d94711ddbd298d8ff50d3a5ae6bfbe029/crates/proof-of-sql/tests/timestamp_integration_tests.rs#L259-L286) which is probably the most significant example, this test confirms that our approach matches that of postgres.

I have also added many other examples in this PR which test inequalities on various timezones as you pointed out coverage was lacking in those areas

crates/proof-of-sql/tests/timestamp_integration_tests.rs
```diff
@@ -134,6 +134,86 @@ mod tests {
...
+            test_timestamps.clone(),
+            vec![i64::MIN, -1, 0],
+        );
+    }
+
+    #[test]
+    fn test_timestamp_inequality_queries_with_timezone_offsets() {
+        // Test with a range of timestamps around the Unix epoch
+        // 60 * 60 = 3600 * 8 (PST offset) = 28800
+        let test_timestamps = vec![i64::MIN, -28800, -28799, 0, 1, i64::MAX];
```
iajoiner, line 170: Maybe you want to use 28800 and 28799 instead so that the filter actually tells the difference between values?

crates/proof-of-sql/tests/timestamp_integration_tests.rs
```diff
@@ -342,3 +422,99 @@ mod tests {
...
+    accessor.add_table(
+        "sxt.table".parse().unwrap(),
+        owned_table([
+            timestamptz(
+                "start_time",
+                PoSQLTimeUnit::Nanosecond,
+                PoSQLTimeZone::Utc,
+                [
+                    0, // Beginning of Unix time
+                    444_972_800_000_000_000,
```
iajoiner, line 442: I'd make another row non-selected so you may as well change just one row a little bit

iajoiner:
BTW it will be really great if you squash your commits
@dalbertom

iajoiner (approved):
Really thanks for submitting this!

PR #56, opened 2024-07-23
feat: add `PostprocessingEvaluator`

iajoiner:
# Rationale for this change
For unprovable expressions it is necessary to evaluate them in the verifier. Hence it is necessary to add an evaluation framework to evaluate any `Expression`.

- add `PostprocessingEvaluator`
- add division utils

# Are these changes tested?
Pending

### DIFF
--- a/crates/proof-of-sql/src/base/database/owned_column.rs
+++ b/crates/proof-of-sql/src/base/database/owned_column.rs
@@ -56,6 +56,60 @@ impl<S: Scalar> OwnedColumn<S> {
         }
     }
 
+    /// Returns element at index as scalar
+    ///
+    /// Note that if index is out of bounds, this function will return None
+    pub(crate) fn scalar_at(&self, index: usize) -> Option<S> {
+        (index < self.len()).then_some(match self {
+            Self::Boolean(col) => S::from(col[index]),
+            Self::SmallInt(col) => S::from(col[index]),
+            Self::Int(col) => S::from(col[index]),
+            Self::BigInt(col) => S::from(col[index]),
+            Self::Int128(col) => S::from(col[index]),
+            Self::Scalar(col) => col[index],
+            Self::Decimal75(_, _, col) => col[index],
+            Self::VarChar(col) => S::from(col[index].clone()),
+            Self::TimestampTZ(_, _, col) => S::from(col[index]),
+        })
+    }
+
+    /// Cast a column of scalars to a numeric or timestamp column
+    pub(crate) fn from_scalars(scalars: Vec<S>, column_type: ColumnType) -> Option<Self> {
+        match column_type {
+            ColumnType::SmallInt => scalars
+                .iter()
+                .map(|s| -> Option<i16> { (*s).try_into().ok() })
+                .collect::<Option<Vec<i16>>>()
+                .map(OwnedColumn::SmallInt),
+            ColumnType::Int => scalars
+                .iter()
+                .map(|s| -> Option<i32> { (*s).try_into().ok() })
+                .collect::<Option<Vec<i32>>>()
+                .map(OwnedColumn::Int),
+            ColumnType::BigInt => scalars
+                .iter()
+                .map(|s| -> Option<i64> { (*s).try_into().ok() })
+                .collect::<Option<Vec<i64>>>()
+                .map(OwnedColumn::BigInt),
+            ColumnType::Int128 => scalars
+                .iter()
+                .map(|s| -> Option<i128> { (*s).try_into().ok() })
+                .collect::<Option<Vec<i128>>>()
+                .map(OwnedColumn::Int128),
+            ColumnType::Scalar => Some(OwnedColumn::Scalar(scalars.to_vec())),
+            ColumnType::Decimal75(precision, scale) => {
+                Some(OwnedColumn::Decimal75(precision, scale, scalars.to_vec()))
+            }
+            ColumnType::TimestampTZ(tu, tz) => scalars
+                .iter()
+                .map(|s| -> Option<i64> { (*s).try_into().ok() })
+                .collect::<Option<Vec<i64>>>()
+                .map(|s| OwnedColumn::TimestampTZ(tu, tz, s)),
+            // Boolean and VarChar columns are not supported by this function
+            _ => None,
+        }
+    }
+
     /// Returns the column with its entries permutated
     pub fn try_permute(&self, permutation: &Permutation) -> Result<Self, PermutationError> {
         Ok(match self {
@@ -108,6 +162,15 @@ impl<S: Scalar> OwnedColumn<S> {
             OwnedColumn::TimestampTZ(_, _, col) => col.is_empty(),
         }
     }
+
+    /// Returns the column as a slice of booleans if it is a boolean column. Otherwise, returns None.
+    pub(crate) fn as_boolean(&self) -> Option<&[bool]> {
+        match self {
+            Self::Boolean(col) => Some(col),
+            _ => None,
+        }
+    }
+
     /// Returns the type of the column.
     pub fn column_type(&self) -> ColumnType {
         match self {
@@ -209,7 +272,7 @@ pub(crate) fn compare_indexes_by_owned_columns_with_direction<S: Scalar>(
                 OwnedColumn::Int(col) => col[i].cmp(&col[j]),
                 OwnedColumn::BigInt(col) => col[i].cmp(&col[j]),
                 OwnedColumn::Int128(col) => col[i].cmp(&col[j]),
-                OwnedColumn::Decimal75(_, _, col) => col[i].cmp(&col[j]),
+                OwnedColumn::Decimal75(_, _, col) => col[i].signed_cmp(&col[j]),
                 OwnedColumn::Scalar(col) => col[i].cmp(&col[j]),
                 OwnedColumn::VarChar(col) => col[i].cmp(&col[j]),
                 OwnedColumn::TimestampTZ(_, _, col) => col[i].cmp(&col[j]),
--- a/crates/proof-of-sql/src/base/database/scalar_and_i256_conversions.rs
+++ b/crates/proof-of-sql/src/base/database/scalar_and_i256_conversions.rs
@@ -14,7 +14,7 @@ const MAX_SUPPORTED_I256: i256 = i256::from_parts(
 pub fn convert_scalar_to_i256<S: Scalar>(val: &S) -> i256 {
     let is_negative = val > &S::MAX_SIGNED;
     let abs_scalar = if is_negative { -*val } else { *val };
-    let limbs = abs_scalar.into();
+    let limbs: [u64; 4] = abs_scalar.into();
 
     let low = (limbs[0] as u128) | ((limbs[1] as u128) << 64);
     let high = (limbs[2] as i128) | ((limbs[3] as i128) << 64);
--- a/crates/proof-of-sql/src/sql/ast/numerical_util.rs
+++ b/crates/proof-of-sql/src/sql/ast/numerical_util.rs
@@ -1,12 +1,17 @@
 use crate::{
     base::{
-        database::{Column, ColumnType},
+        database::{Column, ColumnType, OwnedColumn},
         math::decimal::{scale_scalar, DecimalError, Precision},
         scalar::Scalar,
     },
-    sql::parse::{ConversionError, ConversionError::DecimalConversionError, ConversionResult},
+    sql::{
+        parse::{ConversionError, ConversionError::DecimalConversionError, ConversionResult},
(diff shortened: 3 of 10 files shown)


crates/proof-of-sql/src/base/database/owned_column.rs
```diff
@@ -56,6 +56,23 @@ impl<S: Scalar> OwnedColumn<S> {
...
+            Self::SmallInt(col) => S::from(col[index]),
+            Self::Int(col) => S::from(col[index]),
+            Self::BigInt(col) => S::from(col[index]),
+            Self::Int128(col) => S::from(col[index]),
+            Self::Scalar(col) => col[index],
+            Self::Decimal75(_, _, col) => col[index],
+            Self::VarChar(col) => S::from(col[index].clone()),
+            Self::TimestampTZ(_, _, col) => S::from(col[index]),
+        })
+    }
```
drcapybara, line 74: Would it be better to error here instead of returning ```Option```? Index OOB seems like it could be a critical failure path

crates/proof-of-sql/src/base/math/decimal.rs
```diff
@@ -42,13 +31,17 @@ pub enum DecimalError {
...
-impl From<IntermediateDecimalError> for ConversionError {
-    fn from(err: IntermediateDecimalError) -> ConversionError {
-        DecimalConversionError(IntermediateDecimalConversionError(err))
+/// Result type for decimal operations.
+pub type DecimalResult<T> = Result<T, DecimalError>;
+
+// This exists because `TryFrom<arrow::datatypes::DataType>` for `ColumnType` error is String
+impl From<DecimalError> for String {
+    fn from(error: DecimalError) -> Self {
+        error.to_string()
```
drcapybara, line 44: Since we're here right now, it might be prudent to actually go implement the error variant `ColumnType` error

crates/proof-of-sql/src/sql/ast/numerical_util.rs
```diff
@@ -121,6 +162,55 @@ pub(crate) fn add_subtract_columns<'a, S: Scalar>(
...
+                .precision_value()
+                .expect("Precision of a valid column type should be valid"),
+        )
+        .expect("Precision of a valid column type should be valid"),
+        res_type
+            .scale()
+            .expect("Can't add or subtract non-numeric types"),
+        res,
+    )
+}
```
drcapybara, line 212: Why is add and subtract together? Would it make more sense to have separate functions for add and subtract?

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,267 @@
...
+        }
+    }
+    /// Creates a new `PostprocessingEvaluator` with the given column mapping and within aggregation scope.
+    #[allow(dead_code)]
+    pub(crate) fn new_agg(table: OwnedTable<S>) -> Self {
+        Self {
+            table,
+            _in_agg_scope: true,
+        }
+    }
```
drcapybara, line 41: is there a good reason to allow unused functions and fields here? If they will end up being used in a future PR then it seems fine

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,267 @@
...
+    ) -> PostprocessingResult<OwnedColumn<S>> {
+        let left = self.visit_expr(left)?;
+        let right = self.visit_expr(right)?;
+        let length = left.len();
+        assert_eq!(length, right.len());
+        check_dtypes(left.column_type(), right.column_type(), op)?;
+        match op {
+            BinaryOperator::And => Ok(OwnedColumn::Boolean(
+                left.as_boolean()
+                    .unwrap()
```
drcapybara, line 135: this unwrap seems like it should be handled

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,267 @@
...
+            BinaryOperator::Equal
+            | BinaryOperator::GreaterThanOrEqual
+            | BinaryOperator::LessThanOrEqual => {
+                let lhs_scale = left.column_type().scale().unwrap_or(0);
+                let rhs_scale = right.column_type().scale().unwrap_or(0);
+                let max_scale = lhs_scale.max(rhs_scale);
+                let lhs_upscale_factor =
+                    scale_scalar(S::ONE, max_scale - lhs_scale).expect("Invalid scale factor");
+                let rhs_upscale_factor =
+                    scale_scalar(S::ONE, max_scale - rhs_scale).expect("Invalid scale factor");
```
drcapybara, line 158: these ```.expects``` seem like they are reporting a failure, but it seems like expect should be used when we are 100% certain that this cannot fail, so a message explaining why the failure cannot occur seems more appropriate

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,267 @@
...
+                match op {
+                    BinaryOperator::Equal => {
+                        let res: Vec<bool> = (0..length)
+                            .map(|i| {
+                                left.scalar_at(i).unwrap() * lhs_upscale_factor
+                                    == right.scalar_at(i).unwrap() * rhs_upscale_factor
```
drcapybara, line 164: this unwrap seems like it should be handled, if this cannot panic then should it be .expect?

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,267 @@
...
+                                lhs.signed_cmp(&rhs) != Ordering::Less
+                            })
+                            .collect();
+                        Ok(OwnedColumn::Boolean(res))
+                    }
+                    BinaryOperator::LessThanOrEqual => {
+                        let res: Vec<bool> = (0..length)
+                            .map(|i| {
+                                let lhs = left.scalar_at(i).unwrap() * lhs_upscale_factor;
+                                let rhs = right.scalar_at(i).unwrap() * rhs_upscale_factor;
```
drcapybara, line 183: this unwrap seems like it should be handled, if this cannot panic then should it be .expect?

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,267 @@
...
+                if precision.value() > 38 {
+                    return None;
+                }
+            }
+            if let ColumnType::Decimal75(precision, _) = right_dtype {
+                if precision.value() > 38 {
+                    return None;
+                }
+            }
+            // TODO: inequality support for timestamps
```
drcapybara, line 237: timestamp inequality support was merged in #52, is this comment referring to something different?

crates/proof-of-sql/src/sql/ast/numerical_util.rs
iajoiner: @drcapybara They are together since add and subtract are essentially the same other than a sign.

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
iajoiner: Yes. I will add aggregation once `SelectExpr` is added in the next PR

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,254 @@
...
+impl<S: Scalar> PostprocessingEvaluator<S> {
+    /// Creates a new `PostprocessingEvaluator` with the given owned table and group by columns.
+    pub fn new(table: &OwnedTable<S>, group_by_columns: &[Identifier]) -> Self {
+        Self {
+            table: table.clone(),
+            _group_by_columns: group_by_columns.to_vec(),
+        }
+    }
+    /// Builds a `proofs::sql::ast::ProvableExprPlan` from a `proof_of_sql_parser::intermediate_ast::Expression`
+    pub fn build(&self, expr: &Expression) -> PostprocessingResult<OwnedColumn<S>> {
```
JayWhite2357, line 35: `build` seems like the wrong name.
Also, the docs appear to be wrong.

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,254 @@
...
+use proof_of_sql_parser::{
+    intermediate_ast::{BinaryOperator, Expression, Literal, UnaryOperator},
+    Identifier,
+};
+
+/// Evaluator that evaluates a `proof_of_sql_parser::intermediate_ast::Expression`
+/// on an `OwnedTable` and returns an `OwnedColumn`.
+pub struct PostprocessingEvaluator<S: Scalar> {
+    table: OwnedTable<S>,
+    _group_by_columns: Vec<Identifier>,
```
JayWhite2357, line 23: Why not just add this in a later PR rather than making it unused right now?

crates/proof-of-sql/src/base/database/scalar_and_i256_conversions.rs
```diff
@@ -14,7 +14,7 @@ const MAX_SUPPORTED_I256: i256 = i256::from_parts(
 pub fn convert_scalar_to_i256<S: Scalar>(val: &S) -> i256 {
     let is_negative = val > &S::MAX_SIGNED;
     let abs_scalar = if is_negative { -*val } else { *val };
-    let limbs = abs_scalar.into();
+    let limbs: [u64; 4] = abs_scalar.into();
```
JayWhite2357, line 17: Is this necessary? No reason to include it in the PR if not.

crates/proof-of-sql/src/sql/ast/numerical_util.rs
```diff
@@ -121,6 +163,44 @@ pub(crate) fn add_subtract_columns<'a, S: Scalar>(
...
+            if is_subtract {
+                lhs.scalar_at(i).unwrap() * lhs_upscale_factor
+                    - rhs.scalar_at(i).unwrap() * rhs_upscale_factor
+            } else {
+                lhs.scalar_at(i).unwrap() * lhs_upscale_factor
+                    + rhs.scalar_at(i).unwrap() * rhs_upscale_factor
+            }
+        })
+        .collect();
+    OwnedColumn::from_scalars(res, res_type).expect("Invalid column type")
```
JayWhite2357, line 201: I really don't like converting to Scalars, subtracting, and then converting back.
Additionally, scaling isn't always needed.
For instance, when subtracting two `BigInt` columns, we should do subtraction of `i64`s. Scaling is unneeded in that case.

crates/proof-of-sql/src/sql/ast/numerical_util.rs
```diff
@@ -121,6 +163,44 @@ pub(crate) fn add_subtract_columns<'a, S: Scalar>(
     res
 }
 
+// Add or subtract two owned columns together.
+pub(crate) fn add_subtract_owned_columns<S: Scalar>(
```
JayWhite2357, line 167: Can you put these methods in a separate PR along with unit tests for them?

JayWhite2357: Additionally, why not just `impl Sub for &OwnedColumn<S>`?

JayWhite2357 (changes requested):

crates/proof-of-sql/src/base/database/scalar_and_i256_conversions.rs
iajoiner: I think so since I added `Into<BigInt>` to `Scalar` definition

crates/proof-of-sql/src/base/database/owned_column.rs
```diff
@@ -56,6 +56,60 @@ impl<S: Scalar> OwnedColumn<S> {
...
+            ColumnType::Scalar => Some(OwnedColumn::Scalar(scalars.to_vec())),
+            ColumnType::Decimal75(precision, scale) => {
+                Some(OwnedColumn::Decimal75(precision, scale, scalars.to_vec()))
+            }
+            ColumnType::TimestampTZ(tu, tz) => scalars
+                .iter()
+                .map(|s| -> Option<i64> { (*s).try_into().ok() })
+                .collect::<Option<Vec<i64>>>()
+                .map(|s| OwnedColumn::TimestampTZ(tu, tz, s)),
+            // Boolean and VarChar columns are not supported by this function
```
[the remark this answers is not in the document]
iajoiner: Of course we can support it. It's just that there isn't any need now.

iajoiner:
Closed in favor of #81.

PR #58, opened 2024-07-24
refactor: exclusively use `DecimalError` for `proof-of-sql` decimal-related errors

iajoiner:
# Rationale for this change
This PR was split from #56. For evaluation to work it is good if decimal-related errors remain `DecimalError`s as opposed to requiring `ConversionError`. It also lowers the possibility of circular dependencies.

- eliminate any dependency of `proof-of-sql::base` on `proof-of-sql::sql` including references to `sql::parse::ConversionError`
- remove all references to `sql::parse::ParseError` from `decimal.rs`
- refactor `Scalar` def so that `TryFrom<BigInt>` now has `ScalarConversionError`

# Are these changes tested?
Existing tests should pass

### DIFF
--- a/crates/proof-of-sql/src/base/scalar/mod.rs
+++ b/crates/proof-of-sql/src/base/scalar/mod.rs
@@ -4,7 +4,6 @@ pub use error::ScalarConversionError;
 mod mont_scalar;
 #[cfg(test)]
 mod mont_scalar_test;
-use crate::sql::parse::ConversionError;
 use core::{cmp::Ordering, ops::Sub};
 pub use mont_scalar::Curve25519Scalar;
 pub(crate) use mont_scalar::MontScalar;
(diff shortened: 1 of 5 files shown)


drcapybara (dismissed):
This looks nice, thanks for doing this

crates/proof-of-sql/src/base/scalar/mod.rs
```diff
@@ -68,7 +68,7 @@ pub trait Scalar:
     + std::convert::From<i32>
     + std::convert::From<i16>
     + std::convert::From<bool>
-    + TryFrom<BigInt, Error = ConversionError>
+    + TryFrom<BigInt, Error = DecimalError>
```
JayWhite2357, line 71: This shouldn't really be a DecimalError since there are no decimals involved. It is a conversion from BigInt to Scalar.
It's ultimately used for a decimal conversion, but at this point in the code, isn't decimal related.

JayWhite2357 (approved):

PR #59, opened 2024-07-24
refactor: simplify `proof-of-sql-parser`

iajoiner:
# Rationale for this change
For intermediate AST -> Provable AST and intermediate AST postprocessing evaluation tests in `proof-of-sql` it is a good idea to be able to use shortcuts in `proof_of_sql_parser::test_utility`. 

- put `test_utility.rs` in parser behind a feature flag and allow it to be used by the other crate
- prevent `posql_time` from being unnecessarily convoluted

# Are these changes tested?
Existing tests should pass

### DIFF
--- a/crates/proof-of-sql-parser/src/lib.rs
+++ b/crates/proof-of-sql-parser/src/lib.rs
@@ -9,11 +9,12 @@ extern crate lalrpop_util;
 
 pub mod intermediate_ast;
 
-#[cfg(test)]
+#[cfg(all(test, feature = "parser-test-utility"))]
 mod intermediate_ast_tests;
 
-#[cfg(test)]
-pub(crate) mod test_utility;
+#[cfg(feature = "parser-test-utility")]
+/// Shortcuts to construct intermediate AST nodes.
+pub mod test_utility;
 
 pub(crate) mod select_statement;
 pub use select_statement::SelectStatement;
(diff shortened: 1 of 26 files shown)


drcapybara (dismissed):
This is a nice refactor, thanks. Feel free to move the ```Timestamp``` error module too if it makes sense

drcapybara (approved):
Looks nice, thanks

crates/proof-of-sql-parser/src/lib.rs
```diff
@@ -9,11 +9,12 @@ extern crate lalrpop_util;
 
 pub mod intermediate_ast;
 
-#[cfg(test)]
+#[cfg(all(test, feature = "parser-test-utility"))]
```
JayWhite2357, line 12: Is this change needed? These are just tests, right?
```suggestion
#[cfg(test)]
```

crates/proof-of-sql-parser/src/test_utility.rs
JayWhite2357, line 1: Any reason we can't simply make this available by default? I don't see a problem with calling this `utility` instead of `test_utility` and just simply making it `pub` instead of putting it behind a feature.

My only suggestion in that case would be to make the documentation a bit more verbose.

JayWhite2357 (approved):
Approved if you don't want to apply my suggestions.

iajoiner:
@JayWhite2357 Actually I will..in a future PR very soon.

PR #63, opened 2024-07-26
build!: revert "build: upgrade to arrow 51"

iajoiner:
Temporarily reverts spaceandtimelabs/sxt-proof-of-sql#60 since we should have bumped Rust version in services before bumping Arrow version in dependencies.


iajoiner:
Closed in favor of the next PR

PR #68, opened 2024-07-30
feat: add Dynamic Dory helper methods

JayWhite2357:
# Rationale for this change

We need to add the Dynamic Dory commitment scheme. This is the first PR.

* Added the `compute_dynamic_standard_basis_vecs` method, a helper method used to compute the vectors in the Vector-Matrix-Vector product in the dynamic dory scheme.
* Added `row_and_column_from_index` and `row_start_index` methods. These methods define the structure of the new dynamic dory commitmemt scheme.

# Are these changes tested?

Yes

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_standard_basis_helper.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_standard_basis_helper.rs
@@ -1,0 +1,281 @@
+//! This module provides the `build_standard_basis_vecs` method, which is used in converting a point to a
+//! vector used in a Vector-Matrix-Vector product in the dynamic dory scheme.
+
+use super::F;
+use ark_ff::Field;
+
+#[allow(dead_code)]
+/// This method produces evaluation vectors from a point. This is a helper method for generating a Vector-Matrix-Vector product in the dynamic dory scheme.
+///
+/// The ith element of the lo_vec is essentially the ith monomial basis element (lexicographically).
+/// The ith element of the hi_vec is essentially the jth monomial basis element where j = row_start_index(i).
+///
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_standard_basis_helper.rs
```diff
@@ -0,0 +1,281 @@
...
+use ark_ff::Field;
+
+#[allow(dead_code)]
+/// This method produces evaluation vectors from a point. This is a helper method for generating a Vector-Matrix-Vector product in the dynamic dory scheme.
+///
+/// The ith element of the lo_vec is essentially the ith monomial basis element (lexicographically).
+/// The ith element of the hi_vec is essentially the jth monomial basis element where j = row_start_index(i).
+///
+/// NOTE: the lo_vec and hi_vec are scaled by lo_vec[0] and hi_vec[0] respectively.
+/// NOTE: lo_vec and hi_vec should otherwise consist entirely of zeros in order to ensure correct output.
```
tlovell-sxt, line 14: So what's the purpose of accepting a `lo_vec` and a `hi_vec` then? why not just a `lo_scale` and a `hi_scale` and return `(Vec<F>, Vec<F>)` (or use bumpalo if you have to return slices). performance reasons?

JayWhite2357: Well, this is really just a helper method for another function that constructs the vectors and scale.
I could do that here, but I think it makes the most sense for this function to not handle the setup and simply be a driver of the algorithm.

tlovell-sxt (approved):

PR #69, opened 2024-07-30
refactor: add flexibility for dory verifier folding

JayWhite2357:
# Rationale for this change

We need to add the Dynamic Dory commitment scheme. In order to make the existing methods work with the dynamic dory commitment scheme, the method that folds the s tensors within the verify methods needs to be modified. In particular, it needs to have access to all the folding factors at a time in order to fold properly.

The folding of the s tensor is now done once within the `fold_scalars_0_verify` rather than accumulating within each round.
There are 2 main changes to accomplish this:
* During each round/call of `extended_dory_reduce_verify`, instead of mutating the s tensors, the `alpha` folding factors are added to the `state`.
* `fold_s_tensors_verify` is added as a parameter to `extended_dory_inner_product_verify` and `fold_scalars_0_verify`

# Are these changes tested?

Existing tests pass.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/extended_dory_inner_product.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/extended_dory_inner_product.rs
@@ -1,6 +1,6 @@
 use super::{
     scalar_product_prove, scalar_product_verify, DoryMessages, ExtendedProverState,
-    ExtendedVerifierState, ProverSetup, VerifierSetup,
+    ExtendedVerifierState, ProverSetup, VerifierSetup, F,
 };
 use crate::proof_primitive::dory::{
     extended_dory_reduce_prove, extended_dory_reduce_verify, fold_scalars_0_prove,
@@ -34,6 +34,7 @@ pub fn extended_dory_inner_product_verify(
     transcript: &mut Transcript,
     mut state: ExtendedVerifierState,
     setup: &VerifierSetup,
+    fold_s_tensors_verify: impl Fn(&ExtendedVerifierState) -> (F, F),
 ) -> bool {
     let nu = state.base_state.nu;
     assert!(setup.max_nu >= nu);
@@ -42,6 +43,7 @@ pub fn extended_dory_inner_product_verify(
             return false;
         }
     }
-    let base_state = fold_scalars_0_verify(messages, transcript, state, setup);
+    let base_state =
+        fold_scalars_0_verify(messages, transcript, state, setup, fold_s_tensors_verify);
     scalar_product_verify(messages, transcript, base_state, setup)
 }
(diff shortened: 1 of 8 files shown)


crates/proof-of-sql/src/proof_primitive/dory/extended_dory_inner_product.rs
```diff
@@ -34,6 +34,7 @@ pub fn extended_dory_inner_product_verify(
     transcript: &mut Transcript,
     mut state: ExtendedVerifierState,
     setup: &VerifierSetup,
+    fold_s_tensors_verify: fn(&ExtendedVerifierState) -> (F, F),
```
tlovell-sxt, line 37: NIT: for functions as parameters, using a trait bound instead of a fn pointer is a little more flexible for other use cases (like closures)
```suggestion
    fold_s_tensors_verify: impl Fn(&ExtendedVerifierState) -> (F, F),
```

crates/proof-of-sql/src/proof_primitive/dory/fold_scalars.rs
```diff
@@ -33,11 +33,11 @@ pub fn fold_scalars_0_verify(
     transcript: &mut Transcript,
     mut state: ExtendedVerifierState,
     setup: &VerifierSetup,
+    fold_s_tensors_verify: fn(&ExtendedVerifierState) -> (F, F),
```
tlovell-sxt, line 36: NIT: same comment
```suggestion
    fold_s_tensors_verify: impl Fn(&ExtendedVerifierState) -> (F, F),
```

tlovell-sxt (approved):
approved w/ nits

crates/proof-of-sql/src/proof_primitive/dory/extended_dory_reduce_helper.rs
```diff
@@ -104,12 +104,19 @@ pub fn extended_dory_reduce_verify_update_Es(
...
+            .iter()
+            .zip(state.alphas.iter())
+            .map(|(s, a)| (F::ONE - s) * a + s)
+            .product(),
+        state
+            .s2_tensor
+            .iter()
+            .zip(state.alpha_invs.iter())
+            .map(|(s, a)| (F::ONE - s) * a + s)
+            .product(),
```
jacobtrombetta, line 120: Is `s` an element of the tensor and `a` an element of alpha? I would think `.map(|(s, a)| (F::ONE - s) * a + s)`  should be `.map(|(s, a)| (F::ONE - a) * s + a)` to match the previous code? But the tests are passing, so I might be misunderstanding something?

JayWhite2357: Good eye. This is why it works:

$$(1 - s) \cdot a + s = a - s\cdot a + s = s - a\cdot s + a = (1 - a) \cdot s + a$$

jacobtrombetta (approved):
LGTM

tlovell-sxt (approved):

PR #70, opened 2024-07-30
feat: improve numerical operations

iajoiner:
# Rationale for this change
Postprocessing and proving share a lot of numerical operations so we should clean them up and unify them. This is the first PR in the process.

- move `ColumnType` manipulation from `sql::ast` to `base::database` since it is not for proving only
- define `ColumnOperationError`

# Are these changes tested?
Yes.


crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -0,0 +1,202 @@
...
+            (Self::SmallInt(lhs), Self::SmallInt(rhs)) => Ok(Self::SmallInt(add_vecs(lhs, rhs))),
+            (Self::SmallInt(lhs), Self::Int(rhs)) => Ok(Self::Int(add_vecs_with_casting(lhs, rhs))),
+            (Self::SmallInt(lhs), Self::BigInt(rhs)) => {
+                Ok(Self::BigInt(add_vecs_with_casting(lhs, rhs)))
+            }
+            (Self::SmallInt(lhs), Self::Int128(rhs)) => {
+                Ok(Self::Int128(add_vecs_with_casting(lhs, rhs)))
+            }
+            (Self::SmallInt(lhs), Self::Decimal75(precision, scale, rhs)) => {
+                add_decimal_columns(lhs, rhs, 5, 0, precision.value(), scale)
```
JayWhite2357, line 102: Either this is wrong, or the one below is.
```suggestion
                add_decimal_columns(lhs, rhs, 0, 5, precision.value(), scale)
```

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -0,0 +1,202 @@
...
+    left_scale: i8,
+    right_precision_value: u8,
+    right_scale: i8,
+) -> ColumnOperationResult<OwnedColumn<S>>
+where
+    S: Scalar + core::convert::From<T0> + core::convert::From<T1>,
+    T0: Copy,
+    T1: Copy,
+{
+    let new_scale = left_scale.max(right_scale);
```
JayWhite2357, line 52: I wonder if the following would be simpler:
```
if left_scale = right_scale {
    // call either `add_vecs` or `add_vecs_with_casting`, not sure which one.
} else if left_scale > right_scale {
    // call `add_decimal_columns` with `lhs` and `rhs` reversed.
} else {
    // only the rhs needs to be scaled up, which should simplify the code
}
```

crates/proof-of-sql/src/base/database/column_operation.rs
iajoiner: Nah. Precision before scale.

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -0,0 +1,202 @@
...
+            }
+            (Self::Int128(lhs), Self::Int128(rhs)) => Ok(Self::Int128(add_vecs(lhs, rhs))),
+            (Self::Int128(lhs), Self::Decimal75(precision, scale, rhs)) => {
+                add_decimal_columns(lhs, rhs, 39, 0, precision.value(), scale)
+            }
+            (Self::Int128(lhs), Self::Scalar(rhs)) => {
+                Ok(Self::Scalar(add_vecs_with_casting(lhs, rhs)))
+            }
+            (Self::Decimal75(precision, scale, lhs), Self::SmallInt(rhs)) => {
+                add_decimal_columns(rhs, lhs, 5, 0, precision.value(), scale)
```
JayWhite2357, line 154: This is what I meant. I didn't look close enough on the previous comment.
```suggestion
                add_decimal_columns(rhs, lhs, precision.value(), scale, 5, 0)
```

crates/proof-of-sql/src/base/database/column_operation.rs
JayWhite2357: Right. See my other comment.

crates/proof-of-sql/src/base/database/column_operation.rs
JayWhite2357: Oh, never mind, I see that you have `rhs` coming first here.

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -0,0 +1,364 @@
...
+    let new_scale = new_column_type.scale().expect("numeric columns have scale");
+    let left_upscale = new_scale
+        - left_column_type
+            .scale()
+            .expect("numeric columns have scale");
+    let right_upscale = new_scale
+        - right_column_type
+            .scale()
+            .expect("numeric columns have scale");
+    // One of left_scale and right_scale is 0 so we can avoid scaling when unnecessary
```
JayWhite2357, line 156: What if both are `0`?

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -0,0 +1,364 @@
...
+        let left_precision_value = lhs.precision_value().unwrap_or(0);
+        let right_precision_value = rhs.precision_value().unwrap_or(0);
+        let precision_value = left_precision_value + right_precision_value + 1;
+        let precision = Precision::new(precision_value).map_err(|_| {
+            ColumnOperationError::DecimalConversionError(DecimalError::InvalidPrecision(format!(
+                "Required precision {} is beyond what we can support",
+                precision_value
+            )))
+        })?;
+        let left_scale = lhs.scale().unwrap_or(0);
```
JayWhite2357, line 91: If `scale` returns none, it should mean there is no scale (e.g. VarChar).
If the type actually has a scale of 0, then scale should return `Some(0)`.

JayWhite2357: Same applies to `precision_value`.

crates/proof-of-sql/src/base/database/column_operation.rs
iajoiner: We can make it another special case

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -0,0 +1,371 @@
...
+    let right_upscale = new_scale
+        - right_column_type
+            .scale()
+            .expect("numeric columns have scale");
+    // One of left_scale and right_scale is 0 so we can avoid scaling when unnecessary
+    let scalars: Vec<S> = if right_upscale == 0 {
+        lhs.iter()
+            .zip(rhs)
+            .map(|(l, r)| -> ColumnOperationResult<S> {
+                Ok(scale_scalar(S::from(*l), left_upscale)? + S::from(*r))
```
JayWhite2357, line 163: `scale_scalar` has to compute the 10^left_upscale each time here, which will by far dominate the conversion.
If you do the `upscale_factor = 10^left_upscale` outside of the iter, you can simply do:
```suggestion
S::from(*l) * upscale_factor + S::from(*r)
```
and even avoid the `Ok`.

crates/proof-of-sql/src/base/math/decimal.rs
```diff
@@ -187,17 +187,20 @@ pub(crate) fn try_into_to_scalar<S: Scalar>(
 /// Scale scalar by the given scale factor. Negative scaling is not allowed.
 /// Note that we do not check for overflow.
 pub(crate) fn scale_scalar<S: Scalar>(s: S, scale: i8) -> DecimalResult<S> {
```
JayWhite2357, line 189: See earlier comment. I almost think it would be better to drop this method in favor of a `Scalar::exp10(power: i8) -> Scalar` method. 10 and 1/10 could easily be associated constants on the `Scalar` trait to handle negative powers nicely as well.
Then, we could replace
`scale_scalar(value, scale_factor)` with 
`value * Scalar::exp10(scale_factor)` which is much more readable in my opinion and doesn't need to return a `Result`.

JayWhite2357 (approved):

PR #72, opened 2024-07-31
fix: timestamp precision parsing

drcapybara:
# Rationale for this change

Timestamp literals were parsed in seconds despite containing up to nanosecond precision. This caused a situation where equality and inequality queries would fail across the board. This PR updates timestamp literal parsing to use the correct time unit precision corresponding to whatever data is inserted in the table..

- [x] cover precision match in literal conversion
- [x] update tests to parse rfc3339 strings in addition to epochs

# Are these changes tested?

- [x] test inequality support between strings of differing precision
- [x] test complex query and filter
- [x] cover edge cases with string parsing in addition to epoch parsing
- [x] cover differing timezone tests with string parsing
- [x] cover differing timezone and precision tests with string parsing
- [x] test that timestamps in differing precisions and scales can match appropriately
- [ ] reorganize tests to appropriate modules

### DIFF
--- a/crates/proof-of-sql/src/base/database/owned_table_utility.rs
+++ b/crates/proof-of-sql/src/base/database/owned_table_utility.rs
@@ -17,7 +17,7 @@ use super::{OwnedColumn, OwnedTable};
 use crate::base::scalar::Scalar;
 use core::ops::Deref;
 use proof_of_sql_parser::{
-    posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
+    posql_time::{PoSQLTimeUnit, PoSQLTimeZone, PoSQLTimestamp},
     Identifier,
 };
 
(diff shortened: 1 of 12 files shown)


crates/proof-of-sql-parser/Cargo.toml
```diff
@@ -15,7 +15,6 @@ doctest = true
 test = true
 
 [dependencies]
-arrow = { workspace = true }
```
iajoiner, line 18: :-)

crates/proof-of-sql/src/sql/parse/provable_expr_plan_builder.rs
```diff
@@ -100,9 +101,21 @@ impl ProvableExprPlanBuilder<'_> {
...
+                    and 2262-04-11T23:47:16.854775807."
+                            .to_owned(),
+                    )
+                })?;
+
+                Ok(ProvableExprPlan::new_literal(LiteralValue::TimeStampTZ(
+                    its.timezone,
+                    timestamp,
+                )))
+            }
```
drcapybara: this was the cause of the bug. keeping the time unit would mean matching to each instance of it in the its (intermediate timestamp), but the intermediate timestamp has no way of knowing the precision of the column it will eventually be matched to just from this point

crates/proof-of-sql/tests/timestamp_integration_tests.rs
```diff
@@ -119,98 +173,357 @@ fn run_timestamp_query_test(
...
+            "2009-01-03T18:15:05.000000000Z",
+        ];
+        let expected_timestamps = &["2009-01-03T18:15:05.000000000Z"];
+        run_timestamp_query_test(
+            "SELECT * FROM table WHERE times = timestamp '2009-01-03T18:15:05Z';",
+            test_timestamps,
+            PoSQLTimeUnit::Nanosecond,
+            expected_timestamps,
+            PoSQLTimeUnit::Nanosecond,
+        );
```
drcapybara, line 257: this fails with:
```
assertion `left == right` failed
  left: OwnedTable { table: {Identifier { name: "times" }: TimestampTZ(Nanosecond, Utc, [1231006505000000000])} }
 right: OwnedTable { table: {Identifier { name: "times" }: TimestampTZ(Second, Utc, [1231006505000000000])} }
```

drcapybara:
```rust
run_timestamp_query_test(
    "SELECT * FROM table WHERE times = timestamp '2009-01-03T18:15:05Z';",
    test_timestamps,
    PoSQLTimeUnit::Nanosecond,
    expected_timestamps,
    PoSQLTimeUnit::Second,
);
```
using ```Second``` as the expected time unit gives:

```
assertion `left == right` failed
  left: OwnedTable { table: {Identifier { name: "times" }: TimestampTZ(Nanosecond, Utc, [1231006505000000000])} }
 right: OwnedTable { table: {Identifier { name: "times" }: TimestampTZ(Second, Utc, [1231006505])} }
 ```

File: crates/proof-of-sql/src/base/database/owned_table_utility.rs [hunk 4]
```diff
@@ -231,3 +235,66 @@ pub fn timestamptz<S: Scalar>(
...
+        .collect();
+
+    // Build the OwnedColumn using the time unit from the first timestamp in parsed_data
+    let result = (
+        name.parse().unwrap(),
+        OwnedColumn::TimestampTZ(
+            parsed_data
+                .first()
+                .expect("No timestamps provided")
+                .timeunit,
```
JayWhite2357, line 285: This should fix the tests:
```suggestion
            time_unit,
```

crates/proof-of-sql/src/base/database/owned_table_utility.rs
```diff
@@ -217,10 +217,14 @@ pub fn decimal75<S: Scalar>(
...
-///     timestamptz("event_time", PoSQLTimeUnit::Second, PoSQLTimeZone::Utc, vec![1625072400, 1625076000, 1625079600]),
+///     timestamptz_epoch("event_time",
+///                        PoSQLTimeUnit::Second,
+///                        PoSQLTimeZone::Utc,
+///                        vec![1625072400, 1625076000, 1625079600]
+///     ),
 /// ]);
 /// ```
-pub fn timestamptz<S: Scalar>(
+pub fn timestamptz_epoch<S: Scalar>(
```
JayWhite2357, line 227: I would prefer to keep this named `timestamptz`.

crates/proof-of-sql/src/base/database/owned_table_utility.rs
```diff
@@ -231,3 +235,66 @@ pub fn timestamptz<S: Scalar>(
...
+///         timestamptz("event_time",
+///         PoSQLTimeUnit::Second,
+///         vec![
+///             "1969-12-31T23:59:59Z", // One second before the Unix epoch
+///             "1970-01-01T00:00:00Z", // The Unix epoch
+///             "1970-01-01T00:00:01Z", // One second after the Unix epoch
+///         ].iter().map(|s| s.to_string())),
+/// ]);
+/// ```
+pub fn timestamptz<S: Scalar>(
```
JayWhite2357, line 266: Rename this one.
```suggestion
pub fn timestamptz_from_strings<S: Scalar>(
```
Also, add unit tests for this method since it is non-trivial. (And has a bug.)

Although, I think it makes more sense to remove this method altogether and just handle parsing in the test itself like this:
```rust
timestamptz_epoch("a", PoSQLTimeUnit::Millisecond, PoSQLTimeZone::Utc,
    ["1969-12-31T23:59:59Z", "1970-01-01T00:00:00Z", "1970-01-01T00:00:01Z"].iter()
        .map(|s| 
            PoSQLTimestamp::try_from(s).unwrap().timestamp.timestamp_millis()
        )

```

crates/proof-of-sql/tests/timestamp_integration_tests.rs
JayWhite2357, line 1: Most of these tests should be moved into the appropriate ast node. For example, `test_precision_and_rounding` is really just testing that equality is working properly for a variety of timestamp situations. So, it should be in the `equals_expr_test.rs` file.

This also makes it easier to break up testing the equality and testing the parsing. You should add a test, in `where_expr_builder_tests.rs`, that ensures a string gets parsed into the appropriate ast nodes. This is because the `equals_expr_test.rs` code shouldn't do parsing and only accepts an ast node since that it out of the scope of that unit test.

Doing it this way often means you actually need _fewer_ tests to cover the same amount of cases.

JayWhite2357 (changes requested):

crates/proof-of-sql/src/base/database/owned_table_utility.rs
drcapybara: This function is not creating a slice of ```PoSqlTimestamps```, it is creating an owned table. Removing this function makes it far more cumbersome to instantiate owned tables of ```timestamptz``` elsewhere in the code

PR #73, opened 2024-07-31
feat: range check node

drcapybara:
# Rationale for this change

Range checks are useful constructions in ZK argument systems for efficiently proving that a value falls within a certain range. This PR updates the existing range check mechanisms to improve efficiency and proof asymptotic performance of the existing approach.

range check TODO:

1. - [ ] varying word width - want bits in word size to match log of width of table, need to adjust committable column types, need a lower bound on word size -
2. - [ ] log deriv - needs commitment to all possible values in word
3. - [ ] drop unused words

# Are these changes tested?


crates/proof-of-sql/src/sql/ast/range_check_test_expr.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    fn prover_evaluate<'a>(
+        &self,
+        builder: &mut ProofBuilder<'a, S>,
+        alloc: &'a Bump,
+        accessor: &'a dyn DataAccessor<S>,
+    ) {
+        let a = accessor.get_column(self.column);
+
+        let scalar_vector = a.clone().to_scalar_with_scaling(1);
```
JayWhite2357, line 41:
```suggestion
        let scalar_vector = a.clone().to_scalar_with_scaling(0);
```

crates/proof-of-sql/src/sql/ast/range_check.rs
```diff
@@ -0,0 +1,88 @@
...
+        .map(|s| unsafe {
+            // Convert Scalar to [u64; 4] and then to &[u8]
+            let scalar_array: [u64; 4] = (*s).into(); // Using `Into` trait to convert Scalar directly
+            let scalar_bytes: &[u8] = std::slice::from_raw_parts(
+                scalar_array.as_ptr() as *const u8,
+                32, // Each u64 is 8 bytes, so [u64; 4] is 32 bytes
+            );
+            &scalar_bytes[..31] // Take the first 31 bytes
+        })
+        .collect();
```
JayWhite2357, line 37: I don't think this is doing what you think it's doing. Extract it into a function and add unit tests.

crates/proof-of-sql/src/sql/ast/range_check_test_expr.rs
```diff
@@ -0,0 +1,139 @@
...
+        HashSet::from([self.column])
+    }
+
+    fn count(
+        &self,
+        builder: &mut CountBuilder,
+        _accessor: &dyn MetadataAccessor,
+    ) -> Result<(), ProofError> {
+        builder.count_anchored_mles(0);
+        builder.count_intermediate_mles(32);
```
JayWhite2357, line 78: You are only expecting 31 intermediate mles, right?
```suggestion
        builder.count_intermediate_mles(31);
```

crates/proof-of-sql/src/sql/ast/range_check_test_expr.rs
```diff
@@ -0,0 +1,139 @@
...
+        builder.count_result_columns(0);
+        builder.count_subpolynomials(1);
+        builder.count_degree(2);
```
JayWhite2357, line 81: You don't currently add any sumcheck constraints/subpolynomials.

```suggestion
        builder.count_subpolynomials(0);
        builder.count_degree(2);
```

OR

```suggestion
```

PR #76, opened 2024-08-01
refactor: explicitly mention all enum variants in `ColumnType::precision_value` and `ColumnType::scale`

iajoiner:
# Rationale for this change
This is a clean and explicit solution. Note that for all timestamp types we have `scale = 0` since in #75 time unit is removed anyway. Also this is a refactoring since it doesn't affect functionality.

- Explicitly mention all enum variants in `ColumnType::precision_value` and `ColumnType::scale`

# Are these changes tested?
Yes


tlovell-sxt (approved):
Good stuff, thanks

PR #77, opened 2024-08-01
feat: add division to `column_operation` and `Div` to `OwnedColumn`

iajoiner:
# Rationale for this change
Division is needed in postprocessing hence we need to add it. Please merge #70 before reviewing this PR or alternatively look at the latest commit.

- add division to `column_operation`
- implement `Div` for `OwnedColumn`

# Are these changes tested?
Yes.

### DIFF
--- a/crates/proof-of-sql/src/base/database/column_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_operation.rs
@@ -5,7 +5,11 @@ use crate::base::{
     scalar::Scalar,
 };
 use core::fmt::Debug;
-use num_traits::ops::checked::{CheckedAdd, CheckedMul, CheckedSub};
+use num_bigint::BigInt;
+use num_traits::{
+    ops::checked::{CheckedAdd, CheckedDiv, CheckedMul, CheckedSub},
+    Zero,
+};
 use proof_of_sql_parser::intermediate_ast::BinaryOperator;
 
 // For decimal type manipulation please refer to
@@ -102,6 +106,53 @@ pub fn try_multiply_column_types(
     }
 }
 
+/// Determine the output type of a division operation if it is possible
+/// to multiply the two input types. If the types are not compatible, return
+/// an error.
+pub fn try_divide_column_types(
+    lhs: ColumnType,
+    rhs: ColumnType,
+) -> ColumnOperationResult<ColumnType> {
+    if !lhs.is_numeric()
+        || !rhs.is_numeric()
+        || lhs == ColumnType::Scalar
+        || rhs == ColumnType::Scalar
+    {
+        return Err(ColumnOperationError::BinaryOperationInvalidColumnType {
+            operator: BinaryOperator::Division,
+            left_type: lhs,
+            right_type: rhs,
+        });
+    }
+    if lhs.is_integer() && rhs.is_integer() {
+        // We can unwrap here because we know that both types are integers
+        return Ok(lhs.max_integer_type(&rhs).unwrap());
+    }
+    let left_precision_value = lhs.precision_value().expect("Numeric types have precision") as i16;
+    let right_precision_value = rhs.precision_value().expect("Numeric types have precision") as i16;
+    let left_scale = lhs.scale().expect("Numeric types have scale") as i16;
+    let right_scale = rhs.scale().expect("Numeric types have scale") as i16;
+    let raw_scale = (left_scale + right_precision_value + 1_i16).max(6_i16);
+    let precision_value: i16 = left_precision_value - left_scale + right_scale + raw_scale;
+    let scale = i8::try_from(raw_scale).map_err(|_| {
+        ColumnOperationError::DecimalConversionError(DecimalError::InvalidScale(raw_scale))
+    })?;
+    let precision = u8::try_from(precision_value)
+        .map_err(|_| {
+            ColumnOperationError::DecimalConversionError(DecimalError::InvalidPrecision(
+                precision_value.to_string(),
+            ))
+        })
+        .and_then(|p| {
+            Precision::new(p).map_err(|_| {
+                ColumnOperationError::DecimalConversionError(DecimalError::InvalidPrecision(
+                    p.to_string(),
+                ))
+            })
+        })?;
+    Ok(ColumnType::Decimal75(precision, scale))
+}
+
 /// Try to add two slices of the same length.
 ///
 /// We do not check for length equality here. However, we do check for integer overflow.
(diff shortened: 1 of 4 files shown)


crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -398,6 +507,61 @@ where
     ))
 }
 
+/// Divide an owned column by another.
+///
+/// We do not check for length equality here
```
JayWhite2357, line 512: Can you document the behavior that this gives. (i.e. rounding, precision, scale, divide by zero, etc.)
Can you also add a note of what this is intended to mirror comes from?

crates/proof-of-sql/src/base/scalar/mont_scalar.rs
```diff
@@ -187,7 +190,14 @@ impl<T: MontConfig<4>> From<[u64; 4]> for MontScalar<T> {
         Self(Fp::new(ark_ff::BigInt(value)))
     }
 }
-
+impl<T: MontConfig<4>> From<MontScalar<T>> for BigInt {
+    fn from(val: MontScalar<T>) -> Self {
+        println!("val: {:?}", val.0);
+        let raw: [u8; 32] = bytemuck::cast(val.0 .0 .0);
```
JayWhite2357, line 196: Use `Into<[u64; 4]>`.

iajoiner: @JayWhite2357 Of course! That was taken care of in #78.

crates/proof-of-sql/src/base/database/column_operation.rs
iajoiner: Sure!

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -398,6 +505,63 @@ where
...
+///   - new_scale = max(6, right_precision + left_scale + 1)
+///   - new_precision = left_precision - left_scale + right_scale + new_scale
+pub(crate) fn try_divide_decimal_columns<S, T0, T1>(
+    lhs: &[T0],
+    rhs: &[T1],
+    left_column_type: ColumnType,
+    right_column_type: ColumnType,
+) -> ColumnOperationResult<(Precision, i8, Vec<S>)>
+where
+    S: Scalar + core::convert::From<T0> + core::convert::From<T1>,
```
JayWhite2357, line 524: This would result in bypassing `Scalar` conversions if `lhs` and `rhs` are integers:
```suggestion
    S: Scalar,
    T0: Into<BigInt>,
    T1: Into<BigInt>,
```

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -398,6 +505,63 @@ where
...
+        .expect("numeric columns have precision");
+    let new_scale = new_column_type.scale().expect("numeric columns have scale");
+    let lhs_scale = left_column_type
+        .scale()
+        .expect("numeric columns have scale");
+    let rhs_scale = right_column_type
+        .scale()
+        .expect("numeric columns have scale");
+    let res: Vec<S> = (0..len)
+        .map(|i| -> ColumnOperationResult<S> {
```
JayWhite2357, line 541: NIT: Since `i` is only indexing the `lhs` and `rhs`, you could do this:
```suggestion
    let res: Vec<S> = lhs.iter().zip(rhs)
        .map(|(l, r)| -> ColumnOperationResult<S> {
```

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -256,6 +322,47 @@ where
...
+/// Divide one slice by another of the same length, casting the left slice to the type of the right slice.
+///
+/// We do not check for length equality here
+pub(super) fn try_divide_slices_left_upcast<SmallerType, LargerType>(
+    lhs: &[SmallerType],
+    rhs: &[LargerType],
+) -> ColumnOperationResult<Vec<LargerType>>
+where
+    SmallerType: Copy + Debug,
+    LargerType: From<SmallerType> + CheckedDiv<Output = LargerType> + Copy + Debug,
```
JayWhite2357, line 334: NIT:
```suggestion
    SmallerType: Copy + Debug + Into<LargerType>,
    LargerType: CheckedDiv<Output = LargerType> + Copy + Debug,
```

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -256,6 +322,47 @@ where
...
+/// Divide one slice by another of the same length, casting the right slice to the type of the left slice.
+///
+/// We do not check for length equality here
+pub(super) fn try_divide_slices_right_upcast<SmallerType, LargerType>(
+    lhs: &[LargerType],
+    rhs: &[SmallerType],
+) -> ColumnOperationResult<Vec<LargerType>>
+where
+    SmallerType: Copy + Debug,
+    LargerType: From<SmallerType> + CheckedDiv<Output = LargerType> + Copy + Debug,
```
JayWhite2357, line 355: NIT:
```suggestion
    SmallerType: Copy + Debug + Into<LargerType>,
    LargerType: CheckedDiv<Output = LargerType> + Copy + Debug,
```

JayWhite2357 (changes requested):
Looks good. A few minor changes that I would like to see and then I'll approve.

crates/proof-of-sql/src/base/database/column_operation.rs
iajoiner: @JayWhite2357 If they are both integers we shouldn't even use this function.

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -398,6 +505,63 @@ where
...
+    let res: Vec<S> = (0..len)
+        .map(|i| -> ColumnOperationResult<S> {
+            let lhs_bigint = Into::<BigInt>::into(S::from(lhs[i]));
+            let rhs_bigint = Into::<BigInt>::into(S::from(rhs[i]));
+            if rhs_bigint.is_zero() {
+                return Err(ColumnOperationError::DivisionByZero);
+            }
+            let applied_scale = rhs_scale - lhs_scale + new_scale;
+            let new_bigint = if applied_scale >= 0 {
+                let applied_scale_factor = BigInt::from(10).pow(applied_scale as u32);
```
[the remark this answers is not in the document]
iajoiner: Yup.

crates/proof-of-sql/src/base/database/column_operation.rs
iajoiner: Right. We should implement `From` as opposed to `Into` but when it comes to type bounds we want to use `Into` instead since `From` implies `Into`.

crates/proof-of-sql/src/base/database/column_operation.rs
JayWhite2357: But if one of them are, we do, right?

JayWhite2357 (approved):

PR #78, opened 2024-08-02
feat: add `Into<BigInt>` to `Scalar` & improvement to `Scalar` in general

iajoiner:
# Rationale for this change
This PR is a prerequisite of #77. 

- add implementation as a part of `scalar_conversion_to_int` macro which is necessary since blanket implementation of `TryFrom` or `TryInto` is not allowed for traits by Rust
- improve the macro so that `TryInto` is replaced with the more natural `TryFrom`
- replace `std` with `core`

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/base/database/scalar_and_i256_conversions.rs
+++ b/crates/proof-of-sql/src/base/database/scalar_and_i256_conversions.rs
@@ -14,7 +14,7 @@ const MAX_SUPPORTED_I256: i256 = i256::from_parts(
 pub fn convert_scalar_to_i256<S: Scalar>(val: &S) -> i256 {
     let is_negative = val > &S::MAX_SIGNED;
     let abs_scalar = if is_negative { -*val } else { *val };
-    let limbs = abs_scalar.into();
+    let limbs: [u64; 4] = abs_scalar.into();
 
     let low = (limbs[0] as u128) | ((limbs[1] as u128) << 64);
     let high = (limbs[2] as i128) | ((limbs[3] as i128) << 64);
--- a/crates/proof-of-sql/src/base/scalar/mod.rs
+++ b/crates/proof-of-sql/src/base/scalar/mod.rs
@@ -30,43 +30,44 @@ pub trait Scalar:
     + Sync
     + Send
     + num_traits::One
-    + std::iter::Sum
-    + std::iter::Product
+    + core::iter::Sum
+    + core::iter::Product
     + Sub<Output = Self>
     + Copy
-    + std::ops::MulAssign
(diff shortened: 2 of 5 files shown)


crates/proof-of-sql/src/base/database/scalar_and_i256_conversions.rs
```diff
@@ -14,7 +14,7 @@ const MAX_SUPPORTED_I256: i256 = i256::from_parts(
 pub fn convert_scalar_to_i256<S: Scalar>(val: &S) -> i256 {
     let is_negative = val > &S::MAX_SIGNED;
     let abs_scalar = if is_negative { -*val } else { *val };
-    let limbs = abs_scalar.into();
+    let limbs: [u64; 4] = abs_scalar.into();
```
iajoiner, line 17: Required since `Scalar` has another `Into`

drcapybara (dismissed):
lgtm

crates/proof-of-sql/src/base/scalar/mont_scalar.rs
```diff
@@ -356,6 +356,26 @@ impl<T: MontConfig<4>> Display for MontScalar<T> {
...
+                    num_bigint::Sign::Plus
+                };
+                let value_abs: ark_ff::BigInt<4> =
+                    (if is_negative { -value } else { value }).0.into_bigint();
+                BigInt::from_bytes_le(sign, &value_abs.to_bytes_le())
+            }
+        }
+    };
+}
+pub(crate) use mont_scalar_conversion_to_bigint;
```
tlovell-sxt, line 377: NIT:  I'd prefer to minimize the code that goes in macros as they tend to obfuscate compilation issues. Does writing this conversion generically on `S: Scalar` cause conflicting implementations or something?

If so, could you instead write a generic `fn scalar_to_bigint<S: Scalar>` that has the bulk of this LOC and is called by the macros?

iajoiner: @tlovell-sxt This code actually requires to use implementation details of both `MontScalar` and `Scalar::MAX_SIGNED`. Hence the macro is needed.

JayWhite2357: You should be able to use `Into<[u64; 4]>` rather than the `MontScalar` specific things.

iajoiner: @JayWhite2357 Yup! Done.

JayWhite2357 (approved):

PR #80, opened 2024-08-03
refactor!: remove `DecompressedCommitment` from `VecCommitmentExt`

JayWhite2357:
# Rationale for this change

In the process of allowing for single table commitments, there is some refactoring needed of the commitment API. This PR simplifies it a bit to ease future PR complexity.

`DecompressedCommitment` was made redundant because every commitment is "decompressed" in the sense that there is no concept of compression in the traits.

This commit simply removes it, which consequently allows for a decent simplification of many of the commitment structures.

# Are these changes tested?

By existing tests

### DIFF
--- a/crates/proof-of-sql/src/sql/proof/proof_builder.rs
+++ b/crates/proof-of-sql/src/sql/proof/proof_builder.rs
@@ -4,7 +4,7 @@ use super::{
 };
 use crate::base::{
     bit::BitDistribution,
-    commitment::{CommittableColumn, VecCommitmentExt},
+    commitment::{Commitment, CommittableColumn},
     polynomial::{CompositePolynomial, MultilinearExtension},
     scalar::Scalar,
 };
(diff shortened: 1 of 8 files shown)


crates/proof-of-sql/src/sql/proof/proof_builder.rs
```diff
@@ -99,16 +99,19 @@ impl<'a, S: Scalar> ProofBuilder<'a, S> {
...
         &self,
         offset_generators: usize,
-        setup: &V::CommitmentPublicSetup<'_>,
-    ) -> V {
-        V::from_commitable_columns_with_offset(
+        setup: &C::PublicSetup<'_>,
+    ) -> Vec<C> {
+        let mut commitments = vec![C::default(); self.commitment_descriptor.len()];
+        C::compute_commitments(
+            &mut commitments,
```
tlovell-sxt, line 109: NIT: is it possible to compute the commitments into a `Vec<C>` directly if you import `VecCommitmentExt`?

```suggestion
        let commitments = Vec::<C>::from_committable_columns_with_offset(
```

tlovell-sxt (dismissed):

tlovell-sxt (approved):

PR #74, opened 2024-07-31
feat: define `Add`, `Sub` and `Mul` for `OwnedColumn`

iajoiner:
# Rationale for this change
This PR is split out from #70 since it has become too large. The actual content of this PR (that is, what's not in #70) is here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/74/commits/e1667622c3db3b45af75dd1f1bf167f3e2a646e5

# What changes are included in this PR?

- Define `Add`, `Sub` and `Mul` for `OwnedColumn`

# Are these changes tested?
Yes.

PR #81, opened 2024-08-04
feat: add `expression_evaluation.rs` and add arithmetic to `proof_of_sql_parser::utility`

iajoiner:
# Rationale for this change
This PR replaces #56 since we made substantial changes to evaluations in #70, #74, #76 and #78. Please review #87 first or look at the second commit.

- add `PostprocessingEvaluator`
- add arithmetic to `proof_of_sql_parser::utility`

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql-parser/src/utility.rs
+++ b/crates/proof-of-sql-parser/src/utility.rs
@@ -53,6 +53,42 @@ pub fn or(left: Box<Expression>, right: Box<Expression>) -> Box<Expression> {
     })
 }
 
+/// Construct a new boxed `Expression` A + B
+pub fn add(left: Box<Expression>, right: Box<Expression>) -> Box<Expression> {
+    Box::new(Expression::Binary {
+        op: BinaryOperator::Add,
+        left,
+        right,
+    })
+}
+
+/// Construct a new boxed `Expression` A - B
+pub fn sub(left: Box<Expression>, right: Box<Expression>) -> Box<Expression> {
+    Box::new(Expression::Binary {
+        op: BinaryOperator::Subtract,
+        left,
+        right,
+    })
+}
+
+/// Construct a new boxed `Expression` A * B
+pub fn mul(left: Box<Expression>, right: Box<Expression>) -> Box<Expression> {
+    Box::new(Expression::Binary {
+        op: BinaryOperator::Multiply,
+        left,
+        right,
+    })
+}
+
+/// Construct a new boxed `Expression` A / B
+pub fn div(left: Box<Expression>, right: Box<Expression>) -> Box<Expression> {
+    Box::new(Expression::Binary {
+        op: BinaryOperator::Division,
+        left,
+        right,
+    })
+}
+
 /// Get table from schema and name.
 ///
 /// If the schema is `None`, the table is assumed to be in the default schema.
--- a/crates/proof-of-sql/src/base/database/expression_evaluation.rs
+++ b/crates/proof-of-sql/src/base/database/expression_evaluation.rs
@@ -1,0 +1,92 @@
+use super::{ExpressionEvaluationError, ExpressionEvaluationResult};
+use crate::base::{
+    database::{OwnedColumn, OwnedTable},
+    math::decimal::{try_into_to_scalar, Precision},
+    scalar::Scalar,
+};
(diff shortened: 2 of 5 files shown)


crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+
+    fn visit_column(&self, identifier: Identifier) -> PostprocessingResult<OwnedColumn<S>> {
+        Ok(self
+            .table
+            .inner_table()
+            .get(&identifier)
+            .ok_or(PostprocessingError::ColumnNotFound(identifier.to_string()))?
+            .clone())
```
JayWhite2357, line 56: I don't like this `clone`, but I'm not sure if there is an easy way to avoid it.

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,252 @@
...
+                                let lhs = left.scalar_at(i).unwrap() * lhs_upscale_factor;
+                                let rhs = right.scalar_at(i).unwrap() * rhs_upscale_factor;
+                                lhs.signed_cmp(&rhs) != Ordering::Greater
+                            })
+                            .collect();
+                        Ok(OwnedColumn::Boolean(res))
+                    }
+                    _ => unreachable!(),
+                }
+            }
```
JayWhite2357, line 173: Can we leave this as `todo!()` for now and then add it back in a manner similar to the arithmetic. (i.e. don't do unneeded conversions.)

JayWhite2357, line 173: Two notes here.
1) I think this code should be as simple as `Ok(left.op_eq(&right)?)` or `Ok(column_eq(&left, &right)?)` or something to that effect. I really like the conciseness of the `Add/Subtract/Multiply/Division` branches.
2) This code should be rewritten to be similar to the `Add/Subtract/Multiply/Division` code in the sense that there shouldn't be needless conversions.

For now, maybe just leave it as `todo!()`, and get it with the next PR.

JayWhite2357 (changes requested):

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+    /// Evaluates the given expression on the owned table and returns the result.
+    pub fn evaluate(&self, expr: &Expression) -> PostprocessingResult<OwnedColumn<S>> {
+        self.visit_expr(expr)
+    }
+}
+
+// Private interface
+impl<S: Scalar> PostprocessingEvaluator<S> {
+    fn visit_expr(&self, expr: &Expression) -> PostprocessingResult<OwnedColumn<S>> {
```
JayWhite2357, line 37: Since `PostprocessingEvaluator` is just a wrapper around `OwnedTable`, why not just make it an extension type? Saves a step when calling this code.
```suggestion
pub trait PostprocessingEvaluator {
    type Scalar: Scalar;
    fn evaluate(&self, expr: &Expression) -> PostprocessingResult<OwnedColumn<Self::Scalar>>;
}
impl<S: Scalar> PostprocessingEvaluator for OwnedTable<S> {
    type Scalar = S;
    fn evaluate(&self, expr: &Expression) -> PostprocessingResult<OwnedColumn<S>> {
        visit_expr(self, expr)
    }
}
fn visit_expr<S: Scalar>(
    table: &OwnedTable<S>,
    expr: &Expression,
) -> PostprocessingResult<OwnedColumn<S>> {
```

iajoiner: Good idea. If it is only going to be applied on `OwnedTable` anyway then we may as well put it in `owned_table_evaluation.rs` instead?

crates/proof-of-sql/src/sql/postprocessing/evaluate.rs
iajoiner: @JayWhite2357 I guess we can leave it as is for now?

JayWhite2357 (approved):

PR #82, opened 2024-08-05
fix: timestamp precision

drcapybara:
# Rationale for this change

Timestamp literals were parsed in seconds despite containing up to nanosecond precision. This caused a situation where equality and inequality queries would fail across the board. This is the first small PR to address this issue. Timestamps of varying, user supplied precisions are now parsed correctly.

- [x] cover precision match in literal conversion
- [x] add accessor methods for timestamp  
- [x] update test expectations

# Are these changes tested?

yes. next PR is to refactor existing timestamp tests into appropriate ast tests.

### DIFF
--- a/crates/proof-of-sql-parser/src/posql_time/timestamp.rs
+++ b/crates/proof-of-sql-parser/src/posql_time/timestamp.rs
@@ -16,10 +16,9 @@ pub struct PoSQLTimestamp {
 }
 
 impl PoSQLTimestamp {
-    /// Returns the number of non-leap seconds since
-    /// January 1, 1970 0:00:00 UTC (aka "UNIX timestamp").
-    pub fn timestamp(&self) -> i64 {
-        self.timestamp.timestamp()
+    /// Returns the combined date and time with time zone.
+    pub fn timestamp(&self) -> DateTime<Utc> {
+        self.timestamp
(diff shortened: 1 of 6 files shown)


crates/proof-of-sql-parser/src/posql_time/timestamp.rs
```diff
@@ -6,16 +6,31 @@ use serde::{Deserialize, Serialize};
...
-    pub timestamp: DateTime<Utc>,
+    timestamp: DateTime<Utc>,
 
     /// The precision of the datetime value, e.g., seconds, milliseconds.
-    pub timeunit: PoSQLTimeUnit,
+    timeunit: PoSQLTimeUnit,
 
     /// The timezone of the datetime, either UTC or a fixed offset from UTC.
-    pub timezone: PoSQLTimeZone,
+    timezone: PoSQLTimeZone,
```
iajoiner, line 15: Why did you do that? Getters and setters are usually boilerplate.

drcapybara: so that the fields cannot be changed after they are set. i.e. a timestamp parsed as seconds should not be able to be changed to milliseconds without the appropirate scaling

PR #48, opened 2024-07-11
feat: add `SelectExpr` and `PostprocessingEvaluator`

iajoiner:
# Rationale for this change
Evaluation of `Expression` on `OwnedTable`s is crucial in native Rust postprocessing of not-yet-provable `Expression`. 

# What changes are included in this PR?
- add `SelectExpr`
- add `PostprocessingEvaluator`
- add division utils

# Are these changes tested?
Will be.

PR #83, opened 2024-08-05
feat: add `SelectExpr`

iajoiner:
# Rationale for this change
This is the last PR split out of #48. Please review #87 and then #81 first or look at the third commit.

- add `SelectExpr`

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/postprocessing/select_expr.rs
+++ b/crates/proof-of-sql/src/sql/postprocessing/select_expr.rs
@@ -1,0 +1,41 @@
+use super::{PostprocessingResult, PostprocessingStep};
+use crate::base::{
+    database::{OwnedColumn, OwnedTable},
+    scalar::Scalar,
+};
+use indexmap::IndexMap;
+use proof_of_sql_parser::{intermediate_ast::AliasedResultExpr, Identifier};
+use serde::{Deserialize, Serialize};
+
+/// The select expression used to select, reorder, and apply alias transformations
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct SelectExpr {
+    /// The aliased result expressions we select
+    aliased_result_exprs: Vec<AliasedResultExpr>,
+}
+
+impl SelectExpr {
+    /// Create a new `SelectExpr` node.
+    pub fn new(aliased_result_exprs: Vec<AliasedResultExpr>) -> Self {
+        Self {
+            aliased_result_exprs,
+        }
+    }
+}
+
+impl<S: Scalar> PostprocessingStep<S> for SelectExpr {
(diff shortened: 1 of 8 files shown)


crates/proof-of-sql/src/sql/postprocessing/select_expr.rs
```diff
@@ -0,0 +1,44 @@
...
+use proof_of_sql_parser::{intermediate_ast::AliasedResultExpr, Identifier};
+use serde::{Deserialize, Serialize};
+
+/// The select expression used to select, reorder, and apply alias transformations
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct SelectExpr<S: Scalar> {
+    /// The aliased result expressions we select
+    aliased_result_exprs: Vec<AliasedResultExpr>,
+    _phantom: core::marker::PhantomData<S>,
+}
```
JayWhite2357, line 16: This should work too, right? We might need it later, but maybe not.
```suggestion
pub struct SelectExpr {
    /// The aliased result expressions we select
    aliased_result_exprs: Vec<AliasedResultExpr>,
}
```

crates/proof-of-sql/src/sql/postprocessing/select_expr.rs
```diff
@@ -0,0 +1,41 @@
...
+            .aliased_result_exprs
+            .iter()
+            .map(
+                |aliased_result_expr| -> PostprocessingResult<(Identifier, OwnedColumn<S>)> {
+                    let result_column = owned_table.evaluate(&aliased_result_expr.expr)?;
+                    Ok((aliased_result_expr.alias, result_column))
+                },
+            )
+            .collect::<PostprocessingResult<_>>()?;
+        Ok(OwnedTable::try_new(cols)?)
```
JayWhite2357, line 39: NIT: using `OwnedTable::try_from_iter` is a bit cleaner since you don't have to worry about building the internal `IndexMap`.
```suggestion
        Ok(self
            .aliased_result_exprs
            .iter()
            .map(|aliased_result_expr| {
                PostprocessingResult::Ok((
                    aliased_result_expr.alias,
                    owned_table.evaluate(&aliased_result_expr.expr)?,
                ))
            })
            .process_results(|iter| OwnedTable::try_from_iter(iter))??)
```

JayWhite2357 (approved):

crates/proof-of-sql/src/sql/postprocessing/select_expr.rs
iajoiner: @JayWhite2357 Sure! I think we might need to do a revision for group by so we may as well merge this first.

PR #94, opened 2024-08-08
feat: add `MAX` and `MIN` to `AggregateColumns`

iajoiner:
# Rationale for this change
We need to add `max` and `min` to `AggregateColumns` for postprocessing even though we don't make them provable just yet.

- add max and min to `AggregateColumns`

# Are these changes tested?
Yes.

### DIFF
--- a/crates/proof-of-sql/src/base/database/group_by_util_test.rs
+++ b/crates/proof-of-sql/src/base/database/group_by_util_test.rs
@@ -18,7 +18,7 @@ fn we_can_aggregate_empty_columns() {
     let sum_columns = &[column_c.clone(), column_d.clone()];
     let selection = &[];
     let alloc = Bump::new();
-    let aggregate_result = aggregate_columns(&alloc, group_by, sum_columns, selection)
+    let aggregate_result = aggregate_columns(&alloc, group_by, sum_columns, &[], &[], selection)
         .expect("Aggregation should succeed");
     assert_eq!(
         aggregate_result.group_by_columns,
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/base/database/group_by_util_test.rs
```diff
@@ -289,7 +420,39 @@ fn we_can_sum_aggregate_slice_by_counts_with_empty_result() {
...
+    let expected = &[
+        Curve25519Scalar::from(112 + 111 + 101),
+        Curve25519Scalar::from(110 + 102 + 103 + 104),
+        Curve25519Scalar::from(0),
+    ];
+    let alloc = Bump::new();
+    let result: &[Curve25519Scalar] =
+        sum_aggregate_slice_by_index_counts(&alloc, slice_a, counts, indexes);
+    assert_eq!(result, expected);
+}
```
JayWhite2357, line 452: These should never actually occur since empty groups would not be included in the result.
The tests are ok since they are testing the function, but in a sense this could be undefined behavior of the function.

crates/proof-of-sql/src/base/database/group_by_util_test.rs
```diff
@@ -335,11 +498,12 @@ fn we_can_sum_aggregate_columns_by_counts() {
     let columns_b = Column::Int128::<Curve25519Scalar>(slice_b);
     let columns_c = Column::Scalar(&scals_c);
     let indexes = &[12, 11, 1, 10, 2, 3, 6, 14, 13, 9];
-    let counts = &[3, 3, 4];
+    let counts = &[3, 3, 4, 0];
     let expected = &[
         Curve25519Scalar::from(112 + 111 + 101),
         Curve25519Scalar::from(110 + 102 + 103),
         Curve25519Scalar::from(106 + 114 + 113 + 109),
+        Curve25519Scalar::from(0),
```
JayWhite2357, line 506: Can we keep this test as it was since this behavior is unneeded and tested above?

JayWhite2357 (approved):

PR #95, opened 2024-08-09
refactor: separate OffsetToBytes trait from transpose module (PROOF-897)

jacobtrombetta:
# Rationale for this change
In an effort to support Blitzar's new `packed_msm` interface, we need to make a number of updates to the Dory commitment computation. A future PR will require the use of the `OffsetToBytes` trait. This PR separates the trait into a different file to make future usage easier.

- The `OffsetToBytes` trait is moved to its own file.

# Are these changes tested?
Yes.


crates/proof-of-sql/src/proof_primitive/dory/dory_commitment_helper_gpu.rs
```diff
@@ -94,7 +96,7 @@ fn compute_dory_commitment_impl<'a, T>(
 where
     &'a T: Into<DoryScalar>,
     &'a [T]: Into<Sequence<'a>>,
-    T: AsBytes + Copy + transpose::OffsetToBytes,
+    T: AsBytes + Copy + OffsetToBytes,
```
JayWhite2357, line 99: Can you also change this to simply be:
```suggestion
    &'a [T]: Into<Sequence<'a>>,
    T: OffsetToBytes,
```

crates/proof-of-sql/src/proof_primitive/dory/offset_to_bytes.rs
```diff
@@ -0,0 +1,110 @@
+use ark_bls12_381::Fr;
```
JayWhite2357, line 1:
```suggestion
use super::F;
```

JayWhite2357 (changes requested):

crates/proof-of-sql/src/proof_primitive/dory/offset_to_bytes.rs
jacobtrombetta: Updated, thank you.

crates/proof-of-sql/src/proof_primitive/dory/dory_commitment_helper_gpu.rs
jacobtrombetta: Removing both `T: AsBytes + Copy` causes the compiler to complain. They are still needed because the trait `std::marker::Copy` and `AsBytes` is not implemented for `T` on the `column` parameter. It is required by a bound introduced by `transpose::transpose_for_fixed_msm` call. I will be removing this function in a follow up PR. Do you think it's worth figuring out a fix for this PR?

jacobtrombetta: Thanks for the help, it is removed!

JayWhite2357 (approved):

iajoiner (approved):

PR #96, opened 2024-08-09
test: add mixed committable column tests to the Dory compute commitment test class (PROOF-898)

jacobtrombetta:
# Rationale for this change
In an effort to support Blitzar's new `packed_msm` interface, we need to make a number of updates to the Dory commitment computation. This PR adds additional tests for committable columns of various types. These tests will help ensure the data packing done for the `packed_msm` interface is done correctly for all supported data types.

- Five tests have been added to the Dory compute commitment test class.

# Are these changes tested?
Yes


JayWhite2357 (approved):
Looks good. Thanks, Jacob.

iajoiner (approved):

PR #99, opened 2024-08-10
refactor!: replace `HashSet` and `HashMap` with `IndexSet` and `IndexMap`

JayWhite2357:
# Rationale for this change

There are several locations where a well-defined order is very important. We should be using the more predictable `IndexMap` over a `HashMap`.

See rationale. This is technically a breaking change.

# Are these changes tested?

Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/ast/column_expr.rs
+++ b/crates/proof-of-sql/src/sql/ast/column_expr.rs
@@ -8,9 +8,10 @@ use crate::{
     sql::proof::{CountBuilder, ProofBuilder, VerificationBuilder},
 };
 use bumpalo::Bump;
+use core::marker::PhantomData;
+use indexmap::IndexSet;
 use proof_of_sql_parser::Identifier;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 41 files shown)


crates/proof-of-sql/src/sql/ast/column_expr.rs
```diff
@@ -8,9 +8,10 @@ use crate::{
     sql::proof::{CountBuilder, ProofBuilder, VerificationBuilder},
 };
 use bumpalo::Bump;
+use indexmap::IndexSet;
 use proof_of_sql_parser::Identifier;
 use serde::{Deserialize, Serialize};
-use std::{collections::HashSet, marker::PhantomData};
+use std::marker::PhantomData;
```
iajoiner, line 14: Maybe we should use `core`?

iajoiner (approved):

iajoiner (approved):

iajoiner (dismissed):
Approved again. You might want to get rid of the other two instances of `std::marker` but otherwise it is good.

iajoiner (approved):
Really thanks!

PR #100, opened 2024-08-10
refactor: simplify `QueryProof::new`

JayWhite2357:
# Rationale for this change

The `QueryProof::new_from_builder` method is only used in one place: a test. It adds non-trivial complexity to modifying `QueryProof`.

* This commit refactors the offending test by allowing `sign_expr` to, optionally, allow a column of zeros to be considered negative.
* `QueryProof::new_from_builder` is inlined into `QueryProof::new`.

# Are these changes tested?

Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/ast/inequality_expr.rs
+++ b/crates/proof-of-sql/src/sql/ast/inequality_expr.rs
@@ -22,12 +22,20 @@ pub struct InequalityExpr<C: Commitment> {
     lhs: Box<ProvableExprPlan<C>>,
     rhs: Box<ProvableExprPlan<C>>,
     is_lte: bool,
+    #[cfg(test)]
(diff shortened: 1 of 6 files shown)


crates/proof-of-sql/src/sql/ast/inequality_expr.rs
```diff
@@ -22,12 +22,20 @@ pub struct InequalityExpr<C: Commitment> {
     lhs: Box<ProvableExprPlan<C>>,
     rhs: Box<ProvableExprPlan<C>>,
     is_lte: bool,
+    #[cfg(test)]
```
iajoiner, line 25: Please add comments for it is not clear what `zero_sign` means. Alternatively something like `treat_zeros_as_negative` is more descriptive.

iajoiner (approved):

PR #115, opened 2024-08-20
bench: append rows to tables and update commitments

drcapybara:
# Rationale for this change

Adds a benchmark for appending various numbers of rows to 10 columns of mixed data.

# Included changes

Adds benches and a random owned table generation utility


crates/proof-of-sql/benches/bench_append_rows.rs
```diff
@@ -0,0 +1,185 @@
...
+                    black_box(append_columns.inner_table()),
+                    &black_box(dory_prover_setup),
+                )
+                .unwrap();
+        });
+    });
+}
+
+// append 1000 rows to 10 cols in 1 table in 652ms
+fn bench_append_rows_10x1000(c: &mut Criterion) {
```
iajoiner, line 62: Can we make this and the previous function the same function with the number of cols, rows etc parameters?

iajoiner (changes requested):

crates/proof-of-sql/Cargo.toml
```diff
@@ -87,6 +87,11 @@ name = "criterion_benches"
 harness = false
 required-features = [ "blitzar" ]
 
+[[bench]]
```
iajoiner, line 90: We need to rename the other criterion bench then since `bench_append_rows` is also criterion.

crates/proof-of-sql/benches/bench_append_rows.rs
```diff
@@ -0,0 +1,170 @@
...
+/// data types, and choose the number of rows to append at a time.
+///
+/// ```text
+/// Most recent benches on 13th Gen Intel® Core™ i9-13900H × 20 with 32gb of RAM:
+/// append 10 rows to 10 cols in 1 table = 11.382 ms
+/// append 10 rows to 10 cols in 100 tables = 1.1382 seconds
+/// append 1000 rows to 10 cols in 1 table = 652ms
+/// ```
+fn bench_append_rows(c: &mut Criterion, cols: usize, rows: usize) {
+    let public_parameters = PublicParameters::rand(10, &mut test_rng());
```
iajoiner, line 41: Are you sure this is not `cols`?

iajoiner (approved):

PR #117, opened 2024-08-23
chore: disable some unused default features

jvff:
# Rationale for this change

To build this crate for `wasm32-unknown-unknown` to run on runtimes with a limited custom API available (e.g., Linera contracts), `wasm-bindgen` should not be a dependency. That's because `wasm-bindgen` generates bindings for the Javascript APIs available in browsers (and similar runtimes). For more bare-bones runtimes, these APIs aren't available.

Disabling the `default` features of `chrono`, `rand` and `rand-core`.

Unfortunately, this doesn't complete the work to support `wasm32-unknown-unknown` without JS, because a fix is also needed in `merlin` (https://github.com/zkcrypto/merlin/pull/8).

# Are these changes tested?

Tested with `cargo test`.


JayWhite2357:
Thanks @jvff! Looks good.

Extra note: We will likely be removing merlin from the dependencies in the near future, so `wasm32-unknown-unknown` may work once that gets in.

JayWhite2357 (approved):

PR #118, opened 2024-08-28
feat: range check

drcapybara:
# Rationale for this change

Adds an enhanced range check to the SxT PoSQL proof framework. Should improve range checks for prover by operating over words instead of individual bits.

## Strategy:

Our approach is to use the approach outlined in the paper [Multivariate lookups based on logarithmic derivatives](https://eprint.iacr.org/2022/1530.pdf).

 The general idea is to decomposes a column of scalars into a matrix of words, so that each word column can be
 used to produce an intermediate MLE. Produces intermediate MLEs for:
 * each column of words
 * the count of how many times each word occurs
 * a column from 0 to the max value supported by a word
 * each column of the inversion of word + verifier challenge alpha
 * a column of the inversion of the word values + verifier challenge alpha

 And anchored MLEs for:
 * all possible byte values

 ## Word-sized decomposition:

Example: each row represents the byte decomposition of a scalar, and each column contains the bytes from
 the same byte position across all scalars:

 ```text
 | Column 0           | Column 1           | Column 2           | ... | Column 30           |  
 |--------------------|--------------------|--------------------|-----|---------------------|  
 | Byte 0 of Scalar 0 | Byte 1 of Scalar 0 | Byte 2 of Scalar 0 | ... | Byte 30 of Scalar 0 |  
 | Byte 0 of Scalar 1 | Byte 1 of Scalar 1 | Byte 2 of Scalar 1 | ... | Byte 30 of Scalar 1 |  
 | Byte 0 of Scalar 2 | Byte 1 of Scalar 2 | Byte 2 of Scalar 2 | ... | Byte 30 of Scalar 2 |  
 ```
 After constructing this matrix, each byte column is used to produce an intermediate MLE, anchored MLE for data the verifier can access, and sumcheck subpolynomials for establishing the constraints and sumcheck subpolynomials.

Word matrix, word value column, and word occurrence column

- [x] produce word columns
- [x] produce word value columns
- [x] produce count of word occurrences across entire word matrix
- [x] produce intermediate mles for word columns
- [x] produce intermediate mles for word value columns
- [x] produce anchored mles for word value columns
- [x] Invert words and perturb with verifier challenge
- [x] Invert word values and perturb with verifier challenge
- [x] Invert word columns and perturb with verfier challenge
- [x] Produce mles for inverted perturbed words
- [x] Produce mles for inverted perturbed word values
- [x] Produce mles for inverted perturbed word counts

Current working items:

- [ ] Cache frequently used columns such as decomposed words and their inversions
- [ ] Add sumcheck subpolynomials for identity constraints
- [ ] Add sumcheck subpolynomials for sum constraints
- [ ] Ensure that anything allocated with bump is dropped correctly and safely
- [ ] byte counts produces sparse vector, make it produce dense vector that is equal in size to length of scalar columns
- [ ] pad scalar vector to be power of 2 in length

# Are these changes tested?

## Tests
- [ ] test scalars to words are producing expected results
- [ ] test that inverted words are expected
- [ ] test that sum of counts of word values = number of columns * counts of word values

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/column_bounds.rs
+++ b/crates/proof-of-sql/src/base/commitment/column_bounds.rs
@@ -228,6 +228,7 @@ impl ColumnBounds {
             | CommittableColumn::Decimal75(_, _, _)
             | CommittableColumn::Scalar(_)
             | CommittableColumn::VarChar(_) => ColumnBounds::NoOrder,
+            CommittableColumn::RangeCheckWord(_) => ColumnBounds::NoOrder,
         }
     }
 
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -40,6 +40,8 @@ pub enum CommittableColumn<'a> {
     VarChar(Vec<[u64; 4]>),
     /// Borrowed Timestamp column with Timezone, mapped to `i64`.
     TimestampTZ(PoSQLTimeUnit, PoSQLTimeZone, &'a [i64]),
+    /// Borrowed u8 column for range-check words, mapped to `u64`.
+    RangeCheckWord(&'a [u8]),
 }
 
 impl<'a> CommittableColumn<'a> {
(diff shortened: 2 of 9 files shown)


drcapybara:
A question I might have right now for this is what objects are the words?
- Scalars?
- bytes?
- u64 mod 256?
-something else?

PR #120, opened 2024-08-28
ci: add CI and basic tests for Solidity

JayWhite2357:
# Rationale for this change

In order to facilitate development of a Solidity verifier, we should be running unit tests within the CI.

* `forge-script` is used to call solidity code from rust. This enables writing tests where the test cases are generated by Rust. For example, this is helpful when trying to test that a proof is verified. The proof can be generated by rust, and verified in solidity. Because these are included in the Rust test suite, nothing is added in CI to achieve this. Some example tests are added.
* A `foundry.toml` project file is added to the top directory and `forge test -vvv` is now in CI in order to run pure Solidity tests (functions that start with `test`)
* `solhint` is now part of CI in order to enforce linting of Solidity code.

# Are these changes tested?

Yes
* The example Rust+Solidity tests execute as expected.
* `forge test -vvv` fails on failed tests and shows up as failed in the CI: https://github.com/spaceandtimelabs/sxt-proof-of-sql/actions/runs/10598506903/job/29371258834
* `solhint` fails when a lint is cause and shows up as failed in the CI: https://github.com/spaceandtimelabs/sxt-proof-of-sql/actions/runs/10598506903/job/29371261143


crates/proof-of-sql/src/tests/TestScript.t.sol
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    error DummyError();
+
+    function rustTestWeCanThrowErrorDependingOnParameter(uint256 x) public pure {
+        if (x != 1234) revert DummyError();
+    }
+
+    function rustTestWeCanDecodeCustomStruct(bytes memory x) public pure {
+        (CustomStruct memory y) = abi.decode(x, (CustomStruct)); // solhint-disable-line no-unused-vars
```
waelsy123, line 16: we can decode any struct or type but solidity should be aware of the type (CustomStruct here). unfortunately we can't pass the type as parameter via tx

waelsy123 (dismissed):

crates/proof-of-sql/src/tests/TestScript.t.sol
JayWhite2357: Renamed the method. The function is used to demonstrate that we can, in effect, pass a custom struct from Rust to Solidity. Didn't realize a tx couldn't handle a type as a parameter.

iajoiner (commented):
I wonder whether the Solidity work should be in a separate crate since it is quite independent from the rest of the repo and has unique dependencies.

waelsy123 (dismissed):

JayWhite2357:
> I wonder whether the Solidity work should be in a separate crate since it is quite independent from the rest of the repo and has unique dependencies.

I was questioning this, and if you have a better way, we should do it. This is the best solution I could come up with.

Although the code is independent, it's not fully independent. For example, I want to run a test of just sumcheck:
```rust
fn we_can_verify_proof_with_solidity() {
    // Setup
    ...
    // Generate proof
    let proof = ...;
    // Verify proof in Rust
    ...
    // Verify proof in Solidity
    let args = vec![
        proof,
        ...
    ];
    ScriptArgs {
        path: "./src/proof_primitive/sumcheck/Sumcheck.t.sol".to_string(),
        sig: "rustTestVerification".to_string(),
        args,
        ..Default::default()
    }
    .run_script()
    .await
    .unwrap();
}
```

We also want to ensure that the Solidity verifier is identical to the Rust verifier, and actually is compatible with the Rust prover. Perhaps we can make is a separate crate, but the problem is that sumcheck, for example, is private.

Another example is `compute_truncated_lagrange_basis_sum`, which we will need to port to Solidity. We almost certainly will want tests to ensure that the behavior is identical between the two languages. While we can manually ensure that we have identical tests for both languages, I prefer to have unit tests actually compare the inputs/outputs.

iajoiner (approved):

PR #123, opened 2024-08-30
feat: add simplified `Transcript` trait and implementations

JayWhite2357:
# Rationale for this change

In order to write the verifier in Solidity, we should enable using a Solidity compatibly public-coin transcript. This PR adds an API for doing so.

* added simplified `Transcript` trait
* implemented `Transcript` trait for `merlin::Transcript`
* added Solidity compatible `Keccak256Transcript`
* added tests for `Transcript` and implementations

# Are these changes tested?

Yes

### DIFF
--- a/crates/proof-of-sql/src/base/proof/transcript.rs
+++ b/crates/proof-of-sql/src/base/proof/transcript.rs
@@ -13,7 +13,10 @@ pub trait Transcript {
     /// Appends the provided messages by appending the reversed raw bytes (i.e. assuming the message is bigendian)
     fn extend_as_be<M: FromBytes + AsBytes>(&mut self, messages: impl IntoIterator<Item = M>);
     /// Appends the provided messages by appending the raw bytes (i.e. assuming the message is littleendian)
-    fn extend_as_le<'a, M: AsBytes + 'a>(&mut self, messages: impl IntoIterator<Item = &'a M>);
+    fn extend_as_le_from_refs<'a, M: AsBytes + 'a>(
+        &mut self,
+        messages: impl IntoIterator<Item = &'a M>,
+    );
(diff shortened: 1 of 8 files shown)


crates/proof-of-sql/src/base/proof/transcript.rs
```diff
@@ -0,0 +1,27 @@
...
+///
+/// Implementation note: this is intended to be implemented via [super::transcript_core::TranscriptCore] rather than directly.
+#[allow(dead_code)]
+pub trait Transcript {
+    /// Creates a new transcript
+    fn new() -> Self;
+    /// Appends the provided messages by appending the reversed raw bytes (i.e. assuming the message is bigendian)
+    fn extend_as_be<M: FromBytes + AsBytes>(&mut self, messages: impl IntoIterator<Item = M>);
+    /// Appends the provided messages by appending the raw bytes (i.e. assuming the message is littleendian)
+    fn extend_as_le<'a, M: AsBytes + 'a>(&mut self, messages: impl IntoIterator<Item = &'a M>);
```
iajoiner, line 16: Hmm asymmetry between APIs for be and le do not look intuitive.

JayWhite2357: Yeah, I know. I think I need to rename them. Suggestions?

The reason for the asymmetry is because `le` can be by reference because it is just looking at the raw bytes.
`be` needs to either copy or own the data since it has to reverse the bytes. I prefer forcing the caller to copy, so it needs to own.

Additionally, these are the signatures that I am using downstream.

iajoiner: What about `reverse_and_extend_as_be` / `reverse_and_extend_scalars_as_be`?

iajoiner (approved):

PR #127, opened 2024-09-05
perf: parallelize v2 construction within `eval_vmv_re_prove`

JayWhite2357:
# Rationale for this change

This reduces the `eval_vmv_re_prove` component of the proof time from one of the larger factors to a negligible amount.

The computation of v2 within `eval_vmv_re_prove` now uses a rayon iterator.

# Are these changes tested?

Yes


jacobtrombetta (approved):
LGTM

PR #128, opened 2024-09-05
perf: improve `offset_to_bytes` performance

JayWhite2357:
# Rationale for this change

This PR reduces the cost of packing bytes in order to set up a packed MSM.

`offset_to_bytes` now returns an array rather than allocating a `Vec`.

# Are these changes tested?

Yes


jacobtrombetta (approved):
LGTM

PR #130, opened 2024-09-05
perf: improve signed value offset handling when packing scalars (PROOF-905)

jacobtrombetta:
# Rationale for this change
Blitzar's `packed_msm`, like `fixed_msm`, does not handle signed values. To compute commitments with signed columns we offset signed values by subtracting the minimum of the data type. This requires us to compute the commitments to the minimum value of the data type and add its commitment before pairing.

The way we currently handle the offsets is by duplicating the `commitable_columns` and packing in 8 bit `1`s for every column with a value. The problem with this approach is that we compute a lot of duplicate commits.

This PR addresses this issue by condensing all the duplicate commits. The bit table now has an 8 bit value added to handle the offset, full commit (all `1`s), and the last commit for each commitable column.

## Benchmarks
The following benchmarks use Jaeger benchmarks with default values.

Multi A100
Current main
image

This PR
image

Single A100
Current main
image

This PR
image

- Duplicate offsets of all `1`s are eliminated.
- Offsets no longer double the size of the `packed_scalars` array.
- `par_iter` is removed from the the `bit_table_and_scalars_for_packed_msm` function. Benchmarks showed it was slowing down computation where applied.
- `is_signed` is added to the column module.
- General refactoring for improving loop efficiency.

# Are these changes tested?
Yes


crates/proof-of-sql/src/proof_primitive/dory/dory_commitment_helper_gpu.rs
```diff
@@ -14,33 +14,31 @@ fn compute_dory_commitments_packed_impl(
...
     setup: &DoryProverPublicSetup,
 ) -> Vec<DoryCommitment> {
+    // Make sure that the committable columns are not empty.
     if committable_columns.is_empty() {
         return vec![];
     }
 
+    // Set the parameters.
     let num_columns = 1 << setup.sigma();
+    let gamma_2 = setup.prover_setup().Gamma_2.clone();
```
JayWhite2357, line 24: This should not be cloned.

JayWhite2357 (dismissed):

crates/proof-of-sql/src/proof_primitive/dory/dory_commitment_helper_gpu.rs
jacobtrombetta: Updated, thank you.

JayWhite2357 (approved):

PR #131, opened 2024-09-06
refactor: Replace scalar_conversion_to_int with blanket implementation

stuarttimwhite:
# Rationale for this change

Currently any MontScalar that implements Scalar must call the macro scalar_conversion_to_int to implement various functions that should be available automatically. These functions should be available by default. 

1. Removal of scalar_conversion_to_int and any references.
2. All functions previously implemented in scalar_conversion_to_int have been implemented generically for any MontScalar that implements Scalar.

# Are these changes tested?

No tests included. This is a refactor with no functional change.


drcapybara (approved):
LGTM - Thanks for doing this!

PR #133, opened 2024-09-06
feat: commitable byte columns and unconstrained range check

drcapybara:
# Rationale for this change

Support for a range check over words decomposed byte-wise requires an ``` &[u8]``` to the commitable column. This PR introduces this small change, as well as preliminary support for a range check. It is **unsound** in its current form and does not constrain the claims of integrity made by the prover. 

- [x] support byte-wise word decomposition and commitment
- [x] produce word columns
- [x] produce word value columns
- [x] produce count of word occurrences across entire word matrix
- [x] produce intermediate mles for word columns
- [x] produce intermediate mles for word value columns
- [x] produce anchored mles for word value columns
- [x] Invert words and perturb with verifier challenge
- [x] Invert word values and perturb with verifier challenge
- [x] Invert word columns and perturb with verfier challenge
- [x] Produce mles for inverted perturbed words
- [x] Produce mles for inverted perturbed word values
- [x] Produce mles for inverted perturbed word counts

# Are these changes tested?

- [x] test scalars to words are producing expected results
- [x] test that inverted words match expected
- [ ] test that the prover is producing the correct MLEs? How do we do this?

# Split:

Current working items:

- [ ] Cache frequently used columns such as decomposed words and their inversions
- [ ] Add sumcheck subpolynomials for identity constraints
- [ ] Add sumcheck subpolynomials for sum constraints
- [ ] Ensure that anything allocated with bump is dropped correctly and safely
- [x] byte counts produces sparse vector, make it produce dense vector that is equal in size to length of scalar columns
- [ ] pad scalar vector to be power of 2 in length
- [ ] test that sum of counts of word values = number of columns * counts of word values

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
@@ -15,17 +15,8 @@ use bumpalo::Bump;
 /// ## Word-sized decomposition:
 ///
 /// Each row represents the byte decomposition of a scalar, and each column contains the bytes from
-/// the same byte position across all scalars. First, we produce this word-wise decomposition:
-///
-/// ```text
-/// | Column 0           | Column 1           | Column 2           | ... | Column 31           |  
-/// |--------------------|--------------------|--------------------|-----|---------------------|  
-/// | Byte 0 of Scalar 0 | Byte 1 of Scalar 0 | Byte 2 of Scalar 0 | ... | Byte 31 of Scalar 0 |  
-/// | Byte 0 of Scalar 1 | Byte 1 of Scalar 1 | Byte 2 of Scalar 1 | ... | Byte 31 of Scalar 1 |  
-/// | Byte 0 of Scalar 2 | Byte 1 of Scalar 2 | Byte 2 of Scalar 2 | ... | Byte 31 of Scalar 2 |  
-/// ```
-///
-/// The next step is to compute intermediate MLEs over the word columns:
+/// the same byte position across all scalars. First, we produce this word-wise decomposition,
+/// as well as computing intermediate MLEs over the word columns:
 ///
 /// ```text
 /// | Column 0           | Column 1           | Column 2           | ... | Column 31           |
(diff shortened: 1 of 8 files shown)


crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,137 @@
...
+        for (col, &byte) in columns.iter_mut().zip(scalar_bytes.iter()) {
+            col[i] = byte;
+
+            // Update the byte count in the corresponding position
+            byte_counts[byte as usize] += S::one(); // Increment the count of the byte value
+        }
+    }
+
+    // Allocate and initialize byte_values to represent each possible byte as a scalar directly
+    let byte_values: &mut [S] =
```
JayWhite2357, line 61: NIT: This could just be a slice of `u8`s since the values will fit in that range.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    // 2. Produce the anchored MLE that the verifier has access to, consisting
+    // of all possible word values. These serve as values to lookup
+    // in the lookup table
+    builder.produce_anchored_mle(byte_values as &[S]);
+
+    // 3. Next produce an MLE over the counts of each word value
+    builder.produce_intermediate_mle(byte_counts as &[S]);
+
+    // Invert the scalars, and get the inverted words.
```
JayWhite2357, line 77: The "inverted" words are not the words of the inverted scalars. Instead, they are the "inverses" (really log-derivative) of the individual words. In other words, we should have
```
inverted_word_columns[i][j] == (column[i][j] + alpha).inv()
```

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,137 @@
...
+            col[i] = value;
+            *row_entry = value;
+        }
+        builder.produce_intermediate_mle(&*row);
+    }
+
+    // Now produce an intermediate MLE over the inverted word values + verifier challenge alpha
+    let inverted_word_values: &mut [S] =
+        alloc.alloc_slice_fill_with(256, |i| S::try_from(i.into()).unwrap() + alpha);
+    slice_ops::batch_inversion(&mut inverted_word_values[..]);
```
JayWhite2357, line 112: This is correct.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,137 @@
...
+    builder.produce_anchored_mle(inverted_word_values as &[S]);
```
JayWhite2357, line 113: This is not anchored, since the verifier doesn't have the commitment to it (because it doesn't know alpha ahead of time.)

PR #137, opened 2024-09-09
refactor!: remove old `FilterExec` and `FilterResultExpr` and rename `DenseFilterExec` to `FilterExec`

iajoiner:
# Rationale for this change
`FilterExec` is no longer used and it is increasingly costly to maintain it. Now it is time to finally remove it.

- Remove `FilterExec` and `FilterResultExpr`
- Rename `DenseFilterExec` to `FilterExec`

# Are these changes tested?
Existing tests do pass


crates/proof-of-sql/src/sql/proof_exprs/inequality_expr_test.rs
```diff
@@ -477,28 +477,6 @@ fn we_can_compare_a_constant_column_of_zeros() {
...
-fn the_sign_can_be_0_or_1_for_a_constant_column_of_zeros() {
-    let data = owned_table([bigint("a", [0_i64, 0, 0]), bigint("b", [1_i64, 2, 3])]);
-    let t = "sxt.t".parse().unwrap();
-    let accessor = OwnedTableTestAccessor::<InnerProductProof>::new_from_table(t, data, 0, ());
-    let mut ast = filter(
-        cols_result(t, &["b"], &accessor),
-        tab(t),
-        lte(column(t, "a", &accessor), const_bigint(0)),
-    );
-    if let DynProofPlan::Filter(filter) = &mut ast {
```
JayWhite2357, line 490: Can you replace this with a `DenseFilter` rather than just removing it?

crates/proof-of-sql/src/sql/proof_plans/dyn_proof_plan.rs
```diff
@@ -43,7 +38,6 @@ impl<C: Commitment> ProofPlan<C> for DynProofPlan<C> {
     ) -> Result<(), crate::base::proof::ProofError> {
         match self {
             DynProofPlan::Projection(expr) => expr.count(builder, accessor),
-            DynProofPlan::Filter(expr) => expr.count(builder, accessor),
             DynProofPlan::GroupBy(expr) => expr.count(builder, accessor),
             DynProofPlan::DenseFilter(expr) => expr.count(builder, accessor),
```
JayWhite2357, line 42: Can we rename `DenseFilter` to `Filter` in this PR as well? Keep this PR as one of the commits, and then do the rename in another commit (or two).
That way, there is only one breaking change.

JayWhite2357 (changes requested):

crates/proof-of-sql/src/sql/proof_exprs/inequality_expr_test.rs
iajoiner: Yup.

JayWhite2357 (approved):
Thanks, Ian!

PR #141, opened 2024-09-10
feat: range check: decompose to words and get log derivative

drcapybara:
# Rationale for this change

This PR brings initial support for the updated range check functionality. It includes compartmentalized functions which handles the decomposition of scalars into words, and it handles the addition of the verifier challenge to each word and inverting the sum, forming the logarithmic derivative of the word and challenge.

- [x] scalar decomposition
- [x] logarithmic derivative
- [x] update to commitable column to support u8

# Are these changes tested?

yes

- [x] we_can_decompose_small_scalars_to_words
- [x] we_can_decompose_large_scalars_to_words
- [x] we_can_obtain_a_logarithmic_derivative_from_a_small_scalar
- [x] we_can_obtain_a_logarithmic_derivative_from_a_large_scalar

# Split
- [ ] if alpha = P -  word, then word + alpha mod P = 0 and has no inverse. This breaks completeness and should be addressed

### DIFF
--- a/crates/proof-of-sql/src/base/scalar/mod.rs
+++ b/crates/proof-of-sql/src/base/scalar/mod.rs
@@ -43,6 +43,7 @@ pub trait Scalar:
     + for<'a> core::convert::From<&'a i32> // Required for `Column` to implement `MultilinearExtension`
     + for<'a> core::convert::From<&'a i64> // Required for `Column` to implement `MultilinearExtension`
     + for<'a> core::convert::From<&'a i128> // Required for `Column` to implement `MultilinearExtension`
+    + for<'a> core::convert::From<&'a u8> // Required for `Column` to implement `MultilinearExtension`
     + core::convert::TryInto <bool>
     + core::convert::TryInto <i8>
     + core::convert::TryInto <i16>
--- a/crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
@@ -33,10 +33,7 @@ fn get_logarithmic_derivative<'a, S: Scalar + 'a>(
     // Iterate over each column
     for (i, byte_column) in byte_columns.iter().enumerate() {
         // Convert bytes to field elements and add alpha
-        let mut terms_to_invert: Vec<S> = byte_column
-            .iter()
-            .map(|&w| S::try_from(w.into()).unwrap() + alpha)
-            .collect();
+        let mut terms_to_invert: Vec<S> = byte_column.iter().map(|w| S::from(w) + alpha).collect();
 
         // Invert all the terms in the column at once
         slice_ops::batch_inversion(&mut terms_to_invert);
--- a/crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
@@ -1,0 +1,282 @@
+use crate::base::{scalar::Scalar, slice_ops};
+use bytemuck::cast_slice;
+
+// Decomposes a scalar to requisite words, additionally tracks the total
+// number of occurences of each word for later use in the argument.
+fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
+    scalars: &mut [S],
+    word_columns: &mut [&mut [u8]],
+    byte_counts: &mut [u64],
+) {
+    for (i, scalar) in scalars.iter().enumerate() {
+        let scalar_array: [u64; 4] = (*scalar).into(); // Convert scalar to u64 array
+        let scalar_bytes_full = cast_slice::<u64, u8>(&scalar_array); // Cast u64 array to u8 slice
+        let scalar_bytes = &scalar_bytes_full[..31];
+
+        // Populate the columns of the words table with decomposition of scalar:
+        for (byte_index, &byte) in scalar_bytes.iter().enumerate() {
+            // Each column in word_columns is for a specific byte position across all scalars
+            word_columns[byte_index][i] = byte;
+            byte_counts[byte as usize] += 1;
+        }
+    }
+}
+
+// For a word w and a verifier challenge alpha, compute
+// 1 / (word + alpha), which is the modular multiplicative
+// inverse of (word + alpha) in the scalar field.
+fn get_logarithmic_derivative<'a, S: Scalar + 'a>(
+    byte_columns: &[&mut [u8]],
+    alpha: S,
+    inverted_word_columns: &mut [&mut [S]],
+) {
+    // Iterate over each column
+    for (i, byte_column) in byte_columns.iter().enumerate() {
+        // Convert bytes to field elements and add alpha
+        let mut terms_to_invert: Vec<S> = byte_column
+            .iter()
+            .map(|&w| S::try_from(w.into()).unwrap() + alpha)
+            .collect();
+
+        // Invert all the terms in the column at once
+        slice_ops::batch_inversion(&mut terms_to_invert);
+
+        // Assign the inverted values back to the inverted_word_columns
+        inverted_word_columns[i].copy_from_slice(&terms_to_invert);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        base::scalar::{Curve25519Scalar as S, Scalar},
+        sql::proof_exprs::range_check::{decompose_scalar_to_words, get_logarithmic_derivative},
+    };
+    use num_traits::Inv;
+
+    #[test]
+    fn we_can_decompose_small_scalars_to_words() {
+        let mut scalars: Vec<S> = [1, 2, 3, 255, 256, 257].iter().map(S::from).collect();
+
+        let mut word_columns = vec![vec![0; scalars.len()]; 31];
+        let mut word_slices: Vec<&mut [u8]> = word_columns.iter_mut().map(|c| &mut c[..]).collect();
+        let mut byte_counts = vec![0; 256];
+
+        decompose_scalar_to_words(&mut scalars, &mut word_slices, &mut byte_counts);
+
+        let mut expected_word_columns = vec![vec![0; scalars.len()]; 31];
+        expected_word_columns[0] = vec![1, 2, 3, 255, 0, 1];
+        expected_word_columns[1] = vec![0, 0, 0, 0, 1, 1];
+        // expected_word_columns[2..] is filled with 0s.
+        let mut expected_byte_counts = vec![0; 256];
+        expected_byte_counts[0] = 31 * 6 - 7;
+        expected_byte_counts[1] = 4;
+        expected_byte_counts[2] = 1;
+        expected_byte_counts[3] = 1;
+        // expected_byte_counts[4..255] is filled with 0s.
+        expected_byte_counts[255] = 1;
+
+        assert_eq!(word_columns, expected_word_columns);
+        assert_eq!(byte_counts, expected_byte_counts);
+    }
+
+    #[test]
+    fn we_can_decompose_large_scalars_to_words() {
+        let mut scalars: Vec<S> = [S::MAX_SIGNED, S::from(u64::MAX), S::from(-1)]
+            .iter()
+            .map(S::from)
+            .collect();
+
+        let mut word_columns = vec![vec![0; scalars.len()]; 31];
+        let mut word_slices: Vec<&mut [u8]> = word_columns.iter_mut().map(|c| &mut c[..]).collect();
+        let mut byte_counts = vec![0; 256];
+
+        decompose_scalar_to_words(&mut scalars, &mut word_slices, &mut byte_counts);
+
+        let expected_word_columns = [
+            [246, 255, 236],
+            [233, 255, 211],
+            [122, 255, 245],
+            [46, 255, 92],
+            [141, 255, 26],
+            [49, 255, 99],
+            [9, 255, 18],
+            [44, 255, 88],
+            [107, 0, 214],
+            [206, 0, 156],
+            [123, 0, 247],
+            [81, 0, 162],
+            [239, 0, 222],
+            [124, 0, 249],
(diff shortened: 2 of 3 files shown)


crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,154 @@
...
+        base::scalar::{Curve25519Scalar as S, Scalar},
+        sql::proof_exprs::range_check::{decompose_scalar_to_words, get_logarithmic_derivative},
+    };
+    use bumpalo::Bump;
+    use bytemuck;
+    use num_traits::Inv;
+    use rand::Rng;
+
+    #[test]
+    fn test_decompose_scalar_to_words() {
```
JayWhite2357, line 82: Can you add some more thorough unit tests here?
Couple examples
`we_can_decompose_small_scalars_to_words` (do a very small example that is easy to verify by hand: e.g. `scalars = vec![S::from(0), S::from(1000), S::from(123)]`.)
`we_can_decompose_large_scalars_to_words` (example where the scalars use the full 31 bytes. The output should be done by hand.)
`we_can_decompose_random_u64_values_to_words` (the example that you have)
etc.

In addition to adding more coverage, it also ensures that I can more easily review the code because I have examples of what the code does, so I both understand it, and agree that it is doing what it should be doing.

Same thing should be done for the logarithmic derivative

crates/proof-of-sql/src/base/commitment/column_bounds.rs
```diff
@@ -228,6 +228,7 @@ impl ColumnBounds {
             | CommittableColumn::Decimal75(_, _, _)
             | CommittableColumn::Scalar(_)
             | CommittableColumn::VarChar(_) => ColumnBounds::NoOrder,
+            CommittableColumn::RangeCheckWord(_) => ColumnBounds::NoOrder,
```
JayWhite2357, line 231: I think this could be in the next PR (or totally separate) since this isn't really used in the two functions that this PR is focused on.

JayWhite2357 (changes requested):

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+}
+
+fn get_logarithmic_derivative<'a, S: Scalar + 'a>(
+    all_scalar_bytes: &[&[u8]],
+    alpha: S,
+    inverted_word_columns: &mut [&mut [S]],
+) {
+    // Phase 2: Use the stored byte arrays and alpha
+    for (i, scalar_bytes) in all_scalar_bytes.iter().enumerate() {
```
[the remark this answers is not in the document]
drcapybara: I was told initially we just wanted something that worked and then we would come back with rayon later. I can add it in if it wont increase the size of this PR too much. I was planning to come back with threaded enhancements later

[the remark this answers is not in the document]
JayWhite2357: Right. PR review is a reasonable place for us to add in this perf improvement since it's so simple.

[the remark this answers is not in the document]
drcapybara: Im getting non-trivial synchronization problems when trying this approach, so Im going to leave rayon for a fast-follow PR for now:

```
error[E0277]: `Cell<NonNull<bumpalo::ChunkFooter>>` cannot be shared between threads safely
   --> crates/proof-of-sql/src/sql/proof_exprs/range_check.rs:15:45
    |
15  |       scalars.par_iter().enumerate().for_each(|(i, scalar)| {
    |                                      -------- ^------------
    |                                      |        |
    |  ____________________________________|________within this `{closure@crates/proof-of-sql/src/sql/proof_exprs/range_check.rs:15:45: 15:58}`
    | |                                    |
    | |                                    required by a bound introduced by this call
16  | |         let scalar_array: [u64; 4] = (*scalar).into(); // Convert scalar to u64 array
17  | |         let scalar_bytes_full = cast_slice::<u64, u8>(&scalar_array); // Cast u64 array to u8 slice
18  | |         let scalar_bytes = alloc.alloc_slice_copy(&scalar_bytes_full[..31]); // Limit to 31 bytes and allocate in bumpalo
...   |
34  | |         }
35  | |     });
    | |_____^ `Cell<NonNull<bumpalo::ChunkFooter>>` cannot be shared between threads safely
    |
    = help: within `{closure@crates/proof-of-sql/src/sql/proof_exprs/range_check.rs:15:45: 15:58}`, the trait `Sync` is not implemented for `Cell<NonNull<bumpalo::ChunkFooter>>`, which is required by `{closure@crates/proof-of-sql/src/sql/proof_exprs/range_check.rs:15:45: 15:58}: Sync`
    = note: if you want to do aliasing and mutation between multiple threads, use `std::sync::RwLock`
note: required because it appears within the type `Bump`
   --> /home/debian/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bumpalo-3.16.0/src/lib.rs:292:12
    |
292 | pub struct Bump {
    |            ^^^^
    = note: required because it appears within the type `&Bump`
note: required because it's used within this closure
   --> crates/proof-of-sql/src/sql/proof_exprs/range_check.rs:15:45
    |
15  |     scalars.par_iter().enumerate().for_each(|(i, scalar)| {
    |                                             ^^^^^^^^^^^^^
note: required by a bound in `rayon::iter::ParallelIterator::for_each`
   --> /home/debian/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/mod.rs:384:30
    |
382 |     fn for_each<OP>(self, op: OP)
    |        -------- required by a bound in this associated function
383 |     where
384 |         OP: Fn(Self::Item) + Sync + Send,
    |                              ^^^^ required by this bound in `ParallelIterator::for_each`
    ```

File: crates/proof-of-sql/src/sql/proof_exprs/range_check.rs (hunk 1 above)
Comment (drcapybara) [reply crates/proof-of-sql/src/sql/proof_exprs/range_check.rs:82]: Absolutely

Comment (drcapybara) [reply crates/proof-of-sql/src/sql/proof_exprs/range_check.rs]: Ive added a more thorough explanation and testing setup. Note that when we slice off the 32nd byte, we lose information from the extrema of supported values, such as S::MAX_SIGNED, or S::from(-1), as the 32nd byte appears to be set in both of these cases. This significantly changes the value of the scalar being decomposed. The current configuration keeps the 32nd byte, but the tests and code can be easily modified to remove it if needed.

File: crates/proof-of-sql/src/sql/proof_exprs/range_check.rs [hunk 4]
```diff
@@ -0,0 +1,412 @@
...
+            236, 211, 245, 92, 26, 99, 18, 88, 214, 156, 247, 162, 222, 249, 222, 20, 0, 0, 0, 0,
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 16,
+        ];
+
+        assert_eq!(all_scalar_bytes[2], &res_3);
+    }
+
+    #[test]
+    fn we_can_decompose_small_scalars_to_words() {
+        let mut scalars: Vec<S> = [1, 2, 3].iter().map(S::from).collect();
```
JayWhite2357, line 135: The decomposition of this column into columns of words should be
`vec![1,2,3], vec![0,0,0], vec![0,0,0], ..., vec![0,0,0]`.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,382 @@
...
+use bytemuck::cast_slice;
+
+// Decomposes a scalar to requisite words, additionally tracks the total
+// number of occurences of each word for later use in the argument.
+fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
+    scalars: &mut [S],
+    alloc: &'a Bump,
+    word_columns: &mut [&mut [u8]],
+    byte_counts: &mut [u64],
+    byte_rows: &mut Vec<&'a [u8]>,
```
JayWhite2357, line 12: We won't need `byte_rows`. You should remove this parameter.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,382 @@
...
+        let byte_counts = alloc.alloc_slice_fill_with(256, |_| 0u64);
+
+        // This will store all of the words that we get from decompose_scalar_to_words
+        let mut byte_rows: Vec<&[u8]> = Vec::with_capacity(scalars.len());
+
+        decompose_scalar_to_words(
+            &mut scalars,
+            &alloc,
+            &mut word_columns,
+            byte_counts,
```
JayWhite2357, line 112: You should be testing that all of the outputs are correct. Is `byte_counts` correct in each of these tests?

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,382 @@
...
+            &mut byte_rows,
+        );
+
+        print_scalar_bytes_matrix(&byte_rows);
+
+        // Check that the decomposition columns are correct
+        assert_eq!(&word_columns[0], &[246, 255, 236]);
+        assert_eq!(&word_columns[1], &[233, 255, 211]);
+        assert_eq!(&word_columns[2], &[122, 255, 245]);
+        // ...
```
JayWhite2357, line 197: Don't just do `...`. Actually finish the test.

JayWhite2357 (changes requested):

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
drcapybara: I will need byte rows when we need to compute constraints, specifically the sum constraints. I can remove this from this PR, but I will need to add it back later

crates/proof-of-sql/src/base/scalar/mod.rs
```diff
@@ -43,6 +43,7 @@ pub trait Scalar:
     + for<'a> core::convert::From<&'a i32> // Required for `Column` to implement `MultilinearExtension`
     + for<'a> core::convert::From<&'a i64> // Required for `Column` to implement `MultilinearExtension`
     + for<'a> core::convert::From<&'a i128> // Required for `Column` to implement `MultilinearExtension`
+    + for<'a> core::convert::From<&'a u8> // Required for `Column` to implement `MultilinearExtension`
```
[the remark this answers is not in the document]
drcapybara: Will remove this

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,382 @@
+use crate::base::{scalar::Scalar, slice_ops};
+use bumpalo::Bump;
```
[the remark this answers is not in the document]
drcapybara: I will need it to compute MLEs over columns though. It would be a lot of work to remove this just to put it back later.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
drcapybara: Will add tests for this

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,382 @@
...
+
+        let mut byte_rows: Vec<&[u8]> = Vec::with_capacity(scalars.len());
+
+        decompose_scalar_to_words(
+            &mut scalars,
+            &alloc,
+            &mut word_columns,
+            byte_counts,
+            &mut byte_rows,
+        );
```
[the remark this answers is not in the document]
drcapybara: decompose_scalar_to_words gives the words that I need before computing a log derivative. Instead we can just put in some bytes by hand. it would be nice to see these both working and connected to each other at some point though

JayWhite2357 (dismissed):
LGTM

crates/proof-of-sql/src/base/scalar/mod.rs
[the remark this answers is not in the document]
JayWhite2357: I didn't realize why you had this. It's because you needed to do the conversion.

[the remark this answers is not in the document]
JayWhite2357: I thought this was a relic of the extra `CommitableColumn` variant you had added. But, it's actually useful for converting the bytes to Scalars inside the `get_logarithmic_derivative`.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -0,0 +1,316 @@
...
+    byte_columns: &[&mut [u8]],
+    alpha: S,
+    inverted_word_columns: &mut [&mut [S]],
+) {
+    // Iterate over each column
+    for (i, byte_column) in byte_columns.iter().enumerate() {
+        // Convert bytes to field elements and add alpha
+        let mut terms_to_invert: Vec<S> = byte_column
+            .iter()
+            .map(|&w| S::try_from(w.into()).unwrap() + alpha)
```
JayWhite2357, line 38:
```suggestion
            .map(|w| S::from(w) + alpha)
```

JayWhite2357 (changes requested):

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
JayWhite2357: You will need to add back in the `+ for<'a> core::convert::From<&'a u8>` to `Scalar` to make this work.

drcapybara: Ok, Ill add it back, then clarify this addition in #146

JayWhite2357 (approved):
LGTM

PR #152, opened 2024-09-12
fix: safely handle max integer type

drcapybara:
# Rationale for this change

This PR fixes the function ```max_integer_type``` to safely handle the conversion from ```ColumnType``` into ```integer_bits```, if the type is an integer.

Dont unwrap ```to_integer_bits```, use an ```and_then``` chain instead.

# Are these changes tested?

yes, existing tests pass


iajoiner (approved):
That looks great! Thanks @drcapybara !

PR #159, opened 2024-09-13
feat: add functionality to `Transcript` trait and migrate sumcheck to use it

JayWhite2357:
# Rationale for this change

In order to better support WASM, as well as enable solidity support, we need to be able to use a simplified transcript. However, the existing `Transcript` trait does not have the appropriate functionality to be able to only use the trait.

* The `Transcript::extend_serialize_as_le` method is added in order to be used as a possible replacement for the `<merlin::Transcript as TranscriptProtocol>::append_auto` method.
* The `Transcript::extend_canonical_serialize_as_le` method is added in order to be used as a possible replacement for the `<merlin::Transcript as TranscriptProtocol>::append_canonical_serialize` method.
* The `Transcript::wrap_transcript` method is added for when a `Transcript` needs to be wrapped in another.

* The sumcheck proof is updated to accept `impl Transcript` instead of `merlin::Transcript`.

This is technically a breaking change since proofs are not backward compatible.

# Are these changes tested?

Yes

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/commitment_evaluation_proof.rs
+++ b/crates/proof-of-sql/src/base/commitment/commitment_evaluation_proof.rs
@@ -1,5 +1,5 @@
 use super::Commitment;
-use crate::base::scalar::Scalar;
+use crate::base::{proof::Transcript as _, scalar::Scalar};
 #[cfg(feature = "blitzar")]
 use crate::base::{scalar::MontScalar, slice_ops};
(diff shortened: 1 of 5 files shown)


iajoiner (approved):

iajoiner:
@JayWhite2357 Thanks! BTW please try to avoid uppercase letters in branch names since it makes it necessary to quote the branch name in CLI.

PR #162, opened 2024-09-14
feat: add a range check word type for mle

drcapybara:
# Rationale for this change

The ```CommitableColumn``` types are defined as types that can be used to produce an MLE. We need to add support here for committable columns of u8s, as this is the initial word size we have chosen for the range check.

- [x] Add CommitableColumn RangeCheckWord type
- [x] Add RangeCheckWord match arm in column bounds ordering
- [x] Add support for RangeCheckWord in dory commitment cpu helper
- [x] Add support for RangeCheckWord in pack scalars
- [x] Add from [u8] implementation for CommitableColumn   

# Are these changes tested?

- [x] we_can_commit_to_rangecheckword_column_through_committable_column
- [x] we_can_get_length_of_rangecheckword_column


JayWhite2357 (approved):
LGTM

PR #169, opened 2024-09-19
refactor: add SubpolynomialType to verifier builder

JayWhite2357:
# Rationale for this change

The verifier and prover builders are a bit asymmetric with the `sumcheck_subpolynomial`. Additionally, the explicit use of `random_evaluation` is a bit confusing.

* `produce_sumcheck_subpolynomial_evaluation` now accepts a `SumcheckSubpolynomialType` as a parameter.
* Since this is a (only internal) breaking change, this also modifies `produce_sumcheck_subpolynomial_evaluation` to own it's `eval` parameter rather than borrow.

# Are these changes tested?

Yes, by existing code.


Dustin-Ray (approved):
LGTM - Thanks

PR #172, opened 2024-09-20
refactor: removal of `std` references for initial `no_std` support

JayWhite2357:
# Rationale for this change

In order to be able to update commitments from within a `no_std` environment (like WASM), we need to allow a portion of the `sxt-proof-of-sql` crate to be `no_std` compatible. This PR does the first step of this support.

* The first commit refactors some code to no longer reference `std` directly, but rely on `use` statements instead.
* The second commit adds/modifies `use` statements in order to remove `std` from the relevant portion of the crate.

# Are these changes tested?

Yes, by existing tests.


tlovell-sxt (approved):
LGTM

PR #178, opened 2024-09-24
feat: add `rayon` feature flag

JayWhite2357:
# Rationale for this change

One of the components of this repo that is not `no_std` is rayon.

* The `slice_cast` module is simplified and leveraged more.
* An `if_rayon!` macro is added. This is the key to the majority of this PR.
    ```rust
    macro_rules! if_rayon {
        ($rayon_value: expr, $else_value: expr) => {{
            #[cfg(feature = "rayon")]
            {
                ($rayon_value)
            }
            #[cfg(not(feature = "rayon"))]
            {
                ($else_value)
            }
        }};
    }
    ```
* A `rayon` feature flag is added and the `if_rayon!` macro is used everywhere `rayon` shows up.

# Are these changes tested?

Yes, by existing tests.
An additional test run is added for the non-rayon code. This significantly slows down the CI job, so we may want to think through a better way to test this code. Perhaps the rayon can be isolated to a smaller module, and only that module tested.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -75,6 +75,8 @@ jobs:
           sudo apt-get install -y clang lld
       - name: Run cargo test
         run: cargo test --all-features
+      - name: Run cargo test without rayon
+        run: cargo test --no-default-features --features="arrow blitzar"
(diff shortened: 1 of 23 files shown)


iajoiner (approved):
Thanks @JayWhite2357 !

PR #179, opened 2024-09-24
feat: remove `std` usage by `indexmap`

JayWhite2357:
# Rationale for this change

The `indexmap` crate is `no_std` compatible, but a few changes must be made in order to not use the `std` feature.

A `base::map::IndexMap` and `base::map::IndexSet` type alias are created to be a replacement for `indexmap::IndexMap` and `indexmap::IndexSet` respectively. This is needed because `indexmap` does not have a default `BuildHasher` type parameter in `no_std` mode.
Additionally, `indexmap` does no expose the `indexmap!` macro in `no_std` mode. So, this PR adds a replacement macro.

Finally,
* `Index*::new` is replaced by `Index*::default` which exists in `no_std`.
* `Index*::from` is replaced by `Index*::from_iter` which exists in `no_std`.

# Are these changes tested?

Yes, by existing tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ version = "0.0.0" # DO NOT CHANGE THIS LINE! This will be automatically updated
 license-file = "LICENSE"
 
 [workspace.dependencies]
+ahash = { version = "0.8.11", default-features = false, features = ["runtime-rng"] }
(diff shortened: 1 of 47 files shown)


tlovell-sxt (approved):
LGTM

tlovell-sxt (approved):

Issue #161, opened 2024-09-13
Enable using transcripts other than `merlin::Transcript`.


JayWhite2357:
# Background and Motivation
Currently, the Proof of SQL prover uses the `merlin` crate to provide it's public-coin transcript, created here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/29d334d5cd6e31cf66a0353f21613932df63c059/crates/proof-of-sql/src/sql/proof/query_proof.rs#L62.
However, `merlin` does not support `no_std` (see #117 and https://github.com/zkcrypto/merlin/pull/8) and is not conducive to use within the EVM/Solidity.

https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/123 and https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/159 add a new `Transcript` trait that allows for using different transcripts. (Such as the simple `Keccak256Transcript`.)

If we replace all uses of `merlin::Transcript` with `impl Transcript`, we enable the ability to use any transcript implementation that is desired.

# Requested Changes
We need to replace all usage of `merlin::Transcript` with `impl Transcript`. Tests can still leverage `merlin::Transcript` because they need a concrete implementation, but all other code should be written generically.

- `extend_serialize_as_le` can replace `append_auto`
- `extend_canonical_serialize_as_le` can replace `append_canonical_serialize`
- `scalar_challenge_as_be` can replace `challenge_scalar_single` ( and `challenge_scalars` )
- other methods can be better replacements as relevant

- [ ] `CommitmentEvaluationProof::new`, `CommitmentEvaluationProof::verify_proof`, `CommitmentEvaluationProof::verify_batched_proof` should accept `impl crate::base::proof::Transcript` instead of `merlin::Transcript`.
- [ ] All implementers of `CommitmentEvaluationProof` will need to be updated to implement the changes properly.
    - [ ] `InnerProductProof` - this will still internally need to use `merlin::Transcript`. This is handled via `wrap_transcript`.
    - [ ] `DoryEvaluationProof` - the usages of the transcript bubble down to `DoryMessages`.
- [ ] Either `QueryProof::new` and `QueryProof::verify` can be made generic w.r.t. `Transcript`, or `QueryProof` itself can be made generic w.r.t. `Transcript`. It is unclear which is a better design choice. Both solutions will require replacing the old transcript usage with the new within those functions.
[...]

PR #180, opened 2024-09-24
feat!: put `merlin` behind feature

JayWhite2357:
# Rationale for this change

See #161 for rationale.

NOTE: the individual commits may be good to review individually

* `DoryMessages` accepts `impl Transcript` instead of `merlin::Transcript`. This is a breaking change, making the proofs non-backward compatible. The API deos not change.
* Changed `CommitmentEvaluationProof` to accept `impl Transcript` rather than `merlin::Transcript`.
* Replaced `merlin::Transcript` with `Keccak256Transcript` within `QueryProof`. Although the proof could be made generic, there is no real use-case for `merlin`. In the future, we can easily support other transcripts, but I think it is best to keep it concrete for now.
* Dropped dead code and renamed the old `merlin` module.
* Put `merlin` behind the `blitzar` feature flag since it is still required for the `blitzar::InnerProductProof`.

* The tests are _NOT_ refactored to remove `merlin`. This can be done in a separate PR.

# Are these changes tested?

Yes. Existing tests cover the changes.


stuarttimwhite (approved):

tlovell-sxt (approved):
LGTM. I wasn't aware of the new transcript/core traits but their improved implementation/ergonomics are definitely on display here

Issue #174, opened 2024-09-20
refactor: migrate from `thiserror` to `snafu`


JayWhite2357:
# Background and Motivation
In order to make this rep `no_std` compatible, we need `no_std` compatible errors. `thiserror` is not `no_std` compatible, while `snafu` is. We are currently using `thiserror`.

Note: While https://github.com/dtolnay/thiserror/pull/304 is an open PR, it is not clear how soon this will get merged. Additionally, Substrate uses `snafu`, which would help unify dependencies for Substrate-based integrations.

# Requested Changes

- [ ] Replace `thiserror` with `snafu`

PR #181, opened 2024-09-25
refactor!: migrate from `thiserror` to `snafu`

lgiussan:
# Rationale for this change
This PR should close https://github.com/spaceandtimelabs/sxt-proof-of-sql/issues/174.

NOTE: the individual commits may be good to review individually.
- All error enum variants consisting of tuple structs are transformed into named structs. This is necessary because `snafu` does not support tuple structs.
- Every `#[derive(Error)]` is substituted with the analogous `#[derive(Snafu)]`. In particular:
    * `#[error(...)]` attributes are substituted with equivalent `#[snafu(display(...))]` attributes
    * `#[error(transparent)]` attributes are substituted with equivalent `#[snafu(transparent)]` attributes (which also derive the corresponding `From` implementation)
    * For `ConversionError::TimestampConversionError`, the `#[snafu(context(false), display(...))]` attribute is used for deriving a `From` implementation, and at the same time maintain the custom error message
- A `std` feature is introduced for the `proof-of-sql` crate, which in turns activates the `snafu/std` feature. The `std` feature is required for the `posql_db` example, because `Clap` relies on the `std::error::Error` trait.
- `thiserror` still appears in the dependency tree (`cargo tree -i thiserror`), but only as a transitive dependency (via `blitzar`, and dev-dependencies)

# Are these changes tested?
Yes. This PR is a refactoring, and all existing tests pass.


JayWhite2357:
Thanks for the PR!

Dustin-Ray:
@lgiussan this looks awesome thanks for the effort! Would you mind splitting this into smaller PRs if possible? Then we can quickly review and approve them, 70 files with a diff this large is difficult to review all at once. You can make small PRs and then reference issue #174. In general you can also squash your commits into a single commit as well when you are ready to merge. Thanks!

crates/proof-of-sql/Cargo.toml
```diff
@@ -75,6 +75,7 @@ development = ["arrow-csv"]
 default = ["arrow", "blitzar", "rayon"]
 arrow = ["dep:arrow"]
 test = ["dep:rand"]
+std = ["snafu/std"]
```
JayWhite2357, line 78: This is causing CI to fail. (Take a look if you're curious.)
You _could_ fix the CI here, but I would much rather simply you add `features = ["std"]` to `snafu`.

The reason for this is that I have a PR incoming that adds true `no_std` support and will be adding a `std` feature. It would be much cleaner to add this feature there, rather than here.

I'm now aware of the `posql_db` concern and the `snafu/std` feature. (I may just disable the `clap/std` feature, for example.)

JayWhite2357 (changes requested):
Thanks for the big effort on this! I appreciate the attention to detail.

One small change about the `std` feature. I think the solution is to simply drop the last commit, and it will be good. I'll add an equivalent change in my next 1 or 2 PRs.

tlovell-sxt (dismissed):
This is great, LGTM - obviously Jay's suggestion will help with the failing CI.

crates/proof-of-sql/Cargo.toml
lgiussan: That makes completely sense. I've addressed this in https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/181/commits/31910e5b098e94646f1087cfb2910590933556f0

JayWhite2357 (dismissed):
LGTM.

Looks like you need to rebase and resolve some conflict with main. I'll need to re-approve once you do that, but it should be good once you do.

JayWhite2357 (approved):

tlovell-sxt (approved):

PR #187, opened 2024-09-26
refactor: eliminate std usage

JayWhite2357:
# Rationale for this change

In order to support `no_std`, we need to eliminate uses of `std` from the repo as much as possible.

* explicit uses of `std` are removed.
* implicit uses of `std` are no longer required via usage of `use alloc::vec::Vec` and friends.

It is possible I missed a few spots, but these can be cleaned up in the final PR that enables `no_std`. This PR functionally does nothing until we enable `no_std`.

# Are these changes tested?

Yes, by existing tests.


Dustin-Ray:
LGTM although theres no way to know until ```!#[no_std]``` is added

Dustin-Ray (dismissed):
conditionally approved once CI passes

Dustin-Ray (approved):
conditionally approved once CI passes

PR #188, opened 2024-09-26
refactor: initialised `clippy::missing_panics_doc`

### LINKED ISSUE #163: Enable `clippy::missing_panics_doc`
# Background and Motivation

In order to improve documentation and identify possible panics, we should enable clippy lints that warn when a function that panics but does not document this behavior.

# Requested Changes

- [ ] Enable `clippy::missing_panics_doc` lint.
- [ ] Add `#![cfg_attr(test, allow(clippy::missing_panics_doc))]` in order to not warn on `test`-only functions.
- [ ] Add documentation for every function violating this lint. (At the time of writing, this is 43 functions.)
- [ ] Set the `check-private-items = true` option.
- [ ] Add documentation for the private items violating this lint. (At the time of writing, this is 73 _additional_ functions.)

It is ok for this to be done in separate PRs. (e.g. one PR could enable the lint, and simply write `TODO: add panic docs` everywhere the warning appears.
NOTE: `allow(clippy::missing_panics_doc)` should only be used if a `panic` is actually impossible.

Gmin2:
# Rationale for this change

In order to improve documentation and identify possible panics, we should enable clippy lints that warn when a function that panics but does not document this behavior.

Added panic doc comments for all the `unwrap()` statements

Fixes #163 
/claim #163


Gmin2:
@JayWhite2357 , to complete the doc part of it will take two to three days

Cargo.toml
```diff
@@ -66,3 +66,6 @@ zerocopy = { version = "0.7.34" }
 
 [workspace.lints.rust]
 missing_docs = "warn"
+
+[workspace.lints.clippy]
+missing_panics_doc = "warn"
```
JayWhite2357, line 71: NIT:
```suggestion
missing_panics_doc = "warn"

```

clippy.toml
```diff
@@ -1 +1,2 @@
 missing-docs-in-crate-items = true
+check-private-items = true
```
JayWhite2357, line 2: NIT:
```suggestion
check-private-items = true

```

crates/proof-of-sql-parser/src/identifier.rs
```diff
@@ -85,6 +85,7 @@ impl AsRef<str> for Identifier {
     }
 }
 
+#[allow(clippy::missing_panics_doc)]
```
JayWhite2357, line 88: Rather than putting these on each test module, it should work to put
```rust
#![cfg_attr(test, allow(clippy::missing_panics_doc))]
```
at the top of `lib.rs`.

crates/proof-of-sql-parser/src/intermediate_ast.rs
```diff
@@ -224,6 +224,7 @@ impl Expression {
             expr: Box::new(self),
         })
     }
+    // TODO: add panic docs
     /// Create an `AliasedResultExpr` from an `Expression` using the provided alias.
```
JayWhite2357, line 228: Something like this should allow the lints to pass.
```suggestion
    /// Create an `AliasedResultExpr` from an `Expression` using the provided alias.
    ///
    /// # Panics
    ///
    /// TODO: add panic docs
```

JayWhite2357 (changes requested):
If you get this to pass CI, I'm happy to merge this PR while you work on the docs in a separate PR.
It would prevent any new code from being undocumented in the meantime.

JayWhite2357:
Since you've contributed a couple PRs now, it would be helpful for you to be a bit more verbose on the merge requests.
Check out some of the recently merged PRs for some examples: https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/181, https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/180.

JayWhite2357:
@Gmin2 it looks like I can't approve the CI workflows unless the branch conflicts are resolved.

Gmin2:
Hey @JayWhite2357, i think it will make sense to include docs in this PR only, wdyt?

JayWhite2357:
> Hey @JayWhite2357, i think it will make sense to include docs in this PR only, wdyt?

It doesn't matter to me. Entirely up to you.

Gmin2:
Hey @JayWhite2357, completed all the panic docs for `unwrap()` statements, but the assert statements are giving warnings what should i do about them ?

JayWhite2357:
> Hey @JayWhite2357, completed all the panic docs for `unwrap()` statements, but the assert statements are giving warnings what should i do about them ?

@Gmin2 

If the assert is _guaranteed_ to not panic, then it is safe to use `allow(clippy::missing_panics_doc, reason = "some reasoning why we should not include a panic doc")`.
Be sure that the reason is thorough. If needed, put a short reason in the reason = "..." and if a more verbose explanation is needed, put more in some comments (not docs).

crates/proof-of-sql-parser/src/identifier.rs
JayWhite2357: I still would like this to be done. Is there a reason we can't do this?

crates/proof-of-sql-parser/src/utility.rs
```diff
@@ -1,6 +1,10 @@
 use crate::{intermediate_ast::*, Identifier, SelectStatement};
 use alloc::{boxed::Box, vec, vec::Vec};
 
+///
+/// # Panics
+///
+/// This function will panic if either the `name` or the `schema` (if provided) cannot be parsed.
```
JayWhite2357, line 7: There is no `schema` here.

crates/proof-of-sql-parser/src/utility.rs
```diff
@@ -98,6 +102,9 @@ pub fn div(left: Box<Expression>, right: Box<Expression>) -> Box<Expression> {
 /// Get table from schema and name.
 ///
 /// If the schema is `None`, the table is assumed to be in the default schema.
+/// # Panics
+///
+/// This function will panic if either the `name` or the `schema` (if provided) cannot be parsed.
```
JayWhite2357, line 107:
```suggestion
/// This function will panic if either the `name` or the `schema` (if provided) cannot be parsed as valid [Identifier]s.
```
The reason the "as valid Identifiers" is important is that it is not clear without that clause what would make it parsing not possible.

Please update this in other spots as well.

crates/proof-of-sql/src/base/polynomial/interpolate.rs
```diff
@@ -13,7 +13,14 @@ use num_traits::{Inv, One, Zero};
 ///
 /// For any polynomial, `f(x)`, with degree less than or equal to `d`, we have that:
 /// `f(x) = sum_{i=0}^{d} (-1)^(d-i) * (f(i) / (i! * (d-i)! * (x-i))) * prod_{i=0}^{d} (x-i)`
-/// unless x is one of 0,1,...,d, in which case, f(x) is already known.
+/// unless x is one of 0,1,...,d, in which case, f(x) is already known.\
+/// 
+///  # Panics
+/// * This function panics if any of the following occurs:
+///   * When computing the inverse of `(i! * (d-i)! * (x-i))`, if the value `x - i` is zero
+///     (which is guaranteed to not happen as `i == x` will return early).
```
JayWhite2357, line 21: Since this is guaranteed to not happen, it shouldn't be included in the documentation.

Instead, the `unwrap` should really be an `expect` with a reason.

crates/proof-of-sql/src/base/polynomial/interpolate.rs
```diff
@@ -13,7 +13,14 @@ use num_traits::{Inv, One, Zero};
...
+///   * If the slice of polynomial coefficients is empty, the function will return `F::zero()`
+///     without performing any operations.
```
JayWhite2357, line 23: This isn't a panic, and is normal behavior. The `0` polynomial is always `0` at every point.

Although if this confused you, it appears that this clarifying note should be added to the documentation (outside of the Panic documentation).

crates/proof-of-sql/src/base/polynomial/interpolate.rs
```diff
@@ -13,7 +13,14 @@ use num_traits::{Inv, One, Zero};
 ///
 /// For any polynomial, `f(x)`, with degree less than or equal to `d`, we have that:
 /// `f(x) = sum_{i=0}^{d} (-1)^(d-i) * (f(i) / (i! * (d-i)! * (x-i))) * prod_{i=0}^{d} (x-i)`
-/// unless x is one of 0,1,...,d, in which case, f(x) is already known.
+/// unless x is one of 0,1,...,d, in which case, f(x) is already known.\
+/// 
+///  # Panics
```
JayWhite2357, line 18: I seems as though this should never panic. So, this is a case where `allow`ing the lint is appropriate. (With a `reason =`.)

crates/proof-of-sql/src/base/polynomial/interpolate.rs
```diff
@@ -72,6 +79,13 @@ where
 /// Let d be the evals.len() - 1 and let f be the polynomial such that f(i) = evals[i].
 /// The output of this function is the vector of coefficients of f, leading coefficient first.
 /// That is, `f(x) = evals[j]*x^(d-j)``.
+///
+/// # Panics
+///
+/// This function may panic in the following situations:
+/// - If the evaluation points are insufficient (less than 1), resulting in `n` being 0 or negative, which could lead to incorrect calculations or indexing issues.
```
JayWhite2357, line 86: Check out this:
https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/173ec574d93118b5c6344b9ea1713939b26d0481/crates/proof-of-sql/src/base/polynomial/interpolate_test.rs#L98-L163

These cases behave properly without panic.

crates/proof-of-sql/src/base/polynomial/interpolate.rs
```diff
@@ -72,6 +79,13 @@ where
...
+/// - The `inv()` method on `Product` type can panic if the product is zero, leading to a division by zero situation when calculating the inverse.
```
JayWhite2357, line 87: The product should never be zero because the numbers being multiplies are all non-zero.
This should be explained in an `expect` rather than an `unwrap`.

crates/proof-of-sql/src/base/polynomial/interpolate.rs
```diff
@@ -72,6 +79,13 @@ where
...
+/// - The `unwrap()` calls will panic if the value being unwrapped is `None`, which can occur during the evaluation of Lagrange basis polynomials.
```
JayWhite2357, line 88: The `unwrap_or` doesn't panic. It returns `vec![]` if the `unwrap` _would_ have `panic`ed.

crates/proof-of-sql/src/base/proof/transcript_core.rs
```diff
@@ -15,6 +15,9 @@ pub(super) trait TranscriptCore {
 }
 
 /// private method to facilitate recieving challenges and reversing them. Undefined behavior if the size_of `M` is not 32 bytes.
+///
+/// # Panics
+/// - Panics if `M::read_from(&bytes)` fails to read the bytes into the type `M`.
```
JayWhite2357, line 20: Can this happen? What would cause `read_from` to fail?

crates/proof-of-sql/src/base/scalar/mont_scalar.rs
```diff
@@ -129,6 +129,10 @@ impl<T: MontConfig<4>> MontScalar<T> {
         Self(value)
     }
     /// Create a new `MontScalar<T>` from a `[u64, 4]`. The array is expected to be in non-montgomery form.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if the provided `[u64; 4]` cannot be converted into a valid `BigInt` due to an overflow or invalid input. The method unwraps the result of `Fp::from_bigint`, which will panic if the `BigInt` does not represent a valid field element.
```
JayWhite2357, line 135: Take a look at https://docs.rs/ark-ff/latest/ark_ff/fields/trait.PrimeField.html#tymethod.from_bigint.

Explain what "invlid input" means.

crates/proof-of-sql/src/base/scalar/mont_scalar.rs
```diff
@@ -275,6 +279,10 @@ impl From<Curve25519Scalar> for curve25519_dalek::scalar::Scalar {
 }
 
 impl From<&Curve25519Scalar> for curve25519_dalek::scalar::Scalar {
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if the byte array cannot be converted to a canonical scalar. This can occur if the byte array is not of the expected length (32 bytes), leading to a failure in the from_canonical_bytes` method.
```
JayWhite2357, line 285: I'm not sure this failure can occur.

JayWhite2357 (changes requested):
I started reviewing some of this.

The general feedback I have on this: there are some cases where an `unwrap` or another panic _could_ occur, but based on the logic of the methods, doesn't occur. This should not be included in the Panic documentation, and instead should be `expect`ed away and explained in comments (no docs).

Additionally, there are a couple of spots I noted where the panic documentation could be a bit more descriptive on the root cause of _why_ the panic occurs.

In both cases, you may have to look at, for example, _why_ a method returns `None` so that you can explain either _why_ a panic might occur or _why_ it cannot occur.

JayWhite2357:
@Gmin2 thanks for the effort here. It is looking good!

crates/proof-of-sql-parser/src/identifier.rs
Gmin2: done in this [commit](https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/188/commits/af0941d24812df74eed828087e098f2d203b4002)

crates/proof-of-sql/src/base/proof/transcript_core.rs
Gmin2: ig yes

Gmin2:
> @Gmin2 thanks for the effort here. It is looking good!

Hey @JayWhite2357 , i have added all the changes you have mentioned, can you take a look and one more thing can we wrap this quickly because the merge conflicts are getting bigger and bigger, if somethings needs to be adjusted we can do it in the followup PR

JayWhite2357 (approved):

PR #192, opened 2024-09-27
feat: add `std` feature and almost full `no_std` support.

JayWhite2357:
# Rationale for this change

We want `no_std` support.

* Added a `std` and `perf` feature to `proof-of-sql`.
* Made `proof-of-sql-parser` `no_std`.
* Updated dependency features to disable `std` usage where not needed.
* Added CI jobs to ensure these features always pass.

The one caveat is that lalrpop needs a patch while waiting on https://github.com/lalrpop/lalrpop/pull/953 in order to have true `no_std` support. To test `no_std` compatibility, we can build on a target that does not support `std`. There is a new CI job that targets `thumbv7em-none-eabi`. This fails without the lalrpop patch.

Once `lalrpop` cuts a release, we can simply bump the version and remove the patch in the CI.

# Are these changes tested?

Yes


tlovell-sxt (approved):
LGTM. Thanks

Issue #183, opened 2024-09-25
Add nullable column support


JayWhite2357:
# Background and Motivation
Currently, we have not implemented support for columns with null values. This is a very useful and ubiquitous primitive for any database engine.

# Changes Required
- [ ] Support adding a flag to the columns types in order to support nullable columns. Logically, this is the same as adding a new column datatype for each existing datatype.
- [ ] Support existing operations on nullable columns. (e.g. we should be able to add a nullable bigint to a non-nullable bigint)

JayWhite2357:
@iajoiner Can you scope out this issue a bit more and come up with a concrete plan.

Alternatively, if someone claims the bounty and is able to come up with a satisfactory plan themselves, I will tip an extra $1000 after completion.

JayWhite2357:
@iajoiner Can you scope out this issue a bit more and come up with a concrete plan.

Alternatively, if someone claims the bounty and is able to come up with a satisfactory plan themselves, I will tip some extra toward.

ssddOnTop:
@JayWhite2357 I am new to the project.. can you provide some sample commands or something to test against? Something that can help me to navigate through the code easily and more importantly, understand the issue properly.

ssddOnTop:
Also, can someone please provide link to the discord server for easier communications

JayWhite2357:
Take a look at the [examples](https://github.com/spaceandtimelabs/sxt-proof-of-sql/tree/main?tab=readme-ov-file#examples).
Also, take a look at the tests here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/tree/main/crates/proof-of-sql/tests.

brainless:
@JayWhite2357 I am not planning to attempt this but I would happy to scope this out. I will look into your linked tests.

JayWhite2357:
@brainless I can't promise a tip for just scoping it out. The bounty is specifically for completing the issue.
This is a potentially complex issue, where the correct, full design will likely not be clear until the code is complete.
To this point, @ssddOnTop, you may wish to get feedback from @iajoiner and myself on any design before you actually go through the effort of implementing it.

brainless:
@JayWhite2357 sorry for the mis-communication. I did not mean a tip at all. I feel it is a great way for me to learn about the project before I want to take any bounty in the future.

JayWhite2357:
[...]

PR #194, opened 2024-09-28
feat: add nullable variant in `column`

ssddOnTop:
Part of: #183

I am confused with current flow of code.. so I could not understand why do we need `NullableColumn` and `NullableOwnedColumn`

I am not sure why `Column` needs serialization/deserialization, so ignoring the serde part,
here is the proposed change in `Column` itself which could help mark it nullable.

I edited `try_add_subtract_column_types` in [column_operation.](https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/crates/proof-of-sql/src/base/database/column_operation.rs) as an example.

ps: Sorry for not following the PR format

### DIFF
--- a/crates/proof-of-sql/src/base/database/column.rs
+++ b/crates/proof-of-sql/src/base/database/column.rs
@@ -205,7 +205,7 @@ impl<'a, S: Scalar> Column<'a, S> {
 ///
 /// See `<https://ignite.apache.org/docs/latest/sql-reference/data-types>` for
 /// a description of the native types used by Apache Ignite.
-#[derive(Eq, PartialEq, Debug, Clone, Hash, Serialize, Deserialize, Copy)]
+#[derive(Eq, PartialEq, Debug, Clone, Hash, Serialize, Deserialize)]
 pub enum ColumnType {
     /// Mapped to bool
     #[serde(alias = "BOOLEAN", alias = "boolean")]
@@ -234,6 +234,10 @@ pub enum ColumnType {
     /// Mapped to Curve25519Scalar
     #[serde(alias = "SCALAR", alias = "scalar")]
     Scalar,
+
+    /// Nullable version of the column type
+    #[serde(alias = "NULLABLE", alias = "nullable")]
+    Nullable(Box<ColumnType>),
 }
 
 impl ColumnType {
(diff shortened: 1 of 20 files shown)


ssddOnTop:
@iajoiner can I get a quick review on this

crates/proof-of-sql/src/base/database/nullable_column.rs
```diff
@@ -0,0 +1,22 @@
+use crate::base::database::Column;
+use crate::base::scalar::Scalar;
+
+/// Represents a nullable column with values and an optional presence slice
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub struct NullableColumn<'a, T: Scalar> {
+    values: Column<'a, T>,
```
iajoiner, line 7: We should expose `values` and `presence`

ssddOnTop:
@iajoiner how is the approach to add `Nullable` in `ColumnType`? Am I going in correct direction?

crates/proof-of-sql/src/base/database/column.rs
```diff
@@ -234,28 +234,36 @@ pub enum ColumnType {
     /// Mapped to Curve25519Scalar
     #[serde(alias = "SCALAR", alias = "scalar")]
     Scalar,
+
+    /// Nullable version of the column type
+    #[serde(alias = "NULLABLE", alias = "nullable")]
+    Nullable(Box<ColumnType>),
```
iajoiner, line 240: I think `ColumnType` and nullability should be orthogonal.

ssddOnTop: by orthogonal do you mean that we should create a wrapper struct or enum?

iajoiner: @ssddOnTop `ColumnType` shouldn't change. Instead `NullableColumn` should have a simple `is_nullable` method.

ssddOnTop: @iajoiner what do you think should be the scope of this PR?

Shall I add/modify functions to perform column expr for this PR?

iajoiner:
Closed in favor of #620.

PR #195, opened 2024-09-28
WIP: Enable clippy::pedantic lint group

### LINKED ISSUE #182: Enable `clippy::pedantic` lint group.
# Background and Motivation
We have `cargo clippy` running in our CI in order to enforce code quality. In order to increase our standards, we should enable the `clippy::pedantic` lint group.

# Requested Changes
- [ ] Add the `clippy::pedantic` lint to the bottom of the workspace `Cargo.toml` file.
- [ ] Refactor all of the code warnings so that clippy will pass.

NOTE: `clippy::pedantic` is very aggressive. It is possible that we will want to blanket `allow` some lints.

NOTE: please have good git hygiene on this issue. Make sure the individual commits on a PR are small and individually readable.

mehulmathur16:
/fixes #182 
/claim #182


mehulmathur16:
Hi @JayWhite2357 , I’ve made some initial progress on integrating clippy::pedantic. Let me know if you’re available to connect on Discord for smoother communication.

JayWhite2357:
> @JayWhite2357, could you suggest how we should proceed with this issue? The file changes are extensive, primarily due to autofixes. Would you prefer separate PRs for the Clippy autofixes and the manual fixes?

I would suggest breaking this into smaller PRs, but it's not completely needed. The benefit of multiple PRs is that we could potentially review different lints in parallel and make suggestions on one set while approving another set.

In any case, the best way to proceed would be by making sure that the git commits are well organized.
In this case, we'll want to review it by looking at the commits. So, for example, something like this would be good:

* Commit for manual fixes to 1 specific lint.
* Commit for the autofixes to that same lint. (In the description part of the commit message it would make sense to put the actual clippy command that you ran.)
* Any extra fixes needed after autofixes to that lint.

Then just repeat that pattern for each individual lint. That way it is easy to review one lint at a time, and it's easy to distinguish between autofixes and manual fixes.

Focus on building the commits to make it easy to review.

mehulmathur16:
> > @JayWhite2357, could you suggest how we should proceed with this issue? The file changes are extensive, primarily due to autofixes. Would you prefer separate PRs for the Clippy autofixes and the manual fixes?
> 
> I would suggest breaking this into smaller PRs, but it's not completely needed. The benefit of multiple PRs is that we could potentially review different lints in parallel and make suggestions on one set while approving another set.
> 
> In any case, the best way to proceed would be by making sure that the git commits are well organized. In this case, we'll want to review it by looking at the commits. So, for example, something like this would be good:
> 
> * Commit for manual fixes to 1 specific lint.
> * Commit for the autofixes to that same lint. (In the description part of the commit message it would make sense to put the actual clippy command that you ran.)
> * Any extra fixes needed after autofixes to that lint.
> 
> Then just repeat that pattern for each individual lint. That way it is easy to review one lint at a time, and it's easy to distinguish between autofixes and manual fixes.
> 
> Focus on building the commits to make it easy to review.

Sure

Issue #113, opened 2024-08-16
refactor: remove intermediate decimal and move decimal into PoSQL parser crate


Dustin-Ray:
# Background and Motivation
Currently, the code handling decimal parsing and conversions is somewhat scattered. This issue seeks to resolve this by using the `bigdecimal::BigDecimal` type for as much as possible.

Important Note: the `BigDecimal` type does not trim leading and trailing zeros. Care should be taken when dealing with this.

# Requested Changes
To resolve this issue, we should:
- [ ] Add a `BigDecimalExt` extension trait in the `proof_of_sql::base::math` module. It should have, at minimum,
    - [ ] `precision` and `scale` methods akin to the existing `IntermediateDecimal::precision` and `IntermediateDecimal::scale` methods. NOTE: the precision and scale should be the values _after_ trimming any leading and trailing zeros. This is what the current implementation (implicitly) does.
    - [ ] a `try_into_scalar_with_precision_and_scale` method akin to the exiting `try_into_to_scalar` function.
- [ ] Eliminate the `proof_of_sql_parser::intermediate_decimal::IntermediateDecimal` and `proof_of_sql::base::math::decimal::Decimal` types, in favor of using `bigdecimal::BigDecimal` along with the `BigDecimalExt` extension trait.

Optional:
- [ ] Other refactoring and/or cleanup is welcome, but not immediately clear to me what should be done. These likely should be done in a follow up PR/issue. Some possibilities are:
    - [ ] It is not clear if the `DecimalError` enum is needed or if `bigdecimal::ParseBigDecimalError` or similar is sufficient.
    - [ ] The `scale_scalar` function should likely be moved into the `proof_of_sql::base::scalar` module.

JayWhite2357:
To resolve this issue, we should:
- [ ] Add a `BigDecimalExt` extension trait in the `proof_of_sql::base::math` module. It should have, at minimum,
    - [ ] `precision` and `scale` methods akin to the existing `IntermediateDecimal::precision` and `IntermediateDecimal::scale` methods.
    - [ ] a `try_into_scalar_with_precision_and_scale` method akin to the exiting `try_into_to_scalar` function.
- [ ] Eliminate the `proof_of_sql_parser::intermediate_decimal::IntermediateDecimal` and `proof_of_sql::base::math::decimal::Decimal` types, in favor of using `bigdecimal::BigDecimal` along with the `BigDecimalExt` extension trait.

Optional:
[...]

PR #196, opened 2024-09-29
Added the separated big_decimal_ext 

pulkitgovrani:
/claim #113 

Suggested Solution:
Creating a new big_decimal_ext file in the math section inside proof_of_sql folder & pushing the code conversion there.

Problem Faced by me while implementing above solution:
I was getting circular dependency error while doing cargo build because proof-of-sql would become dependent on proof-of-sql-parser and vice versa.

Therefore, to overcome this problem I created a new folder proof-of-sql-utils containing the file code required by both other modules.

Please do share the feedback


JayWhite2357:
> /claim #113
> 
> Suggested Solution: Creating a new big_decimal_ext file in the math section inside proof_of_sql folder & pushing the code conversion there.
> 
> Problem Faced by me while implementing above solution: I was getting circular dependency error while doing cargo build because proof-of-sql would become dependent on proof-of-sql-parser and vice versa.
> 
> Therefore, to overcome this problem I created a new folder proof-of-sql-utils containing the file code required by both other modules.
> 
> Please do share the feedback

It doesn't look like `proof-of-sql-parser` depends on `proof-of-sql-utils` at all as you have it written.
To me this implies that it would be fine putting it in the math section without any circular dependencies since you wouldn't need `proof-of-sql-parser` to depend on `proof-of-sql`.

Am I missing something here?

JayWhite2357:
@pulkitgovrani I believe that this PR is almost complete. It was very close and just needs a bit of updating. Is there any clarification you need at this point?
Outside of possibly minor changes, I think the only thing that needs to get done is resolving conflicts with `main` and resolving this comment: https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/196#issuecomment-2383077727.

JayWhite2357:
@pulkitgovrani Closing because I don't see activity on this. Feel free to reopen when it becomes ready.

PR #198, opened 2024-09-30
feat: add helper methods to dynamic Dory structure (PROOF-919)

jacobtrombetta:
# Rationale for this change
The GPU implementation of the dynamic Dory commitment computation requires a number of helper methods in the dynamic Dory structure. This PR implements methods that will be used by the GPU implementation.

- Add `index_from_row_and_column` method which is the inverse of `row_and_column_from_index` method. Given a `row` and `column` it will provide the `index` of the original data. The method also handles cases where `row` and `column` do not provide a valid `index`.
- Add `matrix_size` method which returns the width and height of the dynamic Dory matrix structure that will hold the given data.
- Made helper methods `const` for compile-time evaluation.
- Add temporary `#[allow(dead_code)]` to address `cargo clippy` warnings - these will be removed when the dynamic Dory commit computation is implemented for the GPU.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
@@ -30,20 +30,55 @@
 //! ...
 //! ```
 
+/// Sets the data to indicate if no index exists.
+const fn no_index() -> usize {
+    usize::MAX
+}
+
+/// Returns if the index exists.
+pub(crate) const fn index_exists(index: usize) -> bool {
+    index != no_index()
+}
+
+/// Returns the full width of a row in the matrix.
+pub(crate) const fn full_width_of_row(row: usize) -> usize {
+    ((2 * row + 4) / 3).next_power_of_two()
+}
+
 /// Returns the index that belongs in the first column in a particular row.
 ///
 /// Note: when row = 1, this correctly returns 0, even though no data belongs there.
 #[cfg(test)]
-pub(crate) fn row_start_index(row: usize) -> usize {
-    let width_of_row = ((2 * row + 4) / 3).next_power_of_two();
+pub(crate) const fn row_start_index(row: usize) -> usize {
+    let width_of_row = full_width_of_row(row);
     width_of_row * (row - width_of_row / 2)
 }
+
 /// Returns the (row, column) in the matrix where the data with the given index belongs.
-pub(crate) fn row_and_column_from_index(index: usize) -> (usize, usize) {
+pub(crate) const fn row_and_column_from_index(index: usize) -> (usize, usize) {
     let width_of_row = 1 << (((2 * index + 1).ilog2() + 1) / 2);
     let row = index / width_of_row + width_of_row / 2;
     let column = index % width_of_row;
     (row, column)
+}
+
+/// Returns the index of data where the (row, column) belongs.
+pub(crate) const fn index_from_row_and_column(row: usize, column: usize) -> usize {
+    let width_of_row = full_width_of_row(row);
+
+    // Return no_index if the (row, column) is not in dynamic Dory structure.
+    if column >= width_of_row || (row == 1 && column == 0) {
+        return no_index();
+    }
+
+    width_of_row * (row - width_of_row / 2) + column
+}
+
+/// Returns a matrix size that can hold the given number of data points being committed with respect to an offset.
+pub(crate) const fn matrix_size(data_len: usize, offset: usize) -> (usize, usize) {
+    let (last_row, _) = row_and_column_from_index(offset + data_len - 1);
+    let width_of_last_row = full_width_of_row(last_row);
+    (width_of_last_row, last_row + 1)
 }
 
 #[cfg(test)]
@@ -124,5 +159,110 @@ mod tests {
                 )
             }
         }
+    }
+    #[test]
+    fn we_can_correctly_identify_index_existence() {
+        assert_eq!(index_exists(0), true);
+        assert_eq!(index_exists(no_index()), false);
+        assert_eq!(index_exists(no_index() - 1), true);
+    }
+    #[test]
+    fn we_can_find_the_full_width_of_row() {
+        let width_of_row = full_width_of_row(0);
+        assert_eq!(width_of_row, 1);
+
+        let width_of_row = full_width_of_row(1);
+        assert_eq!(width_of_row, 2);
+
+        let width_of_row = full_width_of_row(4);
+        assert_eq!(width_of_row, 4);
+
+        let width_of_row = full_width_of_row(7);
+        assert_eq!(width_of_row, 8);
+
+        let width_of_row = full_width_of_row(11);
+        assert_eq!(width_of_row, 8);
+
+        let width_of_row = full_width_of_row(12);
+        assert_eq!(width_of_row, 16);
+
+        let width_of_row = full_width_of_row(13);
+        assert_eq!(width_of_row, 16);
+    }
+    #[test]
+    fn we_can_produce_the_correct_matrix_size() {
+        let (width, height) = matrix_size(2, 0);
+        assert_eq!(width, 2);
+        assert_eq!(height, 2);
+
+        let (width, height) = matrix_size(3, 0);
+        assert_eq!(width, 2);
+        assert_eq!(height, 3);
+
+        let (width, height) = matrix_size(4, 0);
+        assert_eq!(width, 2);
+        assert_eq!(height, 3);
+
+        let (width, height) = matrix_size(16, 0);
+        assert_eq!(width, 4);
+        assert_eq!(height, 6);
+
+        let (width, height) = matrix_size(17, 0);
+        assert_eq!(width, 8);
+        assert_eq!(height, 7);
+
+        let (width, height) = matrix_size(64, 0);
+        assert_eq!(width, 8);
+        assert_eq!(height, 12);
+
+        let (width, height) = matrix_size(71, 0);
+        assert_eq!(width, 16);
+        assert_eq!(height, 13);
+    }
+    #[test]
+    fn we_can_produce_the_correct_matrix_size_with_offset() {
+        let (width, height) = matrix_size(2, 1);
+        assert_eq!(width, 2);
+        assert_eq!(height, 3);
+
+        let (width, height) = matrix_size(16, 1);
+        assert_eq!(width, 8);
+        assert_eq!(height, 7);
+
+        let (width, height) = matrix_size(64, 1);
+        assert_eq!(width, 16);
+        assert_eq!(height, 13);
+
+        let (width, height) = matrix_size(4, 16);
+        assert_eq!(width, 8);
+        assert_eq!(height, 7);
+    }
+    #[test]
+    fn we_can_find_index_from_row_and_column() {
+        for i in 0..1 << 10 {
+            let (row, column) = row_and_column_from_index(i);
+            let index = index_from_row_and_column(row, column);
+            assert_eq!(index, i);
+        }
+    }
+    #[test]
+    fn we_can_find_index_from_row_and_column_not_in_dynamic_dory_structure() {
+        let index = index_from_row_and_column(0, 1);
+        assert_eq!(index, no_index());
+
+        let index = index_from_row_and_column(1, 0);
+        assert_eq!(index, no_index());
+
+        let index = index_from_row_and_column(1, 2);
+        assert_eq!(index, no_index());
+
+        let index = index_from_row_and_column(3, 4);
+        assert_eq!(index, no_index());
+
+        let index = index_from_row_and_column(6, 8);
+        assert_eq!(index, no_index());
+
+        let index = index_from_row_and_column(12, 16);
+        assert_eq!(index, no_index());
     }
 }
--- a/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
@@ -64,9 +64,11 @@ pub(crate) fn index_from_row_and_column(row: usize, column: usize) -> Option<usi
 }
 
 /// Returns a matrix size, (height, width), that can hold the given number of data points being committed with respect to an offset.
+///
+/// Note: when data_len = 0 and offset = 0, this function returns an empty matrix with size (0, 0).
 #[allow(dead_code)]
 pub(crate) const fn matrix_size(data_len: usize, offset: usize) -> (usize, usize) {
-    if data_len == 0 {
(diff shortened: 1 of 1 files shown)


crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -30,22 +30,62 @@
...
+pub(crate) const fn index_from_row_and_column(row: usize, column: usize) -> usize {
+    let width_of_row = full_width_of_row(row);
+
+    // Return no_index if the (row, column) is not in dynamic Dory structure.
+    if column >= width_of_row || (row == 1 && column == 0) {
+        return no_index();
+    }
+
+    width_of_row * (row - width_of_row / 2) + column
+}
```
JayWhite2357, line 79: This is a textbook example for why the `Option` type exists in Rust.
```suggestion
pub(crate) const fn index_from_row_and_column(row: usize, column: usize) -> Option<usize> {
    let width_of_row = full_width_of_row(row);

    // Return no_index if the (row, column) is not in dynamic Dory structure.
    if column >= width_of_row || (row == 1 && column == 0) {
        return None;
    } else {
        Some(width_of_row * (row - width_of_row / 2) + column)
    }
}
```

Or, equivalently;

```rust
pub(crate) const fn index_from_row_and_column(row: usize, column: usize) -> Option<usize> {
    let width_of_row = full_width_of_row(row);
    (column < width_of_row && (row, column) != (1, 0))
        .then_some(width_of_row * (row - width_of_row / 2) + column)
}
```

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -30,22 +30,62 @@
...
+    usize::MAX
+}
+
+/// Returns if the index exists.
+#[allow(dead_code)]
+pub(crate) const fn index_exists(index: usize) -> bool {
+    index != no_index()
+}
+
+/// Returns the full width of a row in the matrix.
```
JayWhite2357, line 45: Can you add docs for the behavior for row #0 and row#1, since those are unusual?

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -30,22 +30,62 @@
...
-pub(crate) fn row_start_index(row: usize) -> usize {
-    let width_of_row = ((2 * row + 4) / 3).next_power_of_two();
+pub(crate) const fn row_start_index(row: usize) -> usize {
+    let width_of_row = full_width_of_row(row);
     width_of_row * (row - width_of_row / 2)
 }
+
 /// Returns the (row, column) in the matrix where the data with the given index belongs.
-pub(crate) fn row_and_column_from_index(index: usize) -> (usize, usize) {
+pub(crate) const fn row_and_column_from_index(index: usize) -> (usize, usize) {
```
JayWhite2357, line 61: 👍 Nice

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -125,4 +165,109 @@ mod tests {
...
+
+        let index = index_from_row_and_column(3, 4);
+        assert_eq!(index, no_index());
+
+        let index = index_from_row_and_column(6, 8);
+        assert_eq!(index, no_index());
+
+        let index = index_from_row_and_column(12, 16);
+        assert_eq!(index, no_index());
+    }
```
JayWhite2357, line 272: We could make this more thorough by showing that everything _NOT_ generated by `row_and_column_from_index` returns no index.

Here's Copilot's (AI's) best attempt at that. Obviously needs a lot of work, but my prompt was:
> Can you write a test that will create an IndexSet of all possible row,columns that could be created from the index. Then, test that all the other pairs generate no index.
```suggestion
#[test]
fn test_all_possible_row_column_pairs() {
    use std::collections::HashSet;

    const MAX_INDEX: usize = 1 << 10; // You may need to adjust this value based on your requirements
    let mut valid_pairs = HashSet::new();

    // Collect all valid (row, column) pairs
    for i in 0..MAX_INDEX {
        let (row, column) = row_and_column_from_index(i);
        valid_pairs.insert((row, column));
    }

    // Check that all valid pairs generate a valid index
    for &(row, column) in &valid_pairs {
        let index = index_from_row_and_column(row, column);
        assert_ne!(index, no_index(), "Valid pair ({}, {}) generated no index", row, column);
    }

    // Check that all other pairs generate no_index
    for row in 0..MAX_INDEX {
        for column in 0..full_width_of_row(row) {
            if !valid_pairs.contains(&(row, column)) {
                let index = index_from_row_and_column(row, column);
                assert_eq!(index, no_index(), "Invalid pair ({}, {}) generated a valid index", row, column);
            }
        }
    }
}
```

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -125,4 +165,109 @@ mod tests {
...
+        assert_eq!(width_of_row, 8);
+
+        let width_of_row = full_width_of_row(11);
+        assert_eq!(width_of_row, 8);
+
+        let width_of_row = full_width_of_row(12);
+        assert_eq!(width_of_row, 16);
+
+        let width_of_row = full_width_of_row(13);
+        assert_eq!(width_of_row, 16);
```
JayWhite2357, line 195: Can you do this exhaustively? I want to see that this works for rows 0..15 at least. imo, row 2 is still almost certainly an edge case.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -125,4 +165,109 @@ mod tests {
...
+        assert_eq!(height, 7);
+
+        let (width, height) = matrix_size(64, 0);
+        assert_eq!(width, 8);
+        assert_eq!(height, 12);
+
+        let (width, height) = matrix_size(71, 0);
+        assert_eq!(width, 16);
+        assert_eq!(height, 13);
+    }
```
JayWhite2357, line 226: Can you do this exhaustively? I'd like to see it at least 0..16 with some bigger numbers like you have here as well.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -125,4 +165,109 @@ mod tests {
...
+    #[test]
+    fn we_can_produce_the_correct_matrix_size() {
+        let (width, height) = matrix_size(2, 0);
+        assert_eq!(width, 2);
+        assert_eq!(height, 2);
+
+        let (width, height) = matrix_size(3, 0);
+        assert_eq!(width, 2);
+        assert_eq!(height, 3);
```
JayWhite2357, line 205: NIT suggestion: Less verbose, but far less tedious to write (and read).
```suggestion
        // NOTE: returned tuple is (width, height).
        assert_eq!(matrix_size(2, 0), (2, 2));
        assert_eq!(matrix_size(3, 0), (2, 3));
```

JayWhite2357 (changes requested):
Looks good. A couple minor changes and suggestions.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
jacobtrombetta: Updated, thank you!

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
jacobtrombetta: I added a note, similar to the one added to the `row_start_index` function. I am assuming that is the extent of docs to be added? Do we want this documented elsewhere?

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
jacobtrombetta: Good idea, added.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
jacobtrombetta: I listed test cases for `0..16` explicitly with a few extra cases. I could only think of recreating the algorithm in the test case to do it exhaustively for a larger number, which didn't seem appropriate for a test case. Should I spend some more time thinking of an exhaustive `0..N` test?

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -67,23 +57,22 @@ pub(crate) const fn row_and_column_from_index(index: usize) -> (usize, usize) {
...
 #[allow(dead_code)]
 pub(crate) const fn matrix_size(data_len: usize, offset: usize) -> (usize, usize) {
+    if data_len == 0 {
+        return (0, 0);
+    }
+
     let (last_row, _) = row_and_column_from_index(offset + data_len - 1);
     let width_of_last_row = full_width_of_row(last_row);
-    (width_of_last_row, last_row + 1)
+    (last_row + 1, width_of_last_row)
```
JayWhite2357, line 75: I didn't mean to make you switch it. But, (height, width) feels better, so I'm not complaining.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -67,23 +57,22 @@ pub(crate) const fn row_and_column_from_index(index: usize) -> (usize, usize) {
...
+        .then_some(width_of_row * (row - width_of_row / 2) + column)
 }
 
-/// Returns a matrix size that can hold the given number of data points being committed with respect to an offset.
+/// Returns a matrix size, (height, width), that can hold the given number of data points being committed with respect to an offset.
 #[allow(dead_code)]
 pub(crate) const fn matrix_size(data_len: usize, offset: usize) -> (usize, usize) {
+    if data_len == 0 {
+        return (0, 0);
+    }
```
JayWhite2357, line 71: Is this consistent with the other methods?
This is certainly an edge case that I would document.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -67,23 +57,22 @@ pub(crate) const fn row_and_column_from_index(index: usize) -> (usize, usize) {
 
 /// Returns the index of data where the (row, column) belongs.
 #[allow(dead_code)]
-pub(crate) const fn index_from_row_and_column(row: usize, column: usize) -> usize {
+pub(crate) fn index_from_row_and_column(row: usize, column: usize) -> Option<usize> {
```
JayWhite2357, line 60: NIT: Is there a reason you removed `const`?

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
JayWhite2357: I feel like this is mixing two concepts: the "trimmed" matrix (where all the top rows made up of just 0s caused by the offset are removed) and the "untrimmed" matrix.

Both may be useful, and I'm not sure which one you actually want.

What this is doing, when `data_len` is 0, it is always the "trimmed" size. Otherwise, it is the "untrimmed" size. This is confusing and could lead to upstream bugs or more complex code.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
```diff
@@ -166,108 +155,129 @@ mod tests {
...
-        assert_eq!(width, 2);
-        assert_eq!(height, 3);
-
-        let (width, height) = matrix_size(16, 1);
-        assert_eq!(width, 8);
-        assert_eq!(height, 7);
+        // NOTE: returned tuple is (height, width).
+        for i in 0..100 {
+            assert_eq!(matrix_size(0, i), (0, 0));
+        }
```
JayWhite2357, line 217: See my point above. This is strange behavior, considering that `matrix_size(a, b)==matrix_size(c, d)` iff `a+b==c+d` in every case except this one.

JayWhite2357 (changes requested):
Only non-nit is the `matrix_size` behavior. I think the old code was fine with a case needed to handle `len+offset==0`. (Although, it may be better to simply drop the `offset` parameter and just call it with `matrix_size(len+offset)` instead of `matrix_size(len,offset)`.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
JayWhite2357: No, don't worry about an exhaustive test. I just meant explicitly.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
jacobtrombetta: Yes, because of compilation errors it introduced. Adding the return `Option` and keeping `const` on the function produced two errors:

1. `error[E0015]: cannot call non-const operator in constant functions` referring to the `(row, column) != (1, 0)`
2. `error[E0015]: cannot call non-const fn __core::bool::<impl bool>::then_some::<usize> in constant functions` referring to the `.then_some(width_of_row * (row - width_of_row / 2) + column)`

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
jacobtrombetta: `matrix_size` needs to make sure both `data_len` and `offset` aren't `0` to not panic at the line `row_and_column_from_index(offset + data_len - 1)`. My thought was if a caller asked what size of a matrix they need to hold no data, i.e. `data_len == 0`, it was a degenerate case and the result should be an empty matrix regardless of the offset. I see your point about mixing trimmed and untrimmed matrices.

I've updated the code to only return `(0, 0)` when `data_len == offset == 0`. So the method now returns an untrimmed matrix size. I've also added a note to the doc.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_structure.rs
jacobtrombetta: The code has been updated to be `matrix_size(a, b)==matrix_size(c, d)` iff `a+b==c+d` in every case.

PR #201, opened 2024-10-01
refactor(parser): enable clippy pedantic

mehulmathur16:
Part of #195 

This PR fixes clippy pedantic warnings for `proof-of-sql-parser` lib.

## Commit Structure

`autofix | manual fix`: `lint_name` 
`lint_name` refers to clippy lint names as given here (https://rust-lang.github.io/rust-clippy/master/index.html#/)

## Testing this change

Run `cargo clippy --lib -p proof-of-sql-parser -- -W clippy::pedantic`
It should compile without any warnings.


mehulmathur16:
Hi @JayWhite2357, can you please review this PR?

crates/proof-of-sql-parser/src/identifier.rs
```diff
@@ -12,28 +12,35 @@ pub struct Identifier {
 impl Identifier {
     /// Constructor for [Identifier]
     ///
-    /// Note: this constructor should be private within the proof_of_sql_parser crate.
+    /// Note: this constructor should be private within the `proof_of_sql_parser` crate.
     /// This is necessary to guarantee that no one outside the crate
-    /// can create Names, thus securing that ResourceIds and Identifiers
+    /// can create Names, thus securing that `ResourceIds` and Identifiers
```
iajoiner, line 17: `Identifier` needs backticks

crates/proof-of-sql-parser/src/lib.rs
```diff
@@ -36,7 +36,7 @@ pub use resource_id::ResourceId;
 // lalrpop-generated code is not clippy-compliant
 lalrpop_mod!(#[allow(clippy::all, missing_docs, clippy::missing_docs_in_private_items, clippy::pedantic)] pub sql);
 
-/// Implement Deserialize through FromStr to avoid invalid identifiers.
+/// Implement Deserialize through `FromStr` to avoid invalid identifiers.
```
JayWhite2357, line 40: It might be hard to catch them all, but as much as possible try to find all the spots. In this case, it should be `Deserialize`, or even better [`Deserialize`].

crates/proof-of-sql-parser/src/error.rs
```diff
@@ -17,13 +17,13 @@ pub enum ParseError {
...
-    /// Can not parse the resource_id
+    /// Can not parse the `resource_id`
     ResourceIdParseError {
         /// The underlying error
         error: String,
     },
 }
 
-/// General parsing error that may occur, for example if the provided schema/object_name strings
+/// General parsing error that may occur, for example if the provided `schema/object_name` strings
```
JayWhite2357, line 28:
```suggestion
/// General parsing error that may occur, for example if the provided `schema`/`object_name` strings
```

crates/proof-of-sql-parser/src/identifier.rs
JayWhite2357:
```suggestion
    /// can create Names, thus securing that `ResourceId`s and `Identifier`s
```
Or, even better:
```suggestion
    /// can create Names, thus securing that [`ResourceId`]s and [`Identifier`]s
```

crates/proof-of-sql-parser/src/lib.rs
JayWhite2357: Would be nice if you can identify some of the ones that clippy doesn't automatically catch. That's a bit beyond the scope of the lints, so up to you how thorough you want to be here.

```suggestion
/// Implement `Deserialize` through `FromStr` to avoid invalid identifiers.
```

Or

```suggestion
/// Implement [`Deserialize`] through [`FromStr`] to avoid invalid identifiers.
```

JayWhite2357 (changes requested):
Looks good to me.
The only change I would like to see is a bit more attention spend on the `doc_markdown` lint. The autofix does a decent job but messes some things up.

For future reference (i.e., when you do the `proof-of-sql` crate, it would be preferable from a review perspective to have the manual fixes of this be done _before_ the autofix so that the autofix actually does everything correct and the individual commit is good looking.

mehulmathur16:
Sure, @JayWhite2357. I'll keep that in mind and have addressed the requested changes.

JayWhite2357 (approved):
Looks good to me. Thanks!

JayWhite2357:
FYI: @mehulmathur16 I rebased your branch instead of merging in main. Additionally, I cleaned up the commit messages so that they adhere to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0).

PR #206, opened 2024-10-02
feat: bump `lalrpop` to 0.22.0 and remove patch from the no std check in CI

iajoiner:
# Rationale for this change
Bump `lalrpop` to support https://github.com/lalrpop/lalrpop/pull/957

- Bump `lalrpop` to 0.22.0
- Remove patch from the no std check in CI

# Are these changes tested?
Yes


Dustin-Ray (approved):
LGTM - Thank you

PR #207, opened 2024-10-02
refactor!: expose random public parameter generation

Dustin-Ray:
# Rationale for this change

If PublicParameters are generated within sxtDB and not an argument passed in via protobuf, then we need a way to randomly generate the setup within external crates such as sxtdb. This PR exposes a trait bounded method requiring a CSRNG which generates the public parameters securely.

- [x] exposes a crpytographically secure method for public parameter generation
- [x] rename existing rand to ```test_rand```
- [x] refactor feature gates  

# Are these changes tested?

existing tests pass


crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -21,14 +23,33 @@ pub struct PublicParameters {
...
+    }
+
     /// Generate random public parameters for testing purposes.
     pub fn rand<R>(max_nu: usize, rng: &mut R) -> Self
     where
         R: ark_std::rand::Rng + ?Sized,
     {
         use ark_std::UniformRand;
-        let (Gamma_1, Gamma_2) = super::rand_G_vecs(max_nu, rng);
+        let (Gamma_1, Gamma_2) = Self::rand_G_vecs(max_nu, rng);
```
Dustin-Ray, line 52: rand_G_vecs was originally gated behind test so we were not able to call this function externally

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -21,14 +23,33 @@ pub struct PublicParameters {
...
+        Self {
+            Gamma_1,
+            Gamma_2,
+            H_1,
+            H_2,
+            Gamma_2_fin,
+            max_nu,
+        }
+    }
+
```
Dustin-Ray, line 45: instead of changing the visibility of the fields from pub(super) we can just add a constructor here

crates/proof-of-sql/src/proof_primitive/dory/rand_util.rs
```diff
@@ -16,6 +16,7 @@ pub fn test_seed_rng(seed: [u8; 32]) -> impl Rng {
     StdRng::from_seed(seed)
 }
 
+#[allow(dead_code)]
```
Dustin-Ray, line 19: Im not sure how to fix this, there is a web of features and re exports that Im not really familiar with here. All other checks succeed, clippy just complains about this one spot:

```bash
cargo clippy --all-targets --all-features -- -D warnings
```

```
error: unused import: `rand_util::rand_G_vecs`
  --> crates/proof-of-sql/src/proof_primitive/dory/mod.rs:29:5
   |
29 | use rand_util::rand_G_vecs;
   |     ^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(unused_imports)]`

error: function `rand_G_vecs` is never used
  --> crates/proof-of-sql/src/proof_primitive/dory/rand_util.rs:20:8
   |
20 | pub fn rand_G_vecs<R>(nu: usize, rng: &mut R) -> (Vec<G1Affine>, Vec<G2Affine>)
   |        ^^^^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`

error: could not compile `proof-of-sql` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -21,14 +25,19 @@ pub struct PublicParameters {
 }
 
 impl PublicParameters {
+    /// Generate cryptographically secure random public parameters.
+    pub fn crypto_rand<R: CryptoRng + Rng + ?Sized>(max_nu: usize, rng: &mut R) -> Self {
```
JayWhite2357, line 29: This `CryptoRng` enforces cryptographic security, so there isn't a need to name it differently.
```suggestion
    #[cfg(not(any(test, feature = "test")))]
    /// Generate random public parameters.
    pub fn rand<R: CryptoRng + Rng + ?Sized>(max_nu: usize, rng: &mut R) -> Self {
```

If you really don't like this, I would name this `rand` and the test version `test_rand`. But, that is a breaking change.

crates/proof-of-sql/src/proof_primitive/dory/rand_util.rs
JayWhite2357, line 19:
```suggestion
#[cfg(test)]
```
then, hide any newly unused imports behind `#[cfg(test)]` as well.

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
Dustin-Ray: I tried that originally and it produced a rather large diff, this seemed like the least impactful way to add this in

JayWhite2357: You tried the suggestion? Or renaming it to `test_rand`. That code I suggested should be the only change needed.

JayWhite2357 (approved):
LGTM

PR #208, opened 2024-10-03
feat: create a script to run all of the checks in the workflow file

Dustin-Ray:
# Rationale for this change

To make it easier for external contributors, this PR adds a script at the head of the repo that reads in the current github workflow file, grabs all of the cargo jobs that are not tests, and runs them sequentially, emulating the CI process. This script allows users to run the entire check workflow with a single bash script. A separate script can be produced for running the tests in a similar fashion. 

Just a script that lives in the repo to execute the workflows in ``` ".github/workflows/lint-and-test.yml" ```

# Are these changes tested?

yes


Dustin-Ray:
@JayWhite2357 @iajoiner let me know what you think about this. Its a step towards updating the CONTRIBUTING.md document

JayWhite2357 (commented):
Makes sense.
My concern is that there are things that won't run easily on a vanilla setup. In particular, I'm pretty sure the script won't run on a non-linux machine.

Dustin-Ray:
> Makes sense. My concern is that there are things that won't run easily on a vanilla setup. In particular, I'm pretty sure the script won't run on a non-linux machine.

Oh yeah good point. This should work on mac and linux but maybe a powershell version for windows users in a scripts folder or something?

JayWhite2357:
> Oh yeah good point. This should work on mac and linux but maybe a powershell version for windows users and move them all into a scripts folder or something?

My concern has more to do with the fact that `blitzar` only works on linux. As long as we document that, it should be fine, I guess. But, we should indicate that `BLITZAR_BACKEND=cpu` needs to be set in order to run it without appropriate GPU drivers.

run_ci.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+# Exit immediately if a command exits with a non-zero status
+set -e
+
```
JayWhite2357, line 5:
```suggestion

# Display a help text
[ "$1" = "-h" -o "$1" = "--help" ] && echo "Runs all CI checks (excluding tests)." && exit

```

run_ci.sh
```diff
@@ -0,0 +1,40 @@
...
+# Check if the YAML file exists
+if [ ! -f "$YAML_FILE" ]; then
+    echo "YAML file $YAML_FILE does not exist."
+    exit 1
+fi
+
+# Extract all lines that contain 'cargo' commands from the YAML file, 
+# excluding ones with '--ignored' or 'test'
+cargo_commands=$(grep -E '^\s*run:.*cargo' "$YAML_FILE" | grep -v -- '--ignored' | grep -v 'test' | sed -E 's/^\s*run:\s*//')
```
JayWhite2357, line 17: This is the output I get:
```bash
curl https://sh.rustup.rs -sSf | bash -s -- -y --profile minimal && source ~/.cargo/env
cargo check --no-default-features
cargo check --all-targets
cargo check --all-targets --all-features
cargo check -p proof-of-sql --no-default-features
cargo check -p proof-of-sql --all-features
cargo check -p proof-of-sql --no-default-features --features="blitzar"
cargo check -p proof-of-sql --no-default-features --features="arrow"
cargo check -p proof-of-sql --no-default-features --features="rayon"
cargo check -p proof-of-sql --no-default-features --features="perf"
cargo check -p proof-of-sql --no-default-features --features="std"
curl https://sh.rustup.rs -sSf | bash -s -- -y --profile minimal && source ~/.cargo/env
cargo clippy --all-targets --all-features -- -D warnings
cargo fmt --all -- --config imports_granularity=Crate,group_imports=One --check
cargo +nightly udeps --all-targets
```

* This isn't the full CI since it's missing tests. So, we should rename this file to be more descriptive.
    * I see now that the last line says this, but it wasn't obvious to me immediately.
* I don't think we want the `rustup` commands here.

JayWhite2357 (changes requested):

JayWhite2357:
@Dustin-Ray My concerns are `assuaged` since this isn't running tests. This looks good other than my requests on being more clear about the fact that this isn't running tests.

scripts/run_ci_checks.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# Exit immediately if a command exits with a non-zero status
+set -e
+
+# Display a help text
+[ "$1" = "-h" -o "$1" = "--help" ] && echo "Runs all CI checks (excluding tests)." && exit
+
+# The path to the YAML file that defines the CI workflows
+YAML_FILE=".github/workflows/lint-and-test.yml"
```
iajoiner, line 10: We need to make sure this runs when someone is not exactly at the dir root

iajoiner:
```bash
# Initialize the directory we're searching from (current directory)
current_dir=$(pwd)

# Traverse upwards to find the root directory, assuming it exists somewhere above
while [[ ! -f "$current_dir/sxt-proof-of-sql/.github/workflows/lint-and-test.yml" ]]; do
  # Move up one directory
  current_dir=$(dirname "$current_dir")
  
  # If we reach the root directory (i.e., /), stop to prevent an infinite loop
  if [[ "$current_dir" == "/" ]]; then
    echo "Could not find file."
    exit 1
  fi
done
```

iajoiner (approved):
Thanks @Dustin-Ray ! This is nice!

scripts/run_ci_checks.sh
iajoiner: Never mind since a script shouldn't be too smart in the sense of trying to locate things unless necessary.

JayWhite2357 (approved):

PR #210, opened 2024-10-03
feat: add Solidity sumcheck verifier

JayWhite2357:
# Rationale for this change

We want a Solidity verifier. This is the first step.

* A solidity verifier is added
* Tests are added to ensure it behaves like the Rust verifier

# Are these changes tested?

Yes


JayWhite2357:
This PR depends on #205. The [last three commits](https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/210/files/14ae329eb9d2621dd9c51d473f5abc8c91bb9e80..9b08c90943e95bf89038f12964cea10bfa49a609) are the changes specific to this PR.

crates/proof-of-sql/sol_src/proof_primitive/sumcheck/Sumcheck.sol
JayWhite2357, line 1: @rnburn 
This is the primary thing that I would like you to review. If you want to review the other stuff, you can, but this is the main payload.

crates/proof-of-sql/sol_src/proof_primitive/sumcheck/Sumcheck.sol
```diff
@@ -0,0 +1,53 @@
...
+        uint256 claimedSum,
+        uint256 modulus
+    ) public pure returns (Subclaim memory result, bytes32 resultTranscript) {
+        if (proof.coefficients.length != numVariables * (degree + 1)) revert InvalidProofLength();
+        result.evaluationPoint = new uint256[](numVariables);
+        transcript = keccak256(abi.encodePacked(transcript, degree, numVariables));
+        for (uint64 roundIndex = 0; roundIndex < numVariables; ++roundIndex) {
+            uint256 startIndex = roundIndex * (degree + 1);
+            transcript = keccak256(abi.encodePacked(transcript, proof.coefficients[startIndex:startIndex + degree + 1]));
+            uint256 roundEvaluationPoint = uint256(transcript);
```
rnburn, line 31: Is this intended to be a custom transcript protocol?

In the merlin protocol, drawing a challenge would also update the state of the transcript.

crates/proof-of-sql/sol_src/proof_primitive/sumcheck/Sumcheck.sol
```diff
@@ -0,0 +1,53 @@
...
+            result.evaluationPoint[roundIndex] = roundEvaluationPoint % modulus;
```
waelsy123, line 32: how do we insure the randomness of this chosen point?

crates/proof-of-sql/sol_src/proof_primitive/sumcheck/Sumcheck.sol
```diff
@@ -0,0 +1,53 @@
...
+        uint256 expectedEvaluation;
+    }
+
+    function verifyWithoutEvaluation(
+        Proof calldata proof,
+        bytes32 transcript,
+        uint64 degree,
+        uint64 numVariables,
+        uint256 claimedSum,
+        uint256 modulus
```
waelsy123, line 23: What domain are we assuming for the variable values here?

crates/proof-of-sql/sol_src/proof_primitive/sumcheck/Sumcheck.sol
JayWhite2357: Here's the Rust implementation of the same thing:
https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/crates/proof-of-sql/src/base/proof/keccak256_transcript.rs

The intent here is that drawing a challenge always occurs after anything is appended to the transcript. That way, there is only one hash per challenge/response round. So, the current state of the transcript _is_ the current challenge.

If I need two challenges in a row, that would mean the prover should send an empty message and we would do: `transcript = keccak256(abi.encodePacked(transcript));`

crates/proof-of-sql/sol_src/proof_primitive/sumcheck/Sumcheck.sol
JayWhite2357: See my response to Ryan. The point is random because it is a hash of the previous transcript along with the previous prover message.

crates/proof-of-sql/sol_src/proof_primitive/sumcheck/Sumcheck.sol
JayWhite2357: They can be anything. The `modulus` needs to match whatever the prover is using, but otherwise, anything is possible.
Large values for `degree` and `numVariables` are a bit absurd. `degree` is usually around `3` and `numVaraibles` will almost certainly never be larger than `40`, but there's nothing wrong with those values.

waelsy123: do we have any test for that for domain (0,1) ?

waelsy123 (approved):

iajoiner (approved):

PR #221, opened 2024-10-05
feat: read and write prover public setup parameters from file

Dustin-Ray:
# Rationale for this change

The ```PublicParameters``` setup phase involves the calculation of multiple large vectors of group elements in order to instantiate the argument system. The setup phase can potentially be costly in some configurations. The setup phase is transparent and can be cached for later use. This PR adds support for file i/o  for ```PublicParameters``` along with serde.

Currently the file i/o and serialization are gated behind std in order to conform to a ```no_std``` requirement. There might be a more appropriate way to handle this, but currently this allows std enabled test code to easily generate a setup of a configurable size.

- [x] use ark_serialize to implement serde over Vec<G1/G2> affine
- [x] read and write serialization from a file
- [x] verify no_std compatibility or gate file generation feature behind std feature perhaps since file ops are trickier in no_std

# Are these changes tested?

- [x] round trip serialization and expected values
- [x] round trip serialization + file i/o with benches


crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -50,4 +60,277 @@ impl PublicParameters {
             Gamma_2_fin,
         }
     }
+    #[cfg(feature = "std")]
+    /// Function to save `PublicParameters` to a file in binary form
+    pub fn save_to_file(&self, path: &Path) -> std::io::Result<()> {
+        use std::io;
```
iajoiner, line 66: It's probably more clear if you put all the use declarations together (that is, at the top of the file) unless there is a very good reason to not do so.

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -50,4 +60,275 @@ impl PublicParameters {
...
+    fn serialized_size(&self, compress: ark_serialize::Compress) -> usize {
+        let mut size = 0;
+
+        // Size of max_nu
+        size += 8; // u64 is 8 bytes
+
+        // Calculate size of Gamma_1 (Vec<G1Affine>)
+        for g1 in &self.Gamma_1 {
+            size += g1.serialized_size(compress);
+        }
```
iajoiner, line 145: This can be more functional

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -50,4 +60,275 @@ impl PublicParameters {
...
+
+        // Calculate size of Gamma_2 (Vec<G2Affine>)
+        for g2 in &self.Gamma_2 {
+            size += g2.serialized_size(compress);
+        }
```
iajoiner, line 150: Same

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -50,4 +60,275 @@ impl PublicParameters {
...
+
+        // Deserialize Gamma_1 (Vec<G1Affine>)
+        let mut Gamma_1 = Vec::with_capacity(1 << max_nu);
+        for _ in 0..(1 << max_nu) {
+            Gamma_1.push(G1Affine::deserialize_with_mode(
+                &mut reader,
+                compress,
+                validate,
+            )?);
+        }
```
iajoiner, line 184: You can make this functional too

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -50,4 +60,275 @@ impl PublicParameters {
...
+
+        // Deserialize Gamma_2 (Vec<G2Affine>)
+        let mut Gamma_2 = Vec::with_capacity(1 << max_nu);
+        for _ in 0..(1 << max_nu) {
+            Gamma_2.push(G2Affine::deserialize_with_mode(
+                &mut reader,
+                compress,
+                validate,
+            )?);
+        }
```
iajoiner, line 194: Same

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -50,4 +60,275 @@ impl PublicParameters {
             max_nu,
         }
     }
+    #[cfg(feature = "std")]
+    /// Function to save `PublicParameters` to a file in binary form
```
iajoiner, line 64: @Dustin-Ray Do you expect to have people inspect the content of such files? If so text (i.e. strings) are better than binary.

Dustin-Ray: I do not expect users to visually inspect the contents of this file, the reason being that these will be filled with opaque cryptographic group elements that have little context or meaning from visual inspection on their own. Further, printing them as strings will yield the unwieldy string representations used by the arkff crate which usually is printed as an array of u64 values in montgomery form. To make any sense of these values, they must be converted first to standard form, then a string representation is needed to directly print the value. I am unaware of such a method string method is use anywhere in our code. I am aware of a method that we use to parse strings *into* group elements, but not necessarily the other way around.

Lastly, the overhead of string parsing large setups on the order of nu >= 20 also seems untenable in my opinion. It is already going to be rather expensive to generate these larger setups. Adding string parsing to this process will create an undue burden not only in terms of efficiency but also in terms of reliability. Given our time constraints and priorities, I do not have the bandwidth to implement string parsing and writing, as well as the checks involved with ensuring that process is correct.

For visual inspection, one can read in the file and deserialize it, then choose the best way to visually represent the group elements as they see fit in rust code.

If visual string based inspections is something that is heavily requested, it might be something we could create a bounty on. However, I am not aware of any major chain employing the use of a proof system such as Groth16 in zcash, for example, taking the time and effort to add string parsing into their public setups.

iajoiner (approved):
Really thanks @Dustin-Ray !

PR #222, opened 2024-10-05
refactor: move column operations to `ColumnOperation` trait

ssddOnTop:
# Rationale for this change

Now that we are including nullable column support in #183, the comment https://github.com/spaceandtimelabs/sxt-proof-of-sql/issues/183#issuecomment-2379341988 suggests creating new structs for nullable columns

This trait will help us with the easy implementation of methods

I introduced a trait `ColumnOperation` which contains the functions `try_add_subtract_column_types`, `try_multiply_column_types` and `try_divide_column_types`

# Are these changes tested?

The refactor does not affect any core logic, and changes made are covered under the tests.

PS: ext of #194

### DIFF
--- a/crates/proof-of-sql/src/base/database/column_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_operation.rs
@@ -17,44 +17,163 @@ use proof_of_sql_parser::intermediate_ast::BinaryOperator;
 // For decimal type manipulation please refer to
 // https://learn.microsoft.com/en-us/sql/t-sql/data-types/precision-scale-and-length-transact-sql?view=sql-server-ver16
 
-/// Determine the output type of an add or subtract operation if it is possible
-/// to add or subtract the two input types. If the types are not compatible, return
-/// an error.
-///
-/// # Panics
-///
-/// - Panics if `lhs` or `rhs` does not have a precision or scale when they are expected to be numeric types.
-/// - Panics if `lhs` or `rhs` is an integer, and `lhs.max_integer_type(&rhs)` returns `None`.
-pub fn try_add_subtract_column_types(
-    lhs: ColumnType,
-    rhs: ColumnType,
-    operator: BinaryOperator,
-) -> ColumnOperationResult<ColumnType> {
-    if !lhs.is_numeric() || !rhs.is_numeric() {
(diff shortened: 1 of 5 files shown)


ssddOnTop:
@JayWhite2357 can I get a reviewer assigned

crates/proof-of-sql/src/base/database/column_operation.rs
```diff
@@ -17,44 +17,163 @@ use proof_of_sql_parser::intermediate_ast::BinaryOperator;
...
-    operator: BinaryOperator,
-) -> ColumnOperationResult<ColumnType> {
-    if !lhs.is_numeric() || !rhs.is_numeric() {
-        return Err(ColumnOperationError::BinaryOperationInvalidColumnType {
-            operator,
-            left_type: lhs,
-            right_type: rhs,
-        });
+/// A trait for column operations.
+pub trait ColumnOperation {
```
iajoiner, line 21: @ssddOnTop Hmm this looks like a misnomer. What's your goal here, may I ask?

ssddOnTop: @iajoiner I might be confused here but I was thinking that now we will have the normal `Column` and the `NullableColumn` so we could implement a trait for each type so that we could perform dynamic dispatch to avoid massive changes in future.

JayWhite2357:
@ssddOnTop are you waiting on us here? I don't see activity on this.

ssddOnTop:
> @ssddOnTop are you waiting on us here? I don't see activity on this.

I was waiting for a reply on https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/222#discussion_r1789339526

I am finding it hard to communicate here..

crates/proof-of-sql/src/base/database/column_operation.rs
JayWhite2357: I think @iajoiner's point here is that you have `impl ColumnOperation for ColumnType`. So, it probably should be named `ColumnTypeOperation`.

My instinct is that a trait doesn't make sense because it's not clear what the other type that would implement it would be. `ColumnType` and `NullableColumnType` seems to be what you are thinking.

It's not clear that `NullableColumnType` is the right design choice here.

Even if this refactor makes sense, I would like to see other PRs come first that necessitate this refactor.

PR #226, opened 2024-10-05
chore: resolve `redundant_closure_for_method_calls` lint warnings

mehulmathur16:
# Rationale for this change

We have cargo clippy running in our CI in order to enforce code quality. In order to increase our standards, we should enable the clippy::pedantic lint group.

Resolved `redundant_closure_for_method_calls` lint warnings

# Are these changes tested?

Yes.


crates/proof-of-sql/src/base/commitment/naive_commitment.rs
```diff
@@ -119,35 +119,37 @@ impl Commitment for NaiveCommitment {
...
-                        big_int_vec.iter().map(|b| b.into()).collect()
+                        big_int_vec.iter().map(std::convert::Into::into).collect()
                     }
                     CommittableColumn::Int128(int_128_vec) => {
-                        int_128_vec.iter().map(|b| b.into()).collect()
+                        int_128_vec.iter().map(std::convert::Into::into).collect()
                     }
                     CommittableColumn::Decimal75(_, _, u64_vec) => {
-                        u64_vec.iter().map(|b| b.into()).collect()
+                        u64_vec.iter().map(std::convert::Into::into).collect()
```
iajoiner, line 140: We may as well import `core::convert::Into`. Also please bear in mind that we need to support `no_std` so unless it is absolutely necessary let's not use `std`.

crates/proof-of-sql/src/base/database/column.rs
```diff
@@ -1014,8 +1011,11 @@ mod tests {
             "Spazju u Ħin",
         ];
         let scalars = strs.iter().map(Curve25519Scalar::from).collect::<Vec<_>>();
-        let owned_col =
-            OwnedColumn::VarChar(strs.iter().map(|s| s.to_string()).collect::<Vec<String>>());
+        let owned_col = OwnedColumn::VarChar(
+            strs.iter()
+                .map(std::string::ToString::to_string)
```
iajoiner, line 1016: Similarly `to_string` probably should be imported and `std` should be avoided

iajoiner (changes requested):
Really thanks for your hard work, @mehulmathur16 !
Please address the comments.

mehulmathur16:
Thanks @iajoiner.

JayWhite2357 (approved):

iajoiner (approved):
NIT: We really should put `core::convert::Into` etc at the top. I'm fine with this for now though.

JayWhite2357:
> NIT: We really should put `core::convert::Into` etc at the top. I'm fine with this for now though.

Agreed. I'm not too opinionated on this though.

Issue #229, opened 2024-10-07
Add `FixedSizeBinary` with support up to at least 32 bytes.


JayWhite2357:
# Background and Motivation

Many data are binary blobs. In particular, the EVM address and hash types are 20 bytes and 32 bytes respectively. We should support these natively within Proof of SQL. See https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#variant.FixedSizeBinary.

# Changes Requested

Add full support for a `FixedSizeBinary` type. In order to do this there are several considerations.
* Since `Scalar` types support naturally embedding 31 bytes but not 32 bytes, there should be a divergence in how <=31 bytes are handled and 32 bytes are handled.
    * 31 bytes and under should be naturally embedded into the `Scalar` type. This means simply relying on the conversion from `[u64; 4]` to the `Scalar`.
    * 32 bytes should be handled by first hashing the data and then converting it to a `Scalar`. This is identical to how `String`s are handled. See https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/48b2f0bc8b92dbc3ef9deecfdfc04dfbc2b4460c/crates/proof-of-sql/src/base/scalar/mont_scalar_from.rs#L16.
* `ColumnType` should probably have a new variant `FixedSizeBinary(i32)` added. It is not as clear what the variant for `Column` and `OwnedColumn` should be.
    * My initial suggestion would be `FixedSizeBinary(i32, &'a [u8])` and `FixedSizeBinary(i32, Vec<u8>)` respectively, but that may not make sense. The reason for this suggestion is because something like `Vec<Vec<u8>>` has a lot of overhead, and `Vec<[u8]>` is not allowed by Rust.
    * Alternatively, it may make sense to have two variants: one for the <=31 bytes case, and one for the 32 bytes case. In this case, it may make sense to have the latter simply be `Binary` instead of `FixedSizeBinary`.
* #228 is partially motivated by this issue, and may make this issue a bit more ergonomic. However, working on them in parallel will likely lead to some branch conflict, so pay attention to that issue.
* The only expression that needs to be supported is equality (`=`) and inequality (`!=`). This is because there is no way to add, subtract, etc. between binary data. Eventually, we may wish to add bitwise operations, but those are beyond the scope of this issue.
[...]

PR #240, opened 2024-10-08
feat!: Initial Support for `FixedSizeBinary`

tareknaser:
# Rationale for this change
This PR adds support for `FixedSizeBinary` 

> This PR is not meant to close #229 but serves as initial support. We should use this as a stepping stone and, after discussions arising from the code, decide how to build on this and address the issue in a follow-up PR.
- Added support for `FixedSizeBinary` in various parts of the codebase. The proposed representation is `FixedSizeBinary(i32, Vec<u8>)`, with conversion implementations `from (i32, Vec<u8>)` and `(i32, &[u8])`.
- Updated documentation to reflect the new changes (`docs/SQLSyntaxSpecification.md`)
- Modified the implementation of `impl<T: MontConfig<4>> From<&[u8]> for MontScalar<T> {`:
	- Previously, this was only used for strings and hashed the value regardless.
	- Now, if the byte slice is empty, the result is the zero scalar. If the byte slice has length 31 or less, the bytes are directly converted to a scalar. If the byte slice has length 32, the bytes are hashed using blake3 and the result is converted to a scalar.
	- String handling was updated to keep it hashed all the time.
# Are these changes tested?
Yes, new tests were added.
However, there are some "FIXME" comments in places where I wasn't sure if the implementation is correct. These are open for discussion and review.
# Additional Notes
- I have not added literal support yet.
- I have attempted to add support in every required place in the code, but please let me know if I missed any areas!

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -1,6 +1,6 @@
 use crate::base::{
     database::{Column, ColumnType, OwnedColumn},
-    math::decimal::Precision,
+    math::{decimal::Precision, non_negative_i32::NonNegativeI32},
     ref_into::RefInto,
     scalar::Scalar,
 };
@@ -45,6 +45,9 @@ pub enum CommittableColumn<'a> {
     VarChar(Vec<[u64; 4]>),
     /// Borrowed Timestamp column with Timezone, mapped to `i64`.
     TimestampTZ(PoSQLTimeUnit, PoSQLTimeZone, &'a [i64]),
+    /// Borrowed `FixedSizeBinary` column, mapped to a slice of bytes.
+    /// - The i32 specifies the number of bytes per value.
+    FixedSizeBinary(NonNegativeI32, &'a [u8]),
 }
 
 impl CommittableColumn<'_> {
@@ -62,6 +65,7 @@ impl CommittableColumn<'_> {
             | CommittableColumn::Scalar(col)
             | CommittableColumn::VarChar(col) => col.len(),
             CommittableColumn::Boolean(col) => col.len(),
+            CommittableColumn::FixedSizeBinary(_, items) => items.len(),
         }
     }
 
@@ -94,6 +98,7 @@ impl<'a> From<&CommittableColumn<'a>> for ColumnType {
             CommittableColumn::VarChar(_) => ColumnType::VarChar,
             CommittableColumn::Boolean(_) => ColumnType::Boolean,
             CommittableColumn::TimestampTZ(tu, tz, _) => ColumnType::TimestampTZ(*tu, *tz),
+            CommittableColumn::FixedSizeBinary(size, _) => ColumnType::FixedSizeBinary(*size),
         }
     }
 }
@@ -118,6 +123,7 @@ impl<'a, S: Scalar> From<&Column<'a, S>> for CommittableColumn<'a> {
                 CommittableColumn::VarChar(as_limbs)
             }
             Column::TimestampTZ(tu, tz, times) => CommittableColumn::TimestampTZ(*tu, *tz, times),
+            Column::FixedSizeBinary(bw, bytes) => CommittableColumn::FixedSizeBinary(*bw, bytes),
         }
     }
 }
@@ -157,6 +163,9 @@ impl<'a, S: Scalar> From<&'a OwnedColumn<S>> for CommittableColumn<'a> {
             ),
             OwnedColumn::TimestampTZ(tu, tz, times) => {
                 CommittableColumn::TimestampTZ(*tu, *tz, times as &[_])
+            }
+            OwnedColumn::FixedSizeBinary(bw, bytes) => {
+                CommittableColumn::FixedSizeBinary(*bw, bytes)
             }
         }
     }
@@ -220,6 +229,7 @@ impl<'a, 'b> From<&'a CommittableColumn<'b>> for Sequence<'a> {
             | CommittableColumn::VarChar(limbs) => Sequence::from(limbs),
             CommittableColumn::Boolean(bools) => Sequence::from(*bools),
             CommittableColumn::TimestampTZ(_, _, times) => Sequence::from(*times),
+            CommittableColumn::FixedSizeBinary(_, items) => Sequence::from(*items),
         }
     }
 }
--- a/crates/proof-of-sql/src/base/database/column.rs
+++ b/crates/proof-of-sql/src/base/database/column.rs
@@ -1,6 +1,6 @@
 use super::{LiteralValue, OwnedColumn, TableRef};
 use crate::base::{
-    math::decimal::Precision,
+    math::{decimal::Precision, non_negative_i32::NonNegativeI32},
     scalar::{Scalar, ScalarExt},
     slice_ops::slice_cast_with,
 };
@@ -52,6 +52,9 @@ pub enum Column<'a, S: Scalar> {
     /// - the second element maps to a timezone
     /// - the third element maps to columns of timeunits since unix epoch
     TimestampTZ(PoSQLTimeUnit, PoSQLTimeZone, &'a [i64]),
+    /// Fixed size binary columns
+    /// - the i32 specifies the number of bytes per value
+    FixedSizeBinary(NonNegativeI32, &'a [u8]),
 }
 
 impl<'a, S: Scalar> Column<'a, S> {
@@ -72,6 +75,7 @@ impl<'a, S: Scalar> Column<'a, S> {
             Self::TimestampTZ(time_unit, timezone, _) => {
                 ColumnType::TimestampTZ(*time_unit, *timezone)
             }
+            Self::FixedSizeBinary(bw, _) => ColumnType::FixedSizeBinary(*bw),
         }
     }
     /// Returns the length of the column.
@@ -92,6 +96,7 @@ impl<'a, S: Scalar> Column<'a, S> {
             }
             Self::Int128(col) => col.len(),
             Self::Scalar(col) | Self::Decimal75(_, _, col) => col.len(),
+            Self::FixedSizeBinary(bw, col) => col.len() / bw.width_as_usize(),
         }
     }
     /// Returns `true` if the column has no elements.
@@ -177,6 +182,7 @@ impl<'a, S: Scalar> Column<'a, S> {
                 ))
             }
             OwnedColumn::TimestampTZ(tu, tz, col) => Column::TimestampTZ(*tu, *tz, col.as_slice()),
+            OwnedColumn::FixedSizeBinary(bw, col) => Column::FixedSizeBinary(*bw, col.as_slice()),
         }
     }
 
@@ -268,6 +274,15 @@ impl<'a, S: Scalar> Column<'a, S> {
         }
     }
 
+    /// Returns the column as a slice of u8 if it is a fixed size binary column. Otherwise, returns None.
+    #[must_use]
+    pub fn as_fixed_size_binary(&self) -> Option<(NonNegativeI32, &'a [u8])> {
+        match self {
+            Column::FixedSizeBinary(width, data) => Some((*width, *data)),
+            _ => None,
+        }
+    }
+
     /// Returns element at index as scalar
     ///
     /// Note that if index is out of bounds, this function will return None
@@ -282,6 +297,7 @@ impl<'a, S: Scalar> Column<'a, S> {
             Self::Int128(col) => S::from(col[index]),
             Self::Scalar(col) | Self::Decimal75(_, _, col) => col[index],
             Self::VarChar((_, scals)) => scals[index],
+            Self::FixedSizeBinary(_bw, col) => S::from(col[index]),
         })
     }
 
@@ -301,6 +317,16 @@ impl<'a, S: Scalar> Column<'a, S> {
             Self::Int128(col) => slice_cast_with(col, |i| S::from(i) * scale_factor),
             Self::Scalar(col) => slice_cast_with(col, |i| S::from(i) * scale_factor),
             Self::TimestampTZ(_, _, col) => slice_cast_with(col, |i| S::from(i) * scale_factor),
+            Self::FixedSizeBinary(bw, col) => {
+                let num_elements = col.len() / bw.width_as_usize();
+                (0..num_elements)
+                    .map(|i| {
+                        let start = i * bw.width_as_usize();
+                        let end = start + bw.width_as_usize();
+                        S::from(&col[start..end]) * scale_factor
+                    })
+                    .collect()
+            }
         }
     }
 }
@@ -345,6 +371,9 @@ pub enum ColumnType {
     /// Mapped to `S`
     #[serde(alias = "SCALAR", alias = "scalar")]
     Scalar,
+    /// Mapped to fixed size binary
+    #[serde(alias = "FIXEDSIZEBINARY", alias = "fixedsizebinary")]
+    FixedSizeBinary(NonNegativeI32),
 }
 
 impl ColumnType {
@@ -459,7 +488,7 @@ impl ColumnType {
             // Scalars are not in database & are only used for typeless comparisons for testing so we return 0
             // so that they do not cause errors when used in comparisons.
             Self::Scalar => Some(0_u8),
-            Self::Boolean | Self::VarChar => None,
+            Self::Boolean | Self::VarChar | Self::FixedSizeBinary(_) => None,
         }
     }
     /// Returns scale of a [`ColumnType`] if it is convertible to a decimal wrapped in `Some()`. Otherwise return None.
@@ -474,7 +503,7 @@ impl ColumnType {
             | Self::BigInt
             | Self::Int128
             | Self::Scalar => Some(0),
-            Self::Boolean | Self::VarChar => None,
+            Self::Boolean | Self::VarChar | Self::FixedSizeBinary(_) => None,
             Self::TimestampTZ(tu, _) => match tu {
                 PoSQLTimeUnit::Second => Some(0),
                 PoSQLTimeUnit::Millisecond => Some(3),
@@ -496,6 +525,7 @@ impl ColumnType {
             Self::BigInt | Self::TimestampTZ(_, _) => size_of::<i64>(),
             Self::Int128 => size_of::<i128>(),
             Self::Scalar | Self::Decimal75(_, _) | Self::VarChar => size_of::<[u64; 4]>(),
+            Self::FixedSizeBinary(bw) => bw.width_as_usize(),
         }
     }
 
@@ -516,9 +546,12 @@ impl ColumnType {
             | Self::BigInt
             | Self::Int128
             | Self::TimestampTZ(_, _) => true,
-            Self::Decimal75(_, _) | Self::Scalar | Self::VarChar | Self::Boolean | Self::Uint8 => {
-                false
-            }
+            Self::Decimal75(_, _)
+            | Self::Scalar
+            | Self::VarChar
+            | Self::Boolean
+            | Self::Uint8
+            | Self::FixedSizeBinary(_) => false,
         }
     }
 }
@@ -546,6 +579,7 @@ impl Display for ColumnType {
             ColumnType::TimestampTZ(timeunit, timezone) => {
                 write!(f, "TIMESTAMP(TIMEUNIT: {timeunit}, TIMEZONE: {timezone})")
             }
+            ColumnType::FixedSizeBinary(bw) => write!(f, "FIXEDSIZEBINARY({bw})"),
         }
     }
 }
--- a/crates/proof-of-sql/src/base/database/column_repetition_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_repetition_operation.rs
@@ -6,6 +6,7 @@ use crate::base::scalar::Scalar;
 use alloc::vec::Vec;
 use bumpalo::Bump;
 use core::iter::Iterator;
+use std::iter;
 
 pub trait RepetitionOp {
     fn op<T: Clone>(column: &[T], n: usize) -> impl Iterator<Item = T>;
@@ -138,15 +139,10 @@ impl RepetitionOp for ColumnRepeatOp {
     }
 
     fn op_fixed_size_binary(col_bytes: &[u8], width: usize, n: usize) -> Vec<u8> {
-        let num_rows = col_bytes.len() / width;
-        let mut out = Vec::with_capacity(col_bytes.len() * n);
-
-        for _ in 0..n {
-            for row_idx in 0..num_rows {
-                let start = row_idx * width;
-                let end = start + width;
-                out.extend_from_slice(&col_bytes[start..end]);
-            }
+        let rows: Vec<_> = col_bytes.chunks_exact(width).collect();
+        let mut out = Vec::with_capacity(rows.len() * width * n);
+        for row in rows.iter().cycle().take(rows.len() * n) {
+            out.extend_from_slice(row);
         }
         out
     }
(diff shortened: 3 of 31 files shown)


JayWhite2357:
>This PR is not meant to close #229 but serves as initial support. We should use this as a stepping stone and, after discussions arising from the code, decide how to build on this and address the issue in a follow-up PR.

Thanks for the PR! The CI is not passing (which I think you expected).
If there is anything you want our team to give you feedback on, it would be good if you can highlight that with comments since we won't have time for thorough feedback until a PR is ready to be merged/passing CI.

tareknaser:
Thanks, @JayWhite2357 
Could we rerun the CI?

By the way, I’d love to connect on Discord. I joined the server but not sure how to reach you there.

tareknaser:
Thanks, @JayWhite2357 

Could we rerun the CI?
`cargo test --all-features` now passes locally 

> If there is anything you want our team to give you feedback on,

Could you start by taking a look at DataType, Column, OwnedColumn, and the others? I made some decisions where I wasn’t sure of the best approach, so your feedback would be really helpful. I also added some ‘FIXME’ comments.


By the way, I’d love to connect on Discord. I joined the server but not sure how to reach you there.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_cpu.rs
```diff
@@ -61,6 +61,11 @@ fn compute_dory_commitment(
         CommittableColumn::RangeCheckWord(column) => {
             compute_dory_commitment_impl(column, offset, setup)
         }
+        CommittableColumn::FixedSizeBinary(_byte_width, column) => {
+            // FIXME: Is this interpretation correct? or should we chunk the bytes
+            // into `byte_width`-sized chunks and convert each chunk into a scalar?
+            compute_dory_commitment_impl(column, offset, setup)
```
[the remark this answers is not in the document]
JayWhite2357: I believe something like this may work:
```suggestion
            // FIXME: Is this interpretation correct? or should we chunk the bytes
            // into `byte_width`-sized chunks and convert each chunk into a scalar?
            compute_dory_commitment_impl(&column.iter().chunks_exact(byte_width).collect_vec(), offset, setup)
```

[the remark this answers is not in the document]
JayWhite2357: Be sure to add some tests here ensuring that this is behaving correctly. Even though CI is currently passing, this code is currently incorrect.

JayWhite2357 (commented):
Thanks for the PR! This is heading in the right direction.
I don't have an amazing solution to the concerns here, but hopefully my feedback can help with direction.

crates/proof-of-sql/src/base/commitment/committable_column.rs
```diff
@@ -221,6 +236,8 @@ impl<'a, 'b> From<&'a CommittableColumn<'b>> for Sequence<'a> {
             CommittableColumn::Boolean(bools) => Sequence::from(*bools),
             CommittableColumn::TimestampTZ(_, _, times) => Sequence::from(*times),
             CommittableColumn::RangeCheckWord(words) => Sequence::from(*words),
+            // FIXME: Is this the correct way to convert a FixedSizeBinary column to a Sequence?
+            CommittableColumn::FixedSizeBinary(_, bytes) => Sequence::from(*bytes),
```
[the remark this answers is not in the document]
tareknaser: Right. I’ve opened a PR https://github.com/spaceandtimelabs/blitzar-rs/pull/44 to address that. Let me know what you think

crates/proof-of-sql/src/base/database/column.rs
```diff
@@ -266,6 +303,9 @@ pub enum ColumnType {
     /// Mapped to [`Curve25519Scalar`](crate::base::scalar::Curve25519Scalar)
     #[serde(alias = "SCALAR", alias = "scalar")]
     Scalar,
+    /// Mapped to fixed size binary
+    #[serde(alias = "FIXEDSIZEBINARY", alias = "fixedsizebinary")]
+    FixedSizeBinary(i32),
```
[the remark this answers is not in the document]
tareknaser: At first, I wanted to use `FixedSizeBinary(usize)` since it made more sense to me, but [Arrow actually uses `i32` for that](https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#variant.FixedSizeBinary), so I decided to stick with the convention.

> having some way of ensuring that the number of bytes must be in the range `[0, 32]`


where do you think such a check should go?

crates/proof-of-sql/src/base/database/owned_column.rs
```diff
@@ -79,6 +87,10 @@ impl<S: Scalar> OwnedColumn<S> {
             OwnedColumn::TimestampTZ(tu, tz, col) => {
                 OwnedColumn::TimestampTZ(*tu, *tz, permutation.try_apply(col)?)
             }
+            OwnedColumn::FixedSizeBinary(byte_width, col) => {
+                // FIXME: Would that work for all FixedSizeBinary cases?
+                OwnedColumn::FixedSizeBinary(*byte_width, permutation.try_apply(col)?)
+            }
```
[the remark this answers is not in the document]
tareknaser: Should be fixed in 688c6ec

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_cpu.rs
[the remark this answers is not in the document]
tareknaser: Alright, I’ve updated that based on your suggestion and also added a test `we_can_compute_a_dory_commitment_with_fixed_size_binary_values` that is currently failing with the following error:
```sh
panicked at crates/proof-of-sql/src/proof_primitive/dory/pack_scalars.rs:179:14:

source slice length (1) does not match destination slice length (4)
```

tracing it back, I found that the `pack_bit` function is the one panicking. It seems to be related to the implementation of `OffsetToBytes`.
```rust
column.iter().enumerate().for_each(|(i, value)| {
	let index = i + offset;
	let row_offset = (index % num_columns) * bit_table_sum_in_bytes;
	let col_offset = current_byte_size * (index / num_columns);
	let offset_index = row_offset + col_offset + byte_offset;

	packed_scalars[offset_index..offset_index + current_byte_size]
		.copy_from_slice(&value.offset_to_bytes()[..]);
});
```
The issue is that it’s currently iterating over values and calling `offset_to_bytes`, treating each element in the column as its own value. 

This is not ideal for `FixedSizeBinary` because multiple consecutive values form a single value (`byte_size`). I’m not quite sure how to handle this properly. Could you point me in the right direction?

crates/proof-of-sql/src/base/commitment/committable_column.rs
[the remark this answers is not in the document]
tareknaser: Updated to use `Sequence::from_raw_parts_with_size` for `FixedSizeBinary`

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_cpu.rs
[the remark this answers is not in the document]
tareknaser: Any updates on this? I’ve rebased on top of main @JayWhite2357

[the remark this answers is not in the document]
JayWhite2357: This is still marked as draft. Is it ready for review?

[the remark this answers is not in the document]
tareknaser: Marked it as ready for review. The test `we_can_compute_a_dory_commitment_with_fixed_size_binary_values` is still failing the CI, though. Let me know what you think

tareknaser:
@JayWhite2357 
Just wanted to check in. Any updates on this PR?

tareknaser:
Any updates on this?

tareknaser:
So I'm using `i32` to represent the size in `FixedSizeBinary(i32)`, as Jay suggested in the original issue description (and as implemented in `arrow`) but in many parts of the code, I need to use this value in arithmetic operations with `usize` values. 

For that, I’ve added `#[allow(clippy::cast_sign_loss)]` in the relevant places to suppress the Clippy cast warnings for now.

crates/proof-of-sql/src/base/database/column_repetition_operation.rs
```diff
@@ -120,19 +136,48 @@ impl RepetitionOp for ColumnRepeatOp {
...
+
+        for _ in 0..n {
+            for row_idx in 0..num_rows {
+                let start = row_idx * width;
+                let end = start + width;
+                out.extend_from_slice(&col_bytes[start..end]);
+            }
+        }
+        out
+    }
```
Dustin-Ray, line 152: You can shrink this down a bit with ```cycle```:
```suggestion
fn op_fixed_size_binary(col_bytes: &[u8], width: usize, n: usize) -> Vec<u8> {
    let rows: Vec<_> = col_bytes.chunks_exact(width).collect();
    let mut out = Vec::with_capacity(rows.len() * width * n);
    for row in rows.iter().cycle().take(rows.len() * n) {
        out.extend_from_slice(row);
    }
    out
}
```

crates/proof-of-sql/src/base/database/column_repetition_operation.rs
```diff
@@ -120,19 +136,48 @@ impl RepetitionOp for ColumnRepeatOp {
...
+        for row_idx in 0..num_rows {
+            let start = row_idx * width;
+            let end = start + width;
+            let chunk = &col_bytes[start..end];
+            for _ in 0..n {
+                out.extend_from_slice(chunk);
+            }
+        }
+        out
+    }
```
Dustin-Ray, line 174: we can probably make this a little more idiomatic as well
```suggestion
fn op_fixed_size_binary(col_bytes: &[u8], width: usize, n: usize) -> Vec<u8> {
    let mut out = Vec::with_capacity(col_bytes.len() * n);
    out.extend(
        col_bytes
            .chunks_exact(width)
            .flat_map(|row| iter::repeat(row).take(n))
            .flatten()
            .cloned()
    );
    out
}
```

crates/proof-of-sql/src/base/database/slice_operation.rs
```diff
@@ -242,6 +242,22 @@ pub(super) fn repeat_elementwise<S: Clone>(slice: &[S], n: usize) -> impl Iterat
...
+        let start = row_idx * width;
+        let end = start + width;
+        let chunk = &col_bytes[start..end];
+        for _ in 0..n {
+            out.extend_from_slice(chunk);
+        }
+    }
+    out
+}
+
```
Dustin-Ray, line 260: consider a fully functional approach here:
```suggestion
fn repeat_elementwise_fixed_size_binary_no_prealloc(
    col_bytes: &[u8],
    width: usize,
    n: usize,
) -> Vec<u8> {
    col_bytes
        .chunks_exact(width)
        .flat_map(|row| std::iter::repeat(row).take(n)) 
        .flatten()               // turns iterator of &[u8] into iterator of &u8
        .copied()               // copies each &u8 into a real u8
        .collect()
}
```

crates/proof-of-sql/src/base/math/permutation.rs
```diff
@@ -84,6 +84,40 @@ impl Permutation {
...
+
+        let mut result = Vec::with_capacity(slice.len());
+        for &i in &self.permutation {
+            let start = i * chunk_size;
+            let end = start + chunk_size;
+            result.extend_from_slice(&slice[start..end]);
+        }
+
+        Ok(result)
+    }
```
Dustin-Ray, line 120: This _may or not_ work but if it does it helps reduce the preallocation and yields a slightly more functional result:
```suggestion
pub fn try_chunked_apply<T>(
    &self,
    slice: &[T],
    chunk_size: usize,
) -> Result<Vec<T>, PermutationError>
where
    T: Clone,
{
    if slice.len() % chunk_size != 0 {
        return Err(PermutationError::PermutationSizeMismatch {
            permutation_size: self.size(),
            slice_length: slice.len(),
        });
    }
    let num_chunks = slice.len() 
    if self.size() != num_chunks {
        return Err(PermutationError::PermutationSizeMismatch {
            permutation_size: self.size(),
            slice_length: num_chunks,
        });
    }
    let result = self
        .permutation
        .iter()                           
        .flat_map(|&i| {
            slice[i * chunk_size .. (i + 1) * chunk_size]
                .iter()                    
        })
        .cloned()                        
        .collect();                      

    Ok(result)
}
```

tareknaser: Thanks for the review. This works fine. just updated it

crates/proof-of-sql/src/base/database/slice_operation.rs
tareknaser: committed this code. except using `iter` from `core` instead of `std`

Dustin-Ray (approved):
LGTM - thanks a lot for this!

crates/proof-of-sql/src/base/database/column_repetition_operation.rs
```diff
@@ -120,19 +136,40 @@ impl RepetitionOp for ColumnRepeatOp {
     fn op<T: Clone>(column: &[T], n: usize) -> impl Iterator<Item = T> {
         repeat_slice(column, n)
     }
+
+    fn op_fixed_size_binary(col_bytes: &[u8], width: usize, n: usize) -> Vec<u8> {
+        let rows: Vec<_> = col_bytes.chunks_exact(width).collect();
+        let mut out = Vec::with_capacity(rows.len() * width * n);
```
JayWhite2357, line 142: Please don't create temporary `Vec`s. This is inefficient and unneeded.
Please avoid mutability.

crates/proof-of-sql/src/base/database/column_repetition_operation.rs
```diff
@@ -120,19 +136,40 @@ impl RepetitionOp for ColumnRepeatOp {
...
+        for row in rows.iter().cycle().take(rows.len() * n) {
+            out.extend_from_slice(row);
+        }
+        out
```
JayWhite2357, line 146:
```suggestion
        repeat_slice(col_bytes, n)
```

crates/proof-of-sql/src/base/database/column_repetition_operation.rs
```diff
@@ -120,19 +136,40 @@ impl RepetitionOp for ColumnRepeatOp {
...
+    fn op_fixed_size_binary(col_bytes: &[u8], width: usize, n: usize) -> Vec<u8> {
+        let mut out = Vec::with_capacity(col_bytes.len() * n);
+        out.extend(
+            col_bytes
+                .chunks_exact(width)
+                .flat_map(|row| iter::repeat(row).take(n))
+                .flatten()
+                .copied(),
+        );
+        out
```
JayWhite2357, line 165: Please avoid mutability
```suggestion
            col_bytes
                .chunks_exact(width)
                .flat_map(|row| iter::repeat(row).take(n))
                .flatten()
                .copied()
                .collect()
```

crates/proof-of-sql/src/base/database/column_repetition_operation.rs
```diff
@@ -3,14 +3,18 @@ use super::{
...
 use bumpalo::Bump;
-use core::iter::Iterator;
+use core::{iter, iter::Iterator};
 
-#[allow(dead_code)]
 pub trait RepetitionOp {
     fn op<T: Clone>(column: &[T], n: usize) -> impl Iterator<Item = T>;
 
+    // Special case for fixed-size binary columns.
+    fn op_fixed_size_binary(col_bytes: &[u8], width: usize, n: usize) -> Vec<u8>;
```
JayWhite2357, line 14: Something _like_ this should work without needing to have a special case:
```suggestion
    fn op<T: Clone>(column: impl IntoIterator<T>, n: usize) -> impl Iterator<Item = T>;
```

crates/proof-of-sql/src/base/database/filter_util.rs
```diff
@@ -78,5 +78,21 @@ pub fn filter_column_by_index<'a, S: Scalar>(
             *tz,
             alloc.alloc_slice_fill_iter(indexes.iter().map(|&i| col[i])),
         ),
+        Column::FixedSizeBinary(byte_width, col) => {
+            let bw = byte_width.width_as_usize();
+            let required_len = indexes.len() * bw;
+
+            let allocated_bytes = alloc.alloc_slice_fill_default::<u8>(required_len);
+
+            let mut out_pos = 0;
```
JayWhite2357, line 87: Avoid `mut`.
Use something like `allocated_bytes.chunks_mut().zip(indexes)` instead.

crates/proof-of-sql/src/base/database/owned_column.rs
```diff
@@ -53,6 +57,14 @@ impl<S: Scalar> OwnedColumn<S> {
         match self {
             OwnedColumn::Boolean(col) => inner_product_ref_cast(col, vec),
             OwnedColumn::Uint8(col) => inner_product_ref_cast(col, vec),
+            OwnedColumn::FixedSizeBinary(width, col_bytes) => {
+                let bw = width.width_as_usize();
+                let chunked_vals: Vec<S> = col_bytes
+                    .chunks_exact(bw)
+                    .map(|chunk| S::from(chunk))
+                    .collect();
```
JayWhite2357, line 65: Please don't create temporary intermediate `Vec`s. Instead use iterators.

crates/proof-of-sql/src/base/database/slice_operation.rs
```diff
@@ -242,6 +242,19 @@ pub(super) fn repeat_elementwise<S: Clone>(slice: &[S], n: usize) -> impl Iterat
...
+    n: usize,
+) -> Vec<u8> {
+    col_bytes
+        .chunks_exact(width)
+        .flat_map(|row| core::iter::repeat(row).take(n))
+        .flatten() // turns iterator of &[u8] into iterator of &u8
+        .copied() // copies each &u8 into a real u8
+        .collect()
+}
+
```
JayWhite2357, line 257: This appears to be unused.

crates/proof-of-sql/src/base/database/union_util.rs
```diff
@@ -174,6 +174,27 @@ pub fn column_union<'a, S: Scalar>(
...
+            let flattened: Vec<u8> = columns
+                .iter()
+                .flat_map(|col| {
+                    let (col_width, col_data) = col
+                        .as_fixed_size_binary()
+                        .expect("column_type check above should guarantee same type");
+                    assert_eq!(col_width, width, "Inconsistent FixedSizeBinary width");
+                    col_data.iter().copied()
+                })
+                .collect();
```
JayWhite2357, line 189: Please don't create temporary `Vec`s. See the other match arms here for examples of how to avoid this.

crates/proof-of-sql/src/base/math/non_negative_i32.rs
```diff
@@ -0,0 +1,48 @@
...
+        } else {
+            Ok(Self(x))
+        }
+    }
+
+    /// Returns the wrapped value.
+    #[must_use]
+    pub fn width(&self) -> i32 {
+        self.0
+    }
```
JayWhite2357, line 31: I would prefer to see `impl From<NonNegativeI32> for i32`.

crates/proof-of-sql/src/base/math/non_negative_i32.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Returns the wrapped value.
+    #[allow(
+        clippy::cast_sign_loss,
+        reason = "i32 is guaranteed to be non-negative by constructor"
+    )]
+    #[must_use]
+    pub fn width_as_usize(&self) -> usize {
+        self.0 as usize
+    }
```
JayWhite2357, line 41: I would prefer to see `impl From<NonNegativeI32> for usize`.

crates/proof-of-sql/src/base/math/non_negative_i32.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl NonNegativeI32 {
+    /// Returns an error if `x` is negative. Otherwise returns the wrapped value.
+    pub fn new(x: i32) -> Result<Self, WidthError> {
+        if x < 0 {
+            Err(WidthError::NegativeWidth(x))
+        } else {
+            Ok(Self(x))
+        }
+    }
```
JayWhite2357, line 25: This should be named `try_new`.
We should also have `impl TryFrom<i32> for NonNegativeI32`.
Or perhaps we should remove the `try_new` altogether and only have the `TryFrom`.

crates/proof-of-sql/src/base/database/owned_column.rs
```diff
@@ -101,6 +114,10 @@ impl<S: Scalar> OwnedColumn<S> {
             OwnedColumn::TimestampTZ(tu, tz, col) => {
                 OwnedColumn::TimestampTZ(*tu, *tz, permutation.try_apply(col)?)
             }
+            OwnedColumn::FixedSizeBinary(bw, col) => OwnedColumn::FixedSizeBinary(
+                *bw,
+                permutation.try_chunked_apply(col, bw.width_as_usize())?,
+            ),
```
JayWhite2357, line 120: Please add tests for this.

crates/proof-of-sql/src/base/scalar/scalar.rs
```diff
@@ -32,7 +32,8 @@ pub trait Scalar:
     + for<'a> core::convert::From<&'a i64> // Required for `Column` to implement `MultilinearExtension`
     + for<'a> core::convert::From<&'a i128> // Required for `Column` to implement `MultilinearExtension`
     + for<'a> core::convert::From<&'a u8> // Required for `Column` to implement `MultilinearExtension`
-    + for<'a> core::convert::From<&'a u64> // Required for `Column` to implement `MultilinearExtension`
+    + for<'a> core::convert::From<&'a u64> // Required for `Column` to implement `MultilinearExtension`\
+    + for<'a> core::convert::From<&'a [u8]>
```
JayWhite2357, line 36: I want to avoid adding a `From` conversion where the conversion is potentially non-canonical.
Instead, in this case, add a new method to `ScalarExt`.

crates/proof-of-sql/src/proof_primitive/dory/dory_compute_commitments_test.rs
```diff
@@ -9,6 +12,65 @@ use ark_std::test_rng;
 use num_traits::Zero;
 use proof_of_sql_parser::posql_time::{PoSQLTimeUnit, PoSQLTimeZone};
 
+#[test]
+fn we_can_compute_a_dory_commitment_with_fixed_size_binary_values() {
```
JayWhite2357, line 16: Please add tests that cover different offsets as well.

crates/proof-of-sql/src/proof_primitive/dory/pack_scalars.rs
```diff
@@ -165,6 +165,35 @@ fn pack_bit<const LEN: usize, T: OffsetToBytes<LEN>>(
     });
 }
 
+fn pack_fixed_size_binary(
```
JayWhite2357, line 168: This code is mostly just duplicating `pack_bit`. Instead, I believe it should be doable to make the `column` parameter in `pack_bit` be an `impl IntoIterator<T>` and `impl OffsetToBytes for &[u8]`.

crates/proof-of-sql/src/sql/proof/provable_query_result.rs
```diff
@@ -215,6 +218,12 @@ impl ProvableQueryResult {
                         offset += num_read;
                         Ok((field.name(), OwnedColumn::TimestampTZ(tu, tz, col)))
                     }
+                    ColumnType::FixedSizeBinary(byte_width) => {
+                        let bw = byte_width.width_as_usize();
+                        let (col, num_read) = decode_multiple_elements(&self.data[offset..], n)?;
+                        offset += num_read * bw;
+                        Ok((field.name(), OwnedColumn::FixedSizeBinary(byte_width, col)))
+                    }
```
JayWhite2357, line 226: Please add tests here. It is non-obvious to me that this is correct.

crates/proof-of-sql/src/base/math/non_negative_i32.rs
```diff
@@ -0,0 +1,48 @@
+use alloc::fmt::Display;
+use serde::{Deserialize, Serialize};
+
+/// Type-safe non-negative integer, exists for the sole purpose
+/// of preventing negative values from being used as fixed-size
+/// binary slice widths, due to arrow's unfortunate use of i32 as width.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct NonNegativeI32(i32);
```
JayWhite2357, line 8: We should also bound the size of FixedBinary above by 31. We can't embed 32 bytes canonically inside a scalar.

JayWhite2357 (changes requested):

PR #245, opened 2024-10-09
fix: dynamic Dory commitment computation should handle empty columns

jacobtrombetta:
# Rationale for this change
The dynamic Dory commitment computation does not currently handle empty columns as expected. This was uncovered after a test formatting error was discovered. Both the empty column case and formatting error are addressed.

- The empty column case is handled in the `dynamic_dory_commitment_helper_cpu` module. Tests for this function are in - https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/242
- Tests for empty Dory commitments are corrected.

# Are these changes tested?
Yes


JayWhite2357 (approved):
Thanks!

PR #252, opened 2024-10-10
refactor!: remove `Indexes`

iajoiner:

# Rationale for this change
`Indexes` is not useful since we only use dense indexes from 0 to output table length anyway. Now it is time to remove it.

- remove `Indexes`

# Are these changes tested?

Yes


crates/proof-of-sql/src/sql/proof/provable_result_column.rs
```diff
@@ -7,64 +6,60 @@ use crate::{
...
 
 impl<'a, T: ProvableResultElement<'a>> ProvableResultColumn for &[T] {
-    fn num_bytes(&self, selection: &Indexes) -> usize {
-        let mut res = 0;
-        for i in selection.iter() {
-            res += self[i as usize].required_bytes();
-        }
-        res
+    fn num_bytes(&self, length: u64) -> usize {
+        (0..length as usize).map(|i| self[i].required_bytes()).sum()
```
JayWhite2357, line 17: Comment: I have a feeling that `length` may not even be needed. We could test this easily by adding:
```suggestion
    fn num_bytes(&self, length: u64) -> usize {
        assert_eq!(self.len(), length)
        self.iter().map(|e| e.required_bytes()).sum()
```

crates/proof-of-sql/src/sql/proof/provable_query_result.rs
```diff
@@ -51,29 +43,29 @@ impl ProvableQueryResult {
...
-            indexes,
+            table_length,
             data,
         }
     }
 
     /// Form intermediate query result from index rows and result columns
     #[must_use]
-    pub fn new<'a, S: Scalar>(indexes: &'a Indexes, columns: &'a [Column<'a, S>]) -> Self {
+    pub fn new<'a, S: Scalar>(table_length: u64, columns: &'a [Column<'a, S>]) -> Self {
```
JayWhite2357, line 56: I'm suspicious here:
```suggestion
    pub fn new<'a, S: Scalar>(table_length: u64, columns: &'a [Column<'a, S>]) -> Self {
    assert!(columns.iter().all(|column| table_length <= column.len()))
```

crates/proof-of-sql/src/sql/proof/query_proof.rs
```diff
@@ -56,7 +56,7 @@ impl<CP: CommitmentEvaluationProof> QueryProof<CP> {
         let mut result_builder = ResultBuilder::new(table_length);
         let result_cols = expr.result_evaluate(&mut result_builder, &alloc, accessor);
         let provable_result =
-            ProvableQueryResult::new(&result_builder.result_index_vector, &result_cols);
+            ProvableQueryResult::new(result_builder.table_length() as u64, &result_cols);
```
JayWhite2357, line 59: I feel like this is wrong. The result doesn't have the same length as the original data. Why isn't it panicking?

crates/proof-of-sql/src/sql/proof/provable_query_result.rs
```diff
@@ -51,29 +43,36 @@ impl ProvableQueryResult {
...
+    /// # Panics
+    ///
+    /// Will panic if `table_length` is somehow larger than the length of some column
+    /// which should never happen.
     #[must_use]
-    pub fn new<'a, S: Scalar>(indexes: &'a Indexes, columns: &'a [Column<'a, S>]) -> Self {
+    pub fn new<'a, S: Scalar>(table_length: u64, columns: &'a [Column<'a, S>]) -> Self {
+        assert!(columns
+            .iter()
+            .all(|column| table_length <= column.len() as u64));
```
JayWhite2357, line 63: Should this be `==` even?

crates/proof-of-sql/src/sql/proof/query_proof.rs
iajoiner: @JayWhite2357 It is not panicking because currently `result_evaluate` (line 57) will set `table_length` to be the output table length.

After the next PR we will differentiate between input and output table lengths better.

crates/proof-of-sql/src/sql/proof/provable_query_result.rs
iajoiner: Yes.

crates/proof-of-sql/src/sql/proof/query_proof.rs
iajoiner: Maybe we should make it very clear in `provable_query_results.rs` etc that we always use `output_length`, that is, actually rename all the variables. Meanwhile if we use input length(s) we should label them as such.

iajoiner: BTW `result_builder.result_index_vector` serves the exact same purpose since all that it can be in `ProofPlan`s that do exist is `Indexes::Dense((0..output_length))`

JayWhite2357 (dismissed):
Thanks, Ian!

JayWhite2357 (approved):

PR #253, opened 2024-10-10
feat: updated the examples to use the `DynamicDoryEvaluationProof`

### LINKED ISSUE #244: Update examples to use the dynamic Dory commitment scheme.
# Background and Motivation

Our example code leverages the `InnerProductProof` commitment scheme. While this is still valid, it is more difficult as an entry point to the code because it requires Linux (and GPU by default).

# Changes Required

- [x] Update the examples (in `crates/proof-of-sql/examples`) to use the `DynamicDoryEvaluationProof` instead of the `InnerProductProof`.
- [ ] Ensure these examples run correctly with the `blitzar` feature disabled. (i.e. update the CI)
- [ ] Update the READMEs to add examples of how to run this with and without the `blitzar` feature enabled.

Abiji-2020:

# Rationale for this change

To improve add the functionality of everyone to use non-GPU systems from the entry point 

 Closes #244 .
/claim #244 

 

Modified the examples completely to depend and run on `DynamicDoryEvaluationProof` . 

# Are these changes tested?

Yes, the existing test are enough with the same values

### DIFF
--- a/crates/proof-of-sql/Cargo.toml
+++ b/crates/proof-of-sql/Cargo.toml
@@ -83,7 +83,7 @@ workspace = true
 
 [[example]]
 name = "hello_world"
-required-features = [ "blitzar", "test" ]
+required-features = [ "test" ]
 
 [[example]]
 name = "posql_db"
--- a/crates/proof-of-sql/examples/posql_db/main.rs
+++ b/crates/proof-of-sql/examples/posql_db/main.rs
@@ -5,20 +5,23 @@ mod commit_accessor;
 mod csv_accessor;
 /// TODO: add docs
 mod record_batch_accessor;
+use ark_std::rand::thread_rng;
 use arrow::{
     datatypes::{DataType, Field, Schema},
     record_batch::RecordBatch,
 };
-use blitzar::proof::InnerProductProof;
 use clap::{arg, Parser, Subcommand, ValueEnum};
 use commit_accessor::CommitAccessor;
 use csv_accessor::{read_record_batch_from_csv, CsvDataAccessor};
-use curve25519_dalek::RistrettoPoint;
 use itertools::Itertools;
 use proof_of_sql::{
     base::{
         commitment::TableCommitment,
         database::{SchemaAccessor, TableRef},
+    },
+    proof_primitive::dory::{
+        DynamicDoryCommitment, DynamicDoryEvaluationProof, ProverSetup, PublicParameters,
+        VerifierSetup,
     },
     sql::{parse::QueryExpr, proof::VerifiableQueryResult},
 };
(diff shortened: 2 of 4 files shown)


crates/proof-of-sql/examples/posql_db/main.rs
```diff
@@ -149,14 +152,19 @@ fn main() {
     println!("Warming up GPU...");
     blitzar::compute::init_backend();
     println!("Done.");
+
+    let mut rng = thread_rng();
+    let public_parameters = PublicParameters::rand(5, &mut rng);
+    let prover_setup = ProverSetup::from(&public_parameters);
+    let verifier_setup = VerifierSetup::from(&public_parameters);
```
JayWhite2357, line 159: This is the reason that the verification is failing. Because you are using `thread_rng`, the public parameters are different every time this is run. As a result the prover setup and the verifier setup have a mismatch.

To fix this, we just need `rng` to be a deterministic (seeded) random number generator. That should fix this.

crates/proof-of-sql/Cargo.toml
```diff
@@ -83,7 +83,7 @@ workspace = true
 
 [[example]]
 name = "hello_world"
-required-features = [ "blitzar", "test" ]
+required-features = [ "test" ]
```
JayWhite2357, line 86: I think it should be possible to remove even the "test" required feature. I could be wrong, so if not, I'd be curious to know why. (In which case, I will probably want to fix that.)

JayWhite2357 (changes requested):
Looks great. I think you are almost there.
Thanks for the PR!

crates/proof-of-sql/Cargo.toml
Abiji-2020: When the test feature is removed, Getting errors on the usage of the test_rand() function, TestAccessor, etc.. which uses the test feature 

image

JayWhite2357: Makes sense. That's fine, then.

crates/proof-of-sql/examples/posql_db/main.rs
Abiji-2020: Updated the code to use the `rng`. Now the errors are resolved and we can verify the proofs

JayWhite2357 (approved):
This looks great! Thanks for the PR.

With regard to #253, this does not complete the issue. Still outstanding are:
- [ ] Ensure these examples run correctly with the `blitzar` feature disabled. (i.e. update the CI)
- [ ] Update the READMEs to add examples of how to run this with and without the `blitzar` feature enabled.

Abiji-2020:
@JayWhite2357 could you explain more about the ci part.

PR #266, opened 2024-10-12
refactor!: isolate arrow code

varshith257:
Please be sure to look over the pull request guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md#submit-pr.

# Please go through the following checklist
- [ ] The PR title and commit messages adhere to guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md. In particular `!` is used if and only if at least one breaking change has been introduced.
- [ ] I have run the ci check script with `source scripts/run_ci_checks.sh`.

# Rationale for this change
Currently, arrow is one of the elements that must be disabled to enable no_std support. There are many instances of #[cfg(feature = "arrow")] throughout the codebase, which leaves the code cluttered. To make code readable and manageable  , arrow module is introduced.

Closes #237 
/claim #237 

# What changes are included in this PR?

- Moved all  [cfg(feature = "arrow")] related code to separate modules and cracked cyclic dependencies along the way without any massive disruption. 
- Bumped Arrow from 51 -> 53

# Are these changes tested?

Yes

PR #267, opened 2024-10-12
refactor: Isolate arrow code #237

### LINKED ISSUE #237: Isolate `arrow` code
# Background and Motivation

Currently, `arrow` is one of the elements that must be disabled to enable `no_std` support. Additionally, not every use-case requires arrow. So, we have gated it behind a features flag (`feature = "arrow"`). However, there are many instances of `#[cfg(feature = "arrow")]` throughout the codebase, which leave the code cluttered.

# Changes Required

- [ ] Move all arrow code to a single (or a small number) of modules.
    - My suggestion would be creating a new `proof_of_sql::base::arrow` module and moving all code there. However, this may not be possible, so this is not a hard requirement.
    - The resulting code _**must**_ not introduce cyclic dependencies and should resolve any existing ones with arrow related code. This is one of the primary reasons for this cleanup: in order to more easily identify and resolve these.
- [ ] Update any legacy documentation referencing arrow to ensure that it is up to date.
- [ ] Bump the `arrow` crate to version 53.

melsonic:

# Rationale for this change

Currently, `arrow` is one of the elements that must be disabled to enable `no_std` support. Additionally, not every use-case requires `arrow`. So, it is gated behind a features flag (feature = "arrow"). However, there are many instances of #[cfg(feature = "arrow")] throughout the codebase, which leaves the code cluttered.
This PR moves the parts of code that use Arrow to a separate `arrow` module.

This PR isolates the parts of code that uses the `arrow` crate into a separate module. Along with that `arrow` version is also upgraded to 53.

/fixes #237 
/claim #237


JayWhite2357 (commented):
My concern here is that you have pulled too much out of the core code into the arrow modules.
For example, `TableCommitment` is not an arrow specific component, so it should not be moved.

Additionally, I would like to see the entire arrow module behind the arrow feature. Part of the motivation here is that
> [...] there are many instances of `#[cfg(feature = "arrow")]` throughout the codebase, which leave the code cluttered.
So the goal should be to minimize the number of these `cfg` statements.

JayWhite2357:
Closed in favor of #266

PR #270, opened 2024-10-13
chore: major resolved `clippy::cast_possible_wrap` & `clippy::cast_possible_truncation` lint warnings

mehulmathur16:
# Rationale for this change
We have cargo clippy running in our CI in order to enforce code quality. In order to increase our standards, we should enable the clippy::pedantic lint group.

Resolved the following lint warnings

`too_many_lines`
`cast_possible_wrap`
`cast_sign_loss`
`cast_possible_truncation`

# Are these changes tested?
Yes.


crates/proof-of-sql/src/base/database/scalar_and_i256_conversions.rs
```diff
@@ -40,10 +40,10 @@ pub fn convert_i256_to_scalar<S: Scalar>(value: &i256) -> Option<S> {
...
         let (low, high) = abs_value.to_parts();
         let limbs = [
-            low as u64,
-            (low >> 64) as u64,
-            high as u64,
-            (high >> 64) as u64,
+            u64::try_from(low).unwrap_or(u64::MAX),
+            u64::try_from(low >> 64).unwrap_or(u64::MAX),
+            u64::try_from(high).unwrap_or(u64::MAX),
+            u64::try_from(high >> 64).unwrap_or(u64::MAX),
```
JayWhite2357, line 45: The original code was actually intentionally truncating. So, this should be allowed

crates/proof-of-sql/src/base/encode/scalar_varint.rs
```diff
@@ -25,7 +25,7 @@ pub fn write_u256_varint(buf: &mut [u8], mut zig_x: U256) -> usize {
     while zig_x.high != 0 || zig_x.low >= 0b1000_0000 {
         // we read the next 7 bits from `zig_x` casting to u8 and setting
         // the 8-th bit to 1 to indicate that we still need to write more bytes to buf
-        buf[pos] = (zig_x.low as u8) | 0b1000_0000;
+        buf[pos] = u8::try_from(zig_x.low).unwrap_or(0b1000_0000);
```
JayWhite2357, line 28: The original code is correct and this should be allowed.

crates/proof-of-sql/src/base/encode/varint_trait.rs
```diff
@@ -194,12 +194,12 @@ impl VarInt for u64 {
         let mut i = 0;
 
         while n >= 0x80 {
-            dst[i] = MSB | (n as u8);
+            dst[i] = MSB | u8::try_from(n).unwrap_or(u8::MAX);
```
JayWhite2357, line 197: We should allow this truncation, since it was correct.

crates/proof-of-sql/src/base/encode/varint_trait.rs
```diff
@@ -194,12 +194,12 @@ impl VarInt for u64 {
...
             i += 1;
             n >>= 7;
         }
 
-        dst[i] = n as u8;
+        dst[i] = u8::try_from(n).unwrap_or(u8::MAX);
```
JayWhite2357, line 202: We should allow this truncation, since it was correct.

crates/proof-of-sql/src/proof_primitive/dory/dory_vmv_helper.rs
```diff
@@ -56,7 +56,7 @@ pub(super) fn compute_T_vec_prime(
 
     prover_setup.blitzar_msm(
         &mut blitzar_commits,
-        data_size as u32,
+        u32::try_from(data_size).unwrap_or(u32::MAX),
```
JayWhite2357, line 59: rather than `unwrap_or`, I think this one should be `.expect("the size of F should fit in u32")`.

crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -170,7 +170,7 @@ impl CanonicalDeserialize for PublicParameters {
     ) -> Result<Self, SerializationError> {
         // Deserialize max_nu (u64)
         let max_nu_u64 = u64::deserialize_with_mode(&mut reader, compress, validate)?;
-        let max_nu = max_nu_u64 as usize;
+        let max_nu: usize = max_nu_u64.try_into().unwrap_or(usize::MAX);
```
JayWhite2357, line 173: Let's use `.map_err(|_| SerializationError::InvalidData)` here instead of `.unwrap_or`.

crates/proof-of-sql/src/sql/proof/provable_query_result.rs
JayWhite2357, line 1: For `cast_possible_truncation` in this file, I think we should just allow it, but add a TODO comment indicating that this should probably be handled properly and the `allow(cast_possible_truncation)` is not a proper fix.

JayWhite2357 (changes requested):
This looks good. I think these casting lints are challenging to resolve properly, so there are some changes I would like to see.
In general, I'm not super concerned about lints in test, benches, or example code. So, feel free to allow lints in those situations fairly liberally.

crates/proof-of-sql/src/base/encode/scalar_varint.rs
JayWhite2357: This may be why tests are failing.

crates/proof-of-sql/src/base/encode/varint_trait.rs
JayWhite2357: This may be why tests are failing.

crates/proof-of-sql/src/base/encode/varint_trait.rs
JayWhite2357: This may be why tests are failing.

crates/proof-of-sql/src/proof_primitive/dory/pack_scalars.rs
```diff
@@ -325,7 +325,10 @@ pub fn bit_table_and_scalars_for_packed_msm(
     let bit_table_sub_commits_sum = bit_table.iter().sum::<u32>() as usize;
 
     // Add offsets to handle signed values to the bit table.
-    bit_table.extend(iter::repeat(BYTE_SIZE as u32).take(OFFSET_SIZE + committable_columns.len()));
+    bit_table.extend(
+        iter::repeat(u32::try_from(BYTE_SIZE).unwrap_or(u32::MAX))
```
JayWhite2357, line 329:
```suggestion
        iter::repeat(u32::try_from(BYTE_SIZE).expect("BYTE_SIZE fits in u32"))
```

JayWhite2357 (approved):

PR #273, opened 2024-10-16
feat: add `find_bigdecimals`

iajoiner:

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/crates/proof-of-sql/src/utils/mod.rs
+++ b/crates/proof-of-sql/src/utils/mod.rs
@@ -1,0 +1,3 @@
+//! This module contains utilities for working with the library
+/// Parse DDLs and find bigdecimal columns
+pub mod parse;
--- a/crates/proof-of-sql/src/utils/parse.rs
+++ b/crates/proof-of-sql/src/utils/parse.rs
@@ -1,0 +1,94 @@
+use crate::base::map::IndexMap;
+use alloc::{
+    string::{String, ToString},
+    vec::Vec,
+};
+use sqlparser::{
+    ast::{DataType, ExactNumberInfo, Statement},
+    dialect::GenericDialect,
+    parser::Parser,
+};
+
+/// Parse a DDL file and return a map of table names to bigdecimal columns
+///
+/// # Panics
+/// Panics if there is an error parsing the SQL
+#[must_use]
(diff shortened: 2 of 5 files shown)


crates/proof-of-sql/src/utils/parse.rs
```diff
@@ -0,0 +1,91 @@
...
+};
+
+/// Parse a DDL file and return a map of table names to bigdecimal columns
+///
+/// # Panics
+/// Panics if there is an error parsing the SQL
+#[must_use]
+pub fn find_bigdecimals(queries: &str) -> IndexMap<String, Vec<String>> {
+    let dialect = GenericDialect {};
+    let ast = match Parser::parse_sql(&dialect, queries) {
```
stuarttimwhite, line 16: NIT: Could do unwrap_or_else. Not too concerned about it though.

crates/proof-of-sql/src/utils/mod.rs
```diff
@@ -0,0 +1,3 @@
+//! This module contains utilities for working with the library
+/// Parse DDLs and find bigdecimal columns
+pub mod parse;
```
stuarttimwhite, line 3: Could this be more descriptive? Perhaps something like ddl_parser? If I were to just look at the file path, I would instinctively think that it should go in the parser crate.

stuarttimwhite (changes requested):
Just a couple small things

crates/proof-of-sql/src/utils/parse.rs
```diff
@@ -0,0 +1,91 @@
...
+    dialect::GenericDialect,
+    parser::Parser,
+};
+
+/// Parse a DDL file and return a map of table names to bigdecimal columns
+///
+/// # Panics
+/// Panics if there is an error parsing the SQL
+#[must_use]
+pub fn find_bigdecimals(queries: &str) -> IndexMap<String, Vec<String>> {
```
stuarttimwhite, line 14: I don't know for sure if this is necessary, but I will also need precision and scale for the decimals.

iajoiner: @stuarttimwhite Ah. Since we actually have decimals with precisions beyond Decimal256 what about returning `IndexMap<String, Vec<(u8, i8)>>` instead?

stuarttimwhite (approved):
Looks good!

PR #274, opened 2024-10-17
fix: implement `TryFrom<BigInt> for MontScalar<T>` properly

JayWhite2357:
# Rationale for this change

`TryFrom<BigInt> for MontScalar<T>` is implemented improperly, and never overflows

The implementation is fixed.

# Are these changes tested?

Yes. Tests are added.


Dustin-Ray (approved):
LGTM - Good catch

PR #278, opened 2024-10-17
ci: the run_ci_checks shell script should not close the terminal if a failure occurs

jacobtrombetta:
# Rationale for this change
The command `exit 1` is called if a failure occurs when executing the commands in the `scripts/run_ci_checks` shell script. The `exit 1` command closes the terminal in places like VSCode, the Ubuntu terminal, and Terminator. This makes debugging failures difficult since the tests results are not visible to the user.

- The `run_ci_checks.sh` shell script is updated to count the number of test failures when executing the commands and leave the terminal open after the script has run for users to see the test results.

# Are these changes tested?
Yes

### DIFF
--- a/scripts/run_ci_checks.sh
+++ b/scripts/run_ci_checks.sh
@@ -45,13 +45,19 @@ echo "$cargo_commands"
 echo "========================="
 
 # Execute the commands
+failed_tests=0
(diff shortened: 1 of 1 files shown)


Dustin-Ray (approved):
LGTM, thanks a bunch.

PR #261, opened 2024-10-11
refactor: remove `query!` macro

akhilender-bongirwar:
Please be sure to look over the pull request guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md#submit-pr.

# Please go through the following checklist
- [x] The PR title and commit messages adhere to guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md. In particular `!` is used if and only if at least one breaking change has been introduced.
- [x] I have run the ci check script with `source scripts/run_ci_checks.sh`.

# Rationale for this change

This PR addresses part 1 of #236  by removing the use of the `query!` macro in tests and replacing it with explicit construction of the query expressions.

-  [x]  Remove uses of the `query!` macro and replace it with explicit uses of the tested methods.

# What changes are included in this PR?
- Removed instances of the query! macro from the test files, specifically in sxt-proof-of-sql/crates/proof-of-sql/src/sql/parse/query_expr_tests.rs
- Replaced the macro usage with explicit query construction

# Are these changes tested?
The change is in tests.

PR #285, opened 2024-10-21
refactor: remove `query!` macro and replace

### LINKED ISSUE #236: Remove `query!` macro.
# Background and Motivation

Rust macros are a bit of a anti-pattern. In particular, the `query!` macro, which is used as a testing utility macro to test `QueryExpr`, obfuscates the tests.

# Changes Required

- [ ] Remove uses of the `query!` macro and replace it with explicit uses of the tested methods. For example, https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/e2eb590ec5f100ae048c85474dc9f576901303cb/crates/proof-of-sql/src/sql/parse/query_expr_tests.rs#L1166-L1170 can be replaced by
    ```rust
    let query_text = "select max(i) max_sal, i0 d, count(i0) from t group by i0, i1 order by max_sal";
    let (t, accessor) = get_test_accessor();
    let intermediate_ast = SelectStatementParser::new().parse(query_text).unwrap();
    let query = QueryExpr::<NaiveCommitment>::try_new(intermediate_ast, t.schema_id(), &accessor).unwrap();
    ```
- [ ] Refactor or remove `get_test_accessor` so that it is explicit what the tables, columns, and data types are in the `SchemaAccessor`. For example, the above query has a `t` table with the columns `i`, `i0`, and `i1`, all with `BigInt` type. These columns should also be renamed to be slightly less generic and/or more helpful.
- [ ] Remove the `query!` macro.

NOTE: the macro is almost exclusively a macro that builds the query text. The actual functionality being tested should be essentially unchanged.

Harsh9485:

- [X] The PR title and commit messages adhere to guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md. In particular `!` is used if and only if at least one breaking change has been introduced.
- [X] I have run the ci check script with `source scripts/run_ci_checks.sh`.

# Rationale for this change

Fixes #236
/claim #236

This PR addresses part 1 of https://github.com/spaceandtimelabs/sxt-proof-of-sql/issues/236 by removing the use of the query! macro in tests and replacing it with explicit construction of the query expressions.

* Removed instances of the query! macro from the test files, specifically in sxt-proof-of-sql/crates/proof-of-sql/src/sql/parse/query_expr_tests.rs
* Replaced the macro usage with explicit query construction

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/parse/query_expr_tests.rs
+++ b/crates/proof-of-sql/src/sql/parse/query_expr_tests.rs
@@ -1,6 +1,7 @@
 use super::ConversionError;
 use crate::{
     base::{
+        commitment::naive_commitment::NaiveCommitment,
         database::{ColumnType, TableRef, TestSchemaAccessor},
         map::{indexmap, IndexMap, IndexSet},
     },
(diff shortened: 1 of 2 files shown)


iajoiner:
Really thanks for your PR!
Please do NOT include your package-lock.json. In fact please add it to .gitignore.

akhilender-bongirwar:
@Harsh9485,  The part 1 and part 3 of this issue are already fixed in my PR #261

Issue #235, opened 2024-10-07
Replace `proof-of-sql-parser` with `sqlparser`.


JayWhite2357:
# Background and Motivation
Currently, we have an in-house parser that is built on the `lalrpop` parser-generator. This has been good while the supported syntax has been simple. However, as the supported syntax has grown, we need a more comprehensive parser.

The [`sqlparser`](https://crates.io/crates/sqlparser) crate is the parser used by [DataFusion](https://datafusion.apache.org/), which is part of the Arrow ecosystem. It is a feature-rich parser that ultimately will require less code maintenance. It is `no_std` compatible, so there should be no issues integrating it.

# Changes Required

- [ ] Replace `proof-of-sql-parser` usage within the `proof-of-sql` crate with `sqlparser` usage.
- [ ] Remove the workspace and simplify the repo to be a single crate: the `proof-of-sql` crate.

JayWhite2357:
@iajoiner may be able to provide a bit of guidance on this issue, however, there is no concrete plan for this.
It would be wise to discuss this plan with @iajoiner and myself to ensure that things are heading in the right direction before building in earnest.

varshith257:
@JayWhite2357 I will connect on discord for the same

varshith257:
Hi @JayWhite2357 and @iajoiner,I'd like to proceed with the transition from the LALRPOP-based parser to sqlparser for the proof-of-sql-parser crate. I have gone through current implementation and working of proof-of-sql parser. Here's the approach I have in mind:

- [ ] I will add the sqlparser crate to handle the parsing of SQL queries.

- [ ] Then I move ahead to the replace the LALRPOP Parser
   - [ ] The LALRPOP-based parser defined in `sql.lalrpop` will be replaced by a new parsing function that utilizes sqlparser to parse
   - [ ] This will specifically focus on the SELECT statements following PostgreSQL syntax as it currently does(Select Statement Parsing, Expressions and Operators, Lexer Rules, ORDER/GROUP BY, Literals and Identifiers etc...)
  

- [ ] Next I step to map the sqlparser AST to existing intermediate AST (intermediate_ast::SelectStatement and other structures) to minimize changes to the rest of the crate.This will ensure the current logic built on top of the AST stays intact

- [ ] I will thoroughly test the changes incorporate at every stage to ensure all current functionality is maintained.

- [ ] I think we can also adapt the existing `ParseError` and `ParseResult` types to handle errors from sqlparser


[...]

PR #286, opened 2024-10-21
feat!: replace proof-of-sql-parser with sqlparser

varshith257:

# Rationale for this change

Part of #235 

# Are these changes tested?

Docs Reference:

https://docs.rs/sqlparser/latest/sqlparser/ast/


crates/proof-of-sql/src/sql/parse/query_expr.rs
```diff
@@ -11,8 +11,15 @@ use crate::{
     },
 };
 use alloc::{fmt, vec, vec::Vec};
-use proof_of_sql_parser::{intermediate_ast::SetExpression, Identifier, SelectStatement};
+use proof_of_sql_parser::{
```
JayWhite2357, line 14: Since we're still depending on this crate here, we cannot delete the `proof-of-sql-parser` crate yet.

Once `try_new_from_sqlparser` can be used as a replacement for the old method, we can slowly remove the dependencies here.

varshith257: @JayWhite2357 I think @iajoiner has stated a few alternatives. I will check with the existing documentation and will proceed with the next steps outlined first to replace `Identifier` with `sqlparser::ast::ident`

varshith257: I have removed the crate because it will make us look for alternatives rather than check with the existing ones that I and @iajoiner planned. If anything doesn't have replacements let's keep them if needed in the worst case

varshith257: Sorry! I haven't looked at this https://github.com/spaceandtimelabs/sxt-proof-of-sql/issues/235#issuecomment-2427078444 :) Ok I will follow the commented approach

crates/proof-of-sql-parser/src/intermediate_ast.rs
```diff
@@ -4,9 +4,8 @@
 * https://docs.rs/vervolg/latest/vervolg/ast/enum.Statement.html
 ***/
 
-use crate::{posql_time::PoSQLTimestamp, Identifier};
+use crate::{intermediate_decimal::IntermediateDecimal, posql_time::PoSQLTimestamp, Identifier};
```
iajoiner, line 7: We don't have `IntermediateDecimal` any more which should make things easier.

iajoiner:
@varshith257 Please rebase since our code base has changed.

varshith257:
Sure!

PR #291, opened 2024-10-21
perf: dynamic Dory commitment computation on the GPU should efficiently handle the offset

jacobtrombetta:
# Rationale for this change
Currently the dynamic Dory commitment computation on the GPU includes zeros for all offset values. This means adding a single commit with a large offset will require a significant chunk of unnecessary memory. This PR handles offsets more efficiently truncating all zero rows of the dynamic Dory structure matrix when crating the Blitzar metadata tables.

- The `blitzar_metadata_tables` module now truncates rows of the dynamic Dory structure matrix that are fully contained in the offset.
- Tests are updated in the `blitzar_metadata_tables` module to test the new behavior.
- The `Gamma_2` value in the `dynamic_dory_commitment_helper_gpu` module now uses the offset.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/blitzar_metadata_table.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/blitzar_metadata_table.rs
@@ -394,6 +394,7 @@ mod tests {
     #[test]
     fn we_can_populate_blitzar_metadata_tables_with_simple_column_and_non_trivial_offsets() {
         let committable_columns = [CommittableColumn::TinyInt(&[1])];
+
         let expected_bit_table = vec![8, 8];
 
         let offset = 0;
@@ -437,9 +438,10 @@ mod tests {
     fn we_can_populate_blitzar_metadata_tables_with_simple_column_and_offsets_with_4_columns() {
         let committable_columns = [CommittableColumn::TinyInt(&[1])];
 
-        let offsets = vec![4, 8, 12];
         let expected_bit_table = vec![8, 8];
         let expected_length_table = vec![4, 4];
+
+        let offsets = vec![4, 8, 12];
         for &offset in &offsets {
             assert_blitzar_metadata(
                 &committable_columns,
@@ -479,9 +481,10 @@ mod tests {
     fn we_can_populate_blitzar_metadata_tables_with_simple_column_and_offsets_with_8_columns() {
         let committable_columns = [CommittableColumn::TinyInt(&[1])];
 
-        let offsets = vec![16, 24, 32, 40, 48, 56];
         let expected_bit_table = vec![8, 8];
         let expected_length_table = vec![8, 8];
+
+        let offsets = vec![16, 24, 32, 40, 48, 56];
         for &offset in &offsets {
             assert_blitzar_metadata(
                 &committable_columns,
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql/src/proof_primitive/dory/blitzar_metadata_table.rs
```diff
@@ -153,6 +155,10 @@ pub fn create_blitzar_metadata_tables(
         (0, 0)
     };
 
+    // We will ignore the rows that are zero from the offsets.
+    let offset_row = row_and_column_from_index(offset).0;
```
JayWhite2357, line 159: I'm suspicious here.

I think it should be this:
```suggestion
    let offset_row = row_and_column_from_index(offset+1).0-1;
```
Can you add some tests that test with some non-trivial offsets? 0, 1, and 2 are not enough.

I'd like to see tests with non-empty columns for the following offsets at least:
`0,1,2,3,4,5,10,16,20,50,64`
Some fill up rows completely, some do not.

JayWhite2357 (changes requested):
I want some more tests with the offsets.

crates/proof-of-sql/src/proof_primitive/dory/blitzar_metadata_table.rs
Dustin-Ray: Is it possible to add an accessor method instead of using tuple indexing? Its difficult to read and maintain in general.

Dustin-Ray: > I'm suspicious here.
> 
> I think it should be this:
> 
> Can you add some tests that test with some non-trivial offsets? 0, 1, and 2 are not enough.
> 
> I'd like to see tests with non-empty columns for the following offsets at least: `0,1,2,3,4,5,10,16,20,50,64` Some fill up rows completely, some do not.

This is super hard to read

Dustin-Ray: > (offset+1).0-1;

This is really difficult to read. If its test code its fine but I still am not a fan of this syntax

jacobtrombetta: @JayWhite2357 the `max_height` from the `matrix_size` function takes care of offset you are thinking about. It adds the `+1`, so offset doesn't need to worry.

I added tests for the requested offsets.

jacobtrombetta: > Is it possible to add an accessor method instead of using tuple indexing? Its difficult to read and maintain in general.

Removed tuple indexing. Thanks!

Dustin-Ray:
@jacobtrombetta just run cargo format or the check script and then if @JayWhite2357 is good with tests we can merge

crates/proof-of-sql/src/proof_primitive/dory/blitzar_metadata_table.rs
JayWhite2357: Ok, I'm good with this. The tests proved me wrong, but it still took a while for me to understand why this is correct. `offset` is a length, not an index.

JayWhite2357 (approved):
LGTM, thanks!

PR #283, opened 2024-10-20
refactor!: remove `IntermediateDecimal` in favor of `BigDecimal`

JayWhite2357:
# Rationale for this change

Currently, the code handling decimal parsing and conversions is somewhat scattered. This issue seeks to resolve this by using the `bigdecimal::BigDecimal` type for as much as possible.

# What changes are included in this PR?

Remove `IntermediateDecimal` in favor of `BigDecimal`
Please read the individual commit messages for detailed changes.

# Are these changes tested?
Yes.

PR #294, opened 2024-10-22
refactor: tidy up `decimal` module

JayWhite2357:
# Rationale for this change

After #283, the `math::decimal` module is a bit untidy, especially around error types and tests. 

This reorganizes the module, tidies things up, and adds some more thorough unit tests.
Please look at the individual commits.

# Are these changes tested?
Yes


crates/proof-of-sql/src/base/math/precision.rs
```diff
@@ -0,0 +1,68 @@
...
+/// e.g. precision above 75/76/whatever set by Scalar
+/// or non-positive aka `InvalidPrecision`
+pub struct InvalidPrecisionError {
+    precision: String,
+}
+
+#[derive(Eq, PartialEq, Debug, Clone, Hash, Serialize, Copy)]
+/// limit-enforced precision
+pub struct Precision(u8);
+pub(super) const MAX_SUPPORTED_PRECISION: u8 = 75;
```
Dustin-Ray, line 17: this could use a doc comment

crates/proof-of-sql/src/sql/proof_exprs/comparison_util.rs
```diff
@@ -83,13 +78,7 @@ pub(crate) fn scale_and_subtract<'a, S: Scalar>(
...
         );
         // Check if the precision is valid
-        let _max_precision = Precision::new(max_precision_value).map_err(|_| {
-            ConversionError::DecimalConversionError {
-                source: DecimalError::InvalidPrecision {
-                    error: max_precision_value.to_string(),
-                },
-            }
-        })?;
+        let _max_precision = Precision::new(max_precision_value)?;
```
Dustin-Ray, line 81: why is this unused? would it be better to have a check function that didnt require an unused variable?

PR #295, opened 2024-10-22
feat: proversetup fileio

Dustin-Ray:
# Rationale for this change

This PR adds support for a conversion from pre-generated public parameters and a blitzar handle into a ProverSetup. This helps skip an expensive step to generate parameters and should significantly decrease the amount of time to load the parameters. It serves as a step towards file i/o for the parameters.

- [x] convert public params and blitzar handle into a proversetup

# Are these changes tested?

existing tests pass. a round trip test is forthcoming in a future PR.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
@@ -18,6 +18,7 @@ use std::{
 /// Note: even though `H_1` and `H_2` are marked as blue, they are still needed.
 ///
 /// Note: `Gamma_1_fin` is unused, so we leave it out.
+#[derive(Clone)]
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/proof_primitive/dory/public_parameters.rs
```diff
@@ -18,6 +18,7 @@ use std::{
 /// Note: even though `H_1` and `H_2` are marked as blue, they are still needed.
 ///
 /// Note: `Gamma_1_fin` is unused, so we leave it out.
+#[derive(Clone)]
```
JayWhite2357, line 21: What is this for?

JayWhite2357 (approved):

JayWhite2357 (approved):

PR #316, opened 2024-10-25
perf: make dynamic dory cpu commitment computation more efficient

JayWhite2357:
# Rationale for this change

In order for lightweight environments, without GPUs, to compute commitments, we should be able to compute commitments faster on a CPU.

The dynamic dory commitment computation now computes fewer Scalar Multiplications.
Also, rayon is leveraged for parallelism.

# Are these changes tested?
Yes.


jacobtrombetta (approved):
LGTM

PR #325, opened 2024-10-28
feat: generate params utility

Dustin-Ray:
# Rationale for this change

Initial parameter generation can be expensive depending on the size of the setups. Since the prover needs to be restarted as easily and quickly as possible, parameters should be pre-generated in some transparent process and made available for easy access.

- [x] Add a CLI for generation of public parameters and setups
- [x] Support file I/O for various proof components

# Are these changes tested?
- [x] Round trip file I/O test
- [x] Test digests successfully created


crates/proof-of-sql/utils/generate-parameters/main.rs
```diff
@@ -0,0 +1,299 @@
...
+fn main() {
+    // Set the BLITZAR_PARTITION_WINDOW_WIDTH environment variable
+    env::set_var(BLITZAR_PARTITION_WINDOW_WIDTH, "14");
+
+    // Confirm that it was set by reading it back
+    match env::var(BLITZAR_PARTITION_WINDOW_WIDTH) {
+        Ok(value) => {
+            println!("Environment variable {BLITZAR_PARTITION_WINDOW_WIDTH} set to {value}");
+        }
+        Err(e) => println!("Failed to set {BLITZAR_PARTITION_WINDOW_WIDTH}: {e}"),
```
iajoiner, line 57: Are we supposed to exit if we can't set the env variable? If so let's use `std::process::exit(-1)`.

crates/proof-of-sql/utils/generate-parameters/main.rs
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    // Parse command-line arguments
+    let args = Args::parse();
+
+    // Ensure the target directory exists
+    match fs::create_dir_all(&args.target) {
+        Ok(()) => generate_parameters(&args),
+        Err(_) => eprint!(
```
iajoiner, line 66: Same. A nonzero exit code indicates error.

Dustin-Ray: good catch thank you

crates/proof-of-sql/utils/generate-parameters/main.rs
```diff
@@ -0,0 +1,304 @@
...
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    /// The value for `nu` (number of public parameters)
+    #[arg(short, long, default_value_t = 8)]
+    nu: usize,
+
+    /// Mode for generating parameters: "p" for Prover, "v" for Verifier
+    #[arg(short, long, default_value = "pv")]
+    mode: String,
```
iajoiner, line 37: I'd make this an enum and allow Prover, Verifier and Both as options.

Dustin-Ray: sure sounds good to me

crates/proof-of-sql/utils/generate-parameters/main.rs
```diff
@@ -0,0 +1,322 @@
...
+    target: String,
+}
+
+// An enum representing possible modes of operation,
+// abbreviated to keep clap commands concise, instead
+// of requiring the user to type "ProverAndVerifier",
+// they type "pv"
+#[derive(Debug, Clone, ValueEnum)]
+enum Mode {
+    P,  //Prover
```
iajoiner, line 52: I think using the names "Prover", "Verifier" and "Both" is useful. Shortcuts can be provided in Clap.

Dustin-Ray: I tried this and it was forcing me to type prover, verifier, and proverandverifier in the command line to select these

crates/proof-of-sql/utils/generate-parameters/main.rs
```diff
@@ -0,0 +1,326 @@
...
+    // Set the BLITZAR_PARTITION_WINDOW_WIDTH environment variable
+    env::set_var(BLITZAR_PARTITION_WINDOW_WIDTH, "14");
+
+    // Confirm that it was set by reading it back
+    match env::var(BLITZAR_PARTITION_WINDOW_WIDTH) {
+        Ok(value) => {
+            println!("Environment variable {BLITZAR_PARTITION_WINDOW_WIDTH} set to {value}");
+        }
+        Err(e) => {
+            eprintln!("Failed to set {BLITZAR_PARTITION_WINDOW_WIDTH}: {e}");
```
Dustin-Ray, line 67: this is not a fatal failure it will just make the process take a bit longer, safe to proceed here

iajoiner (approved):
Really thanks @Dustin-Ray !

PR #331, opened 2024-10-31
feat: add `ColumnarValue` operations

iajoiner:

# Rationale for this change
To prepare to remove `table_length` from `ProofExpr::result_evaluate` and return `ColumnarValue` we need to add operations on `ColumnarValue`.

- add comparison operations on `ColumnarValue`.
- add numerical operations on `ColumnarValue`.

# Are these changes tested?

Will be tested in a future PR.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/comparison_util.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/comparison_util.rs
@@ -31,12 +31,12 @@ fn unchecked_subtract_impl<'a, S: Scalar>(
     Ok(result)
 }
 
-#[allow(
-    clippy::missing_panics_doc,
-    reason = "precision and scale are validated prior to calling this function, ensuring no panic occurs"
-)]
 /// Scale LHS and RHS to the same scale if at least one of them is decimal
 /// and take the difference. This function is used for comparisons.
+///
+/// # Panics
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql/src/sql/proof_exprs/comparison_util.rs
```diff
@@ -31,6 +31,70 @@ fn unchecked_subtract_impl<'a, S: Scalar>(
     Ok(result)
 }
 
+#[allow(
+    clippy::missing_panics_doc,
+    reason = "precision and scale are validated prior to calling this function, ensuring no panic occurs"
+)]
+/// Scale LHS and RHS to the same scale if at least one of them is decimal
+/// and take the difference. This function is used for comparisons.
```
JayWhite2357, line 39: You should only use `allow(clippy::missing_panics_doc)` if the function is guarenteed to not panic. If invalid parameters are passed, then it will/should panic, so we should document that.
```suggestion
/// Scale LHS and RHS to the same scale if at least one of them is decimal
/// and take the difference. This function is used for comparisons.
///
/// # Panics
/// The precision and scale passed must be valid. A panic may occur if they are not.
```

JayWhite2357 (changes requested):

JayWhite2357 (approved):

PR #333, opened 2024-11-01
feat: add `column_struct_operation.rs`

iajoiner:

# Rationale for this change
The equivalent of `owned_column_operation.rs` for `Column`s.

# Are these changes tested?

### DIFF
--- a/crates/proof-of-sql/src/base/database/column_struct_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_struct_operation.rs
@@ -1,0 +1,2284 @@
+use super::{ColumnOperationError, ColumnOperationResult};
+use crate::base::{
+    database::{
+        column_operation::{
+            eq_decimal_columns, ge_decimal_columns, le_decimal_columns, slice_and, slice_eq,
+            slice_eq_with_casting, slice_ge, slice_ge_with_casting, slice_le,
+            slice_le_with_casting, slice_not, slice_or, try_add_decimal_columns, try_add_slices,
+            try_add_slices_with_casting, try_multiply_decimal_columns, try_multiply_slices,
+            try_multiply_slices_with_casting, try_subtract_decimal_columns, try_subtract_slices,
+            try_subtract_slices_left_upcast, try_subtract_slices_right_upcast,
+        },
+        Column,
+    },
+    scalar::Scalar,
+};
+use alloc::vec::Vec;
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/base/database/column_struct_operation.rs
```diff
@@ -0,0 +1,2284 @@
...
+                operator: BinaryOperator::Or,
+                left_type: self.column_type(),
+                right_type: rhs.column_type(),
+            }),
+        }
+    }
+
+    /// Element-wise equality check for two columns
+    #[allow(clippy::too_many_lines)]
+    pub fn element_wise_eq(&self, rhs: &Self, alloc: &'a Bump) -> ColumnOperationResult<Self> {
```
JayWhite2357, line 119: Since all of these functions are identical (I think), could we just replace them all with a single function that we can pass in a few slice_op parameters?

```suggestion
    pub fn element_wise_binary_compare_op<F,G,H>(&self, rhs: &Self, alloc: &'a Bump, slice_op: F, slice_op_with_casting: G, decimal_slice_op: H) -> ColumnOperationResult<Self> {
```

Not sure if the generic parameters are needed.

crates/proof-of-sql/src/base/database/column_struct_operation.rs
```diff
@@ -0,0 +1,2284 @@
...
+    pub fn element_wise_add(&self, rhs: &Self, alloc: &'a Bump) -> ColumnOperationResult<Self> {
+        if self.len() != rhs.len() {
+            return Err(ColumnOperationError::DifferentColumnLength {
+                len_a: self.len(),
+                len_b: rhs.len(),
+            });
+        }
+        match (&self, &rhs) {
+            (Self::TinyInt(lhs), Self::TinyInt(rhs)) => Ok(Self::TinyInt(
+                alloc.alloc_slice_copy(try_add_slices(lhs, rhs)?.as_ref()),
```
JayWhite2357, line 824: Same comment here as above. I don't like unnecessary copy. This is how it could look using the `process_results` method from `itertools`:
```suggestion
                try_add_slices(lhs, rhs).process_results(|iter| alloc.alloc_slice_copy(iter)),
```

iajoiner:
Closed since we are going to use the approach from #359.

PR #361, opened 2024-11-08
feat: compute min and max of indices for multi-table queries

iajoiner:

# Rationale for this change

# Are these changes tested?


crates/proof-of-sql/src/sql/proof/proof_plan.rs
```diff
@@ -46,6 +46,21 @@ pub trait ProofPlan: Debug + Send + Sync + ProverEvaluate {
...
     /// Return all the tables referenced in the Query
     fn get_table_references(&self) -> IndexSet<TableRef>;
+
+    /// Return the row number range of tables referenced in the Query
+    ///
+    /// Basically we are looking for the smallest offset and the largest offset + length
+    /// so that we have an index range of the table rows that the query is referencing.
+    fn get_index_range(&self, accessor: &dyn MetadataAccessor) -> (usize, usize) {
+        self.get_table_references().iter().fold(
+            (usize::MAX, 0),
```
JayWhite2357, line 56: * What happens if there are no tables referenced? I would make the output optional rather than have an incorrect default.
* I would prefer to see this take `table_references` as an argument rather than `self`. That way, `table_references` can be used outside this method multiple times if needed without needing to call `get_table_references` twice.

crates/proof-of-sql/src/sql/proof/query_proof.rs
```diff
@@ -47,15 +47,15 @@ impl<CP: CommitmentEvaluationProof> QueryProof<CP> {
         accessor: &impl DataAccessor<CP::Scalar>,
         setup: &CP::ProverPublicSetup<'_>,
     ) -> (Self, ProvableQueryResult) {
-        let table_length = expr.get_length(accessor);
-        let num_sumcheck_variables = cmp::max(log2_up(table_length), 1);
-        let generator_offset = expr.get_offset(accessor);
```
JayWhite2357, line 52: Shouldn't `get_length` and `get_offset` be removed from the trait since it is now dead code?

crates/proof-of-sql/src/sql/proof/proof_plan.rs
iajoiner: @JayWhite2357 If there are no tables referenced the amount of rows here is 1 since the only possible situation is `EmptyExec`. Max of intermediate `ProofPlan`s can be higher with unions but that requires more refactoring (mostly replacement of accessors with tables in `result_evaluate` and `final_round_evaluate`.

crates/proof-of-sql/src/sql/proof/query_proof.rs
iajoiner: Yup. I originally wanted to do it in a future PR but can of course do it now.

crates/proof-of-sql/src/sql/proof/query_proof_test.rs
```diff
@@ -116,7 +109,13 @@ fn verify_a_trivial_query_proof_with_given_offset(n: usize, offset_generators: u
         offset: offset_generators,
         ..Default::default()
     };
-    let accessor = UnimplementedTestAccessor::new_empty();
+    let column: Vec<i64> = vec![0_i64; n];
+    let accessor = OwnedTableTestAccessor::<InnerProductProof>::new_from_table(
+        "sxt.test".parse().unwrap(),
+        owned_table([bigint("a1", column)]),
```
JayWhite2357, line 115: Very minor NIT. Slightly more readable and compact in my opinion.
```suggestion
    let accessor = OwnedTableTestAccessor::<InnerProductProof>::new_from_table(
        "sxt.test".parse().unwrap(),
        owned_table([bigint("a1", [0; n])]),
```

JayWhite2357 (approved):

crates/proof-of-sql/src/base/database/mod.rs
```diff
@@ -66,6 +66,7 @@ pub use expression_evaluation_error::{ExpressionEvaluationError, ExpressionEvalu
 mod test_accessor;
 pub use test_accessor::TestAccessor;
 #[cfg(test)]
+#[allow(unused_imports)]
```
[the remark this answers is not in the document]
iajoiner: @JayWhite2357 Yes it is. If it is indeed intended to be useless I will remove it.

crates/proof-of-sql/src/sql/proof/proof_plan.rs
```diff
@@ -19,15 +19,12 @@ pub trait ProofPlan: Debug + Send + Sync + ProverEvaluate {
...
-    /// The offset of the query, that is, how many rows to skip before starting to read the input table
-    fn get_offset(&self, accessor: &dyn MetadataAccessor) -> usize;
-
     /// Check if the input table is empty
     fn is_empty(&self, accessor: &dyn MetadataAccessor) -> bool {
-        self.get_length(accessor) == 0
+        let table_refs = self.get_table_references();
+        table_refs
+            .iter()
+            .all(|table_ref| accessor.get_length(*table_ref) == 0)
```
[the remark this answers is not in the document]
iajoiner: Yup. I think `is_empty` is really only used once. The real question is whether we should even single out that as a separate case.

PR #371, opened 2024-11-13
refactor!: replace output of `result_evaluate` and `final_round_evaluate` of `ProofPlan` with `Table`

iajoiner:

# Rationale for this change
To make `ProofPlan`s composable it is more natural for us to return `Table`s in `result_evaluate` and `final_round_evaluate` in `ProofPlan`.

- replace output of `result_evaluate` and `final_round_evaluate` of `ProofPlan` with `Table`
- add `Table::columns`, `ProvableQueryResult::new_from_table` to simplify the code

# Are these changes tested?

Yes.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof/provable_query_result.rs
+++ b/crates/proof-of-sql/src/sql/proof/provable_query_result.rs
@@ -1,6 +1,6 @@
 use super::{decode_and_convert, decode_multiple_elements, ProvableResultColumn, QueryError};
 use crate::base::{
-    database::{Column, ColumnField, ColumnType, OwnedColumn, OwnedTable},
+    database::{Column, ColumnField, ColumnType, OwnedColumn, OwnedTable, Table},
     polynomial::compute_evaluation_vector,
     scalar::Scalar,
 };
@@ -53,6 +53,13 @@ impl ProvableQueryResult {
             table_length,
             data,
         }
+    }
+
+    /// Form intermediate query result from a table
+    #[must_use]
+    pub fn new_from_table<S: Scalar>(table: &Table<S>) -> Self {
+        let columns = table.columns().copied().collect::<Vec<_>>();
+        Self::new(table.num_rows() as u64, &columns)
     }
 
     /// Form intermediate query result from index rows and result columns
--- a/crates/proof-of-sql/src/sql/proof/provable_query_result.rs
+++ b/crates/proof-of-sql/src/sql/proof/provable_query_result.rs
@@ -55,13 +55,6 @@ impl ProvableQueryResult {
         }
     }
 
-    /// Form intermediate query result from a table
-    #[must_use]
-    pub fn new_from_table<S: Scalar>(table: &Table<S>) -> Self {
-        let columns = table.columns().copied().collect::<Vec<_>>();
-        Self::new(table.num_rows() as u64, &columns)
-    }
-
     /// Form intermediate query result from index rows and result columns
     /// # Panics
     ///
@@ -224,5 +217,17 @@ impl ProvableQueryResult {
         assert_eq!(owned_table.num_columns(), self.num_columns());
 
         Ok(owned_table)
+    }
+}
+
+impl<S: Scalar> From<Table<'_, S>> for ProvableQueryResult {
+    fn from(table: Table<S>) -> Self {
+        let num_rows = table.num_rows();
+        let columns = table
+            .into_inner()
+            .into_iter()
+            .map(|(_, col)| col)
+            .collect::<Vec<_>>();
+        Self::new(num_rows as u64, &columns)
     }
 }
--- a/crates/proof-of-sql/src/sql/proof/query_proof_test.rs
+++ b/crates/proof-of-sql/src/sql/proof/query_proof_test.rs
@@ -7,9 +7,9 @@ use crate::{
         database::{
             owned_table_utility::{bigint, owned_table},
             Column, ColumnField, ColumnRef, ColumnType, DataAccessor, OwnedTable,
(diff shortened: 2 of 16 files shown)


crates/proof-of-sql/src/sql/proof/provable_query_result.rs
```diff
@@ -55,6 +55,13 @@ impl ProvableQueryResult {
         }
     }
 
+    /// Form intermediate query result from a table
+    #[must_use]
+    pub fn new_from_table<S: Scalar>(table: &Table<S>) -> Self {
+        let columns = table.columns().copied().collect::<Vec<_>>();
```
JayWhite2357, line 61: Not sure if this would work, but could you just derive `Clone` for `Table`, and then pass `table` by value instead of reference? Then, you can `.clone` upstream only if needed.

Then, you could have `Table` also implement `IntoIterator` and I think do something like `let columns = Vec::from_iter(table)` instead.

crates/proof-of-sql/src/sql/proof/query_proof_test.rs
```diff
@@ -44,9 +44,12 @@ impl ProverEvaluate for TrivialTestProofPlan {
...
         alloc: &'a Bump,
         _accessor: &'a dyn DataAccessor<S>,
-    ) -> Vec<Column<'a, S>> {
+    ) -> Table<'a, S> {
         let col = alloc.alloc_slice_fill_copy(self.length, self.column_fill_value);
-        vec![Column::BigInt(col)]
+        Table::<'a, S>::try_new(indexmap! {
+            "a1".parse().unwrap() => Column::BigInt(col),
+        })
+        .unwrap()
```
JayWhite2357, line 52: Can't you use `table_utility` here? (And other spots)

crates/proof-of-sql/src/sql/proof_plans/empty_exec.rs
```diff
@@ -67,8 +67,8 @@ impl ProverEvaluate for EmptyExec {
         &self,
         _alloc: &'a Bump,
         _accessor: &'a dyn DataAccessor<S>,
-    ) -> Vec<Column<'a, S>> {
-        Vec::new()
+    ) -> Table<'a, S> {
+        Table::<'a, S>::try_new(IndexMap::default()).unwrap()
```
JayWhite2357, line 71: This is where it make sense to set a default number of rows, since this is where that decision gets made.

crates/proof-of-sql/src/sql/proof/provable_query_result.rs
iajoiner: @JayWhite2357 We can though that won't be intuitive since naturally `Item` here should be `(Identifier, Column)` as opposed to `Column`. If you want that instead I can do it but I need to document this.

crates/proof-of-sql/src/sql/proof_plans/group_by_exec.rs
```diff
@@ -288,16 +295,23 @@ impl ProverEvaluate for GroupByExec {
...
             .into_iter()
             .chain(sum_result_columns_iter)
-            .chain(core::iter::once(Column::BigInt(count_column)))
+            .chain(iter::once(Column::BigInt(count_column)))
             .collect::<Vec<_>>();
+        let res = Table::<'a, S>::try_from_iter(
+            self.get_column_result_fields()
+                .into_iter()
+                .map(|field| field.name())
+                .zip(columns.clone().into_iter()),
```
JayWhite2357, line 308: NIT:
We shouldn't need `clone`s/`collect`s.
```suggestion
            .chain(iter::once(Column::BigInt(count_column)));
        let res = Table::<'a, S>::try_from_iter(
            self.get_column_result_fields()
                .into_iter()
                .map(|field| field.name())
                .zip(columns),
```

crates/proof-of-sql/src/sql/proof_plans/group_by_exec.rs
```diff
@@ -288,16 +295,23 @@ impl ProverEvaluate for GroupByExec {
...
+        )
+        .expect("Failed to create table from column references");
         // 5. Produce MLEs
-        res.iter().copied().for_each(|column| {
+        for column in columns {
```
JayWhite2357, line 312:
```suggestion
        for column in res.columns() {
```

crates/proof-of-sql/src/sql/proof_plans/table_exec.rs
```diff
@@ -100,9 +105,9 @@ impl ProverEvaluate for TableExec {
         builder: &mut FinalRoundBuilder<'a, S>,
         alloc: &'a Bump,
         accessor: &'a dyn DataAccessor<S>,
-    ) -> Vec<Column<'a, S>> {
+    ) -> Table<'a, S> {
         let table = self.get_table(accessor);
-        for column in &table {
+        for column in table.inner_table().values() {
```
JayWhite2357, line 110: NIT:
```suggestion
        for column in table.columns() {
```

crates/proof-of-sql/src/sql/proof_plans/table_exec.rs
```diff
@@ -33,17 +33,22 @@ impl TableExec {
...
                     self.table_ref,
                     field.name(),
                     field.data_type(),
-                ))
-            })
-            .collect()
+                )),
+            )
+        }))
+        .expect("Failed to create table from column references")
```
JayWhite2357, line 51: This is problematic if the schema is empty.
Why not just use `accessor.get_table`, which already handles this?

JayWhite2357 (changes requested):

JayWhite2357 (approved):

PR #373, opened 2024-11-15
refactor: simplify `result_evaluations` calculation

JayWhite2357:
# Rationale for this change

In order to port the verifier to Solidity, some simplifications should be made to the Rust verifier to keep the implementations as similar as possible.

The `result_evaluations` calculation in `QueryProof::verify` is simplified.

# Are these changes tested?
Yes, by existing tests.


iajoiner (approved):
Really thanks for the PR!

PR #381, opened 2024-11-19
refactor!: generalize one eval to arbitrary lengths

iajoiner:

- [x] PR #380 has been merged.

# Rationale for this change
When `ProofPlan`s compose it makes no sense to have `table_length` in `VerificationBuilder`. Instead it becomes necessary to record each and every instance of length used for one_eval and pass the info to the verifier. 

- remove `VerificationBuilder::table_length`
- generalize one eval from `input_length` and `output_length` to arbitrary lengths
- add `singleton_one_evaluation` to `SumcheckMleEvaluations`
- add consumption of the lengths in `VerificationBuilder`
- add `TableEvaluation`
- add `one_eval` as an arg to `ProofExpr::verifier_evaluate`
- return `TableEvaluation` in `ProofPlan::verifier_evaluate`
- return `(Table<'a, S>, Vec<usize>)` in `ProofPlan::result_evaluate` to return one eval lengths
- remove `range_length` logic from `FirstRoundBuilder` as `ProofPlan::result_evaluate` already tracks potentially large output lengths in intermediate tables

# Are these changes tested?

Yes.


crates/proof-of-sql/src/sql/proof/query_proof.rs
```diff
@@ -52,6 +52,8 @@ fn get_index_range(
 pub struct QueryProof<CP: CommitmentEvaluationProof> {
     /// Bit distributions
     pub bit_distributions: Vec<BitDistribution>,
+    /// One evaluation lengths
+    pub one_evaluation_lengths: Vec<usize>,
```
JayWhite2357, line 56: The proof expressions, at the very least, shouldn't need to produce/consume the length, since they are all determined by the table they are on. So, perhaps the one_evaluation should be passed into verifier_evaluate for the proof expressions, and possibly into the proof plans as well.)

The reason for this is that most of these don't need to be part of the proof since they are determined by the child nodes.

JayWhite2357 (changes requested):

crates/proof-of-sql/src/sql/proof/query_proof.rs
iajoiner: @JayWhite2357 I agree. The only place where one_eval probably needs to be a part of the proof is the output one used in `FilterExec`. Even in that case what we really need is records of table lengths.

crates/proof-of-sql/src/sql/proof_plans/filter_exec.rs
```diff
@@ -282,6 +298,7 @@ pub(super) fn prove_filter<'a, S: Scalar + 'a>(
     n: usize,
     m: usize,
 ) {
+    builder.push_one_evaluation_length(m);
```
JayWhite2357, line 301: This is inside the `prove_filter`, but the corresponding `consume` is outside the `verify_filter`. Functionally it is correct, but we should keep the prove and verify mirrored. Either move them both in, or move them both out.

JayWhite2357 (changes requested):

JayWhite2357 (approved):

PR #382, opened 2024-11-19
refactor!: `proof_of_sql_parser::intermediate_ast::Identifier` with `sqlparser::ast::Ident` in the proof-of-sql crate 

varshith257:

# Rationale for this change
This PR addresses the need to replace the `proof_of_sql_parser::Identifier` with the `sqlparser::ast::Ident` in the `proof-of-sql` crate as part of a larger transition toward integrating the `sqlparser` .

This change is a subtask of issue #235, with the main goal of streamlining the repository by switching to the `sqlparser` crate and gradually replacing intermediary constructs like `proof_of_sql_parser::intermediate_ast` with `sqlparser::ast`.

- All instances of `proof_of_sql_parser::Identifier` have been replaced with `sqlparser::ast::Ident`
- A few of them required an identifier (e.g. Expression::Column, etc..), which is dependent on the Identifier and will be migrated at the refactoring of Exprs.
-  Every usage of `Identifier` has been updated to maintain the original functionality, ensuring no changes to the logic or behavior.
-  The breaking change here is that `Ident` doesn't support `Copy` trait so we have needed the clones in the places where values are moved
- Deleted the test `we_cannot_convert_a_record_batch_if_it_has_repeated_column_names`  because the `sqlparser` now differentiates between uppercase and lowercase identifiers. Case normalization is no longer applied and `sqlparser` treats `a` and `A` as distinct identifiers.

- Examples are updated  to align with `sqlparser`'s case-sensitive behavior.

# Are these changes tested?

Yes

Part of #235

### DIFF
--- a/crates/proof-of-sql/src/sql/parse/query_context_builder.rs
+++ b/crates/proof-of-sql/src/sql/parse/query_context_builder.rs
@@ -112,14 +112,8 @@ impl<'a> QueryContextBuilder<'a> {
     )]
     fn lookup_schema(&self) -> Vec<(Ident, ColumnType)> {
         let table_ref = self.context.get_table_ref();
-        let mut columns = self.schema_accessor.lookup_schema(*table_ref);
+        let columns = self.schema_accessor.lookup_schema(*table_ref);
         assert!(!columns.is_empty(), "At least one column must exist");
-        // Normalize all column names
-        for (ident, _) in &mut columns {
-            let normalized_name = normalize_ident(ident.clone());
-            *ident = Ident::new(normalized_name);
-        }
-
         columns
     }
 
--- a/crates/proof-of-sql/src/sql/parse/query_context_builder.rs
+++ b/crates/proof-of-sql/src/sql/parse/query_context_builder.rs
@@ -22,6 +22,8 @@ pub struct QueryContextBuilder<'a> {
     context: QueryContext,
     schema_accessor: &'a dyn SchemaAccessor,
 }
+use crate::base::sqlparser::normalize_ident;
+use sqlparser::ast::Ident;
 
 // Public interface
 impl<'a> QueryContextBuilder<'a> {
@@ -36,15 +38,16 @@ impl<'a> QueryContextBuilder<'a> {
     pub fn visit_table_expr(
         mut self,
         table_expr: &[Box<TableExpression>],
-        default_schema: Identifier,
+        default_schema: Ident,
     ) -> Self {
         assert_eq!(table_expr.len(), 1);
         match *table_expr[0] {
             TableExpression::Named { table, schema } => {
-                self.context.set_table_ref(TableRef::new(ResourceId::new(
-                    schema.unwrap_or(default_schema),
-                    table,
-                )));
+                let default_schema_identifier = Identifier::try_from(default_schema)
+                    .expect("Failed to convert default_schema Ident to Identifier");
+                let schema_identifier = schema.unwrap_or(default_schema_identifier);
+                self.context
+                    .set_table_ref(TableRef::new(ResourceId::new(schema_identifier, table)));
             }
         }
         self
@@ -87,12 +90,9 @@ impl<'a> QueryContextBuilder<'a> {
         self
     }
 
-    pub fn visit_group_by_exprs(
-        mut self,
-        group_by_exprs: Vec<Identifier>,
-    ) -> ConversionResult<Self> {
+    pub fn visit_group_by_exprs(mut self, group_by_exprs: Vec<Ident>) -> ConversionResult<Self> {
         for id in &group_by_exprs {
-            self.visit_column_identifier(*id)?;
+            self.visit_column_identifier(id)?;
         }
         self.context.set_group_by_exprs(group_by_exprs);
         Ok(self)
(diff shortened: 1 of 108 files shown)


iajoiner (commented):
@varshith257 Really thanks for your hard work! This is mostly good.
One comment:
Please remove changes to the examples. Changes in the rest of the code base should be sufficient in getting them to work.

crates/proof-of-sql/src/sql/parse/query_context_builder.rs
```diff
@@ -110,17 +110,27 @@ impl<'a> QueryContextBuilder<'a> {
...
+    fn lookup_schema(&self) -> Vec<(Ident, ColumnType)> {
         let table_ref = self.context.get_table_ref();
-        let columns = self.schema_accessor.lookup_schema(*table_ref);
+        let mut columns = self.schema_accessor.lookup_schema(*table_ref);
         assert!(!columns.is_empty(), "At least one column must exist");
+        // Normalize all column names
+        for (ident, _) in &mut columns {
+            let normalized_name = normalize_ident(ident.clone());
+            *ident = Ident::new(normalized_name);
+        }
```
iajoiner, line 121: This can and should be done in a more functional manner.

iajoiner (approved):

iajoiner (changes requested):
Let's remove this change. Parsing should work regardless of casing and regardless of which Accessor we use.

iajoiner (approved):

iajoiner (dismissed):

iajoiner (dismissed):

crates/proof-of-sql/src/sql/parse/query_context_builder.rs
```diff
@@ -36,15 +38,16 @@ impl<'a> QueryContextBuilder<'a> {
     pub fn visit_table_expr(
         mut self,
         table_expr: &[Box<TableExpression>],
-        default_schema: Identifier,
+        default_schema: Ident,
```
JayWhite2357, line 41: Since this is immediately getting converted, let's leave this as Identifier, so we don't have to use `expect`.

Instead, do the conversion inside QueryExpr::try_new and return an Err instead of panicking.

```suggestion
        default_schema: Identifier,
```

crates/proof-of-sql/src/sql/postprocessing/group_by_postprocessing.rs
```diff
@@ -70,20 +71,26 @@ fn get_free_identifiers_from_expr(expr: &Expression) -> IndexSet<Identifier> {
...
     match expr {
         Expression::Column(_) | Expression::Literal(_) | Expression::Wildcard => expr,
         Expression::Aggregation { op, expr } => {
             let key = (op, (*expr));
-            if aggregation_expr_map.contains_key(&key) {
-                Expression::Column(*aggregation_expr_map.get(&key).unwrap())
+            if let Some(ident) = aggregation_expr_map.get(&key) {
+                Expression::Column(
+                    Identifier::try_from(ident.clone())
+                        .expect("Failed to convert new Ident to Identifier"),
```
JayWhite2357, line 83: Please return an Err instead of panicking here. Bubble that error up.

crates/proof-of-sql/src/sql/postprocessing/group_by_postprocessing.rs
```diff
@@ -70,20 +71,26 @@ fn get_free_identifiers_from_expr(expr: &Expression) -> IndexSet<Identifier> {
...
-                    .parse()
-                    .unwrap();
-                aggregation_expr_map.insert(key, new_col_id);
-                Expression::Column(new_col_id)
+                let new_ident = Ident {
+                    value: format!("__col_agg_{}", aggregation_expr_map.len()),
+                    quote_style: None,
+                };
+                let new_identifier = Identifier::try_from(new_ident.clone())
+                    .expect("Failed to convert new Ident to Identifier");
```
JayWhite2357, line 91: Please return an Err instead of panicking here. Bubble that error up.

JayWhite2357 (changes requested):
Thanks for this PR!
Two requests, and then I'll merge it.
1) Don't normalize the identifier names. Leave them as is.
2) Don't panic when converting Ident to Identifier. Instead return an Error. Not every valid Ident is a valid Identifier.

crates/proof-of-sql/src/sql/parse/query_context_builder.rs
```diff
@@ -110,17 +107,21 @@ impl<'a> QueryContextBuilder<'a> {
...
     }
 
+    /// # Panics
+    /// This function will panic if the conversion from `Ident` to `Identifier` fails.
     fn visit_select_all_expr(&mut self) -> ConversionResult<()> {
         for (column_name, _) in self.lookup_schema() {
-            let col_expr = Expression::Column(column_name);
-            self.visit_aliased_expr(AliasedResultExpr::new(col_expr, column_name))?;
+            let column_identifier =
+                Identifier::try_from(column_name).expect("Failed to convert Ident to Identifier");
```
JayWhite2357, line 122: This is still potentially panicking, rather than returning a `Result`.

iajoiner:
@JayWhite2357 In terms of WASM sizes I think we can proceed with the changes.

JayWhite2357 (approved):

iajoiner (approved):
@varshith257 Thanks again for your hard work!

iajoiner (approved):

PR #385, opened 2024-11-20
feat: add union utils

iajoiner:

# Rationale for this change
In order to support `UnionExec` we need to add union utils and have them tested.

- rename `Column::as_scalar` to `Column::into_scalar` since `as_` methods such as `as_boolean` serve different purposes
- add `as_` method for every enum variant in `Column`
- add `union_util.rs` which contains column and table unioning.

# Are these changes tested?

Will be.

### DIFF
--- a/crates/proof-of-sql/src/base/database/union_util.rs
+++ b/crates/proof-of-sql/src/base/database/union_util.rs
@@ -20,6 +20,7 @@ fn are_schemas_compatible(left: &[ColumnField], right: &[ColumnField]) -> bool {
 ///
 /// # Panics
 /// This function should never panic as long as it is written correctly
+#[allow(clippy::too_many_lines)]
 pub fn column_union<'a, S: Scalar>(
     columns: &[&Column<'a, S>],
     alloc: &'a Bump,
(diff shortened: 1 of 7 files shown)


crates/proof-of-sql/src/base/database/union_util.rs
```diff
@@ -0,0 +1,337 @@
...
+            correct_type: column_type,
+        });
+    }
+    Ok(match column_type {
+        ColumnType::Boolean => {
+            let result: Vec<_> = columns
+                .iter()
+                .flat_map(|col| col.as_boolean().expect("Column types should match"))
+                .copied()
+                .collect();
```
JayWhite2357, line 45: Should be no need to collect here.
Use `alloc_slice_fill_iter` or look at the source of `alloc_slice_fill_iter` (i.e. use `alloc_slice_fill_with`) instead.

crates/proof-of-sql/src/sql/proof_plans/projection_exec.rs
```diff
@@ -135,7 +135,7 @@ impl ProverEvaluate for ProjectionExec {
         .expect("Failed to create table from iterator");
         // 2. Produce MLEs
         res.inner_table().values().for_each(|column| {
-            builder.produce_intermediate_mle(column.as_scalar(alloc));
+            builder.produce_intermediate_mle(column.to_scalar(alloc));
```
JayWhite2357, line 138: Let's just delete `as_scalar`/`to_scalar`. I don't like it anyway.
```suggestion
            builder.produce_intermediate_mle(column));
```

JayWhite2357 (changes requested):

JayWhite2357 (approved):

PR #386, opened 2024-11-20
fix: remove blitzar as a dev-dependency

### LINKED ISSUE #383: Is it possible to add support for the ARM architecture?
Could you consider making GPU compilation optimizations for the Mac system? Currently, it is not possible to compile according to the work instructions on non-Linux machines and those without GPU resources. Good luck.

tlovell-sxt:
Closes #383 

# Rationale for this change

Tests cannot currently be run on alternative architectures because they
enable the blitzar dependency.
However, this is not strictly necessary, because all tests using blitzar
are behind the blitzar feature flag.
So, this removes blitzar as a dev dependency, as well as making a couple related changes to the README and CI.

- fix: remove blitzar as a dev-dependency
- docs: simplify README cargo test command excluding blitzar feature
- ci: check all targets in feature-specific checks of Check package job

# Are these changes tested?

These changes do not change existing functionality, which is verified by existing tests.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -45,19 +45,33 @@ jobs:
       - name: Run cargo check (proof-of-sql) (no features)
         run: cargo check -p proof-of-sql --no-default-features
       - name: Run cargo check (proof-of-sql) (all features)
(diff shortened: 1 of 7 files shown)


jacobtrombetta (approved):
LGTM

PR #396, opened 2024-11-27
feat: add `column_index_operation`

iajoiner:

# Rationale for this change
In sort-merge joins we need to be able to apply columns to possibly repetitive indexes. Hence let's add this before actually adding `sort_merge_join`.

- add `ColumnOperationError::IndexOutOfBounds`.
- add `apply_slice_to_indexes` in `slice_operation.rs`.
- add `IndexOp` and `ApplyIndexOp` to apply the new op to `Column`s.

# Are these changes tested?

Yes.

### DIFF
--- a/crates/proof-of-sql/src/base/database/column_index_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_index_operation.rs
@@ -1,0 +1,150 @@
+use super::{slice_operation::apply_slice_to_indexes, Column, ColumnOperationResult, ColumnType};
+use crate::base::scalar::Scalar;
+use bumpalo::Bump;
+
+/// Apply a `Column` to a vector of indexes, returning a new `Column` with the
+/// values at the given indexes. Repetitions are allowed.
+///
+/// # Panics
+/// Panics if any of the indexes are out of bounds.
+#[allow(dead_code)]
+pub(crate) fn apply_column_to_indexes<'a, S>(
+    column: &Column<'a, S>,
+    alloc: &'a Bump,
+    indexes: &[usize],
+) -> ColumnOperationResult<Column<'a, S>>
+where
+    S: Scalar,
+{
+    match column.column_type() {
+        ColumnType::Boolean => {
+            let raw_values = apply_slice_to_indexes(
+                column.as_boolean().expect("Column types should match"),
+                indexes,
+            )?;
+            Ok(Column::Boolean(alloc.alloc_slice_copy(&raw_values) as &[_]))
+        }
+        ColumnType::TinyInt => {
+            let raw_values = apply_slice_to_indexes(
+                column.as_tinyint().expect("Column types should match"),
+                indexes,
+            )?;
+            Ok(Column::TinyInt(alloc.alloc_slice_copy(&raw_values) as &[_]))
+        }
+        ColumnType::SmallInt => {
+            let raw_values = apply_slice_to_indexes(
+                column.as_smallint().expect("Column types should match"),
+                indexes,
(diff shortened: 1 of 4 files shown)


crates/proof-of-sql/src/base/database/column_index_operation.rs
```diff
@@ -0,0 +1,157 @@
+use super::{slice_operation::apply_slice_to_indexes, Column, ColumnOperationResult, ColumnType};
+use crate::base::scalar::Scalar;
+use alloc::vec::Vec;
+use bumpalo::Bump;
+
+#[allow(dead_code)]
+pub trait IndexOp {
```
tlovell-sxt, line 7: Could use some docs.

tlovell-sxt, line 7: I'm curious what other use cases this abstraction will have?

I see that implementing applying an index to a column is aided by using a generic function that does the same on a generic slice, but couldn't we just have a method on `Column` like `fn apply_index(self, indixes &[usize], alloc: &'a Bump) -> Column`?

I see that this abstraction allows you to essentially also apply some generic function to the column you're applying the index to in the process, but that potential is unused in this PR and could be done by applying a more basic abstraction for writing generic-over-column-type functions like after applying the index. Something like `trait GenericOverColumnTypeFn` that has a `fn call<T: PrimitiveColumnType>(value: T) -> T` and a method on `Column` like `fn generic_map(f: GenericOverColumnTypeFn, alloc: &'a Bump) -> Column`

Anyway, if this "writing a function for generic slices that takes indexes" has more use cases than just applying an index, or maybe applying an index and a function, I'd be interested to hear them.

crates/proof-of-sql/src/base/database/column_index_operation.rs
```diff
@@ -0,0 +1,157 @@
...
+                    tu,
+                    tz,
+                    alloc.alloc_slice_copy(&raw_values) as &[_],
+                ))
+            }
+        }
+    }
+}
+
+pub struct ApplyIndexOp {}
```
tlovell-sxt, line 103: Why not use a unit struct?
```suggestion
pub struct ApplyIndexOp;
```

tlovell-sxt, line 103: This could use docs too

crates/proof-of-sql/src/base/database/column_index_operation.rs
iajoiner: You are right that in this case we don't really have other use cases.

crates/proof-of-sql/src/base/database/column_index_operation.rs
```diff
@@ -0,0 +1,150 @@
...
+/// values at the given indexes. Repetitions are allowed.
+///
+/// # Panics
+/// Panics if any of the indexes are out of bounds.
+#[allow(dead_code)]
+pub(crate) fn apply_column_to_indexes<'a, S>(
+    column: &Column<'a, S>,
+    alloc: &'a Bump,
+    indexes: &[usize],
+) -> ColumnOperationResult<Column<'a, S>>
```
tlovell-sxt, line 15: NIT: This comment might not be useful depending on how this function will be used later. Without that context, I'd say it would be better if the error type was just a new `IndexOutOfBounds` error type since most of the column operation errors cannot occur in this function. But if this function is only ever used in contexts that are dealing with that error type, then it's probably fine.

tlovell-sxt (approved):

PR #399, opened 2024-12-03
chore: add conversion from `Table` to `OwnedTable`.

JayWhite2357:
# Rationale for this change

Conversions from `Table` to `OwnedTable` will be useful in future PRs.

`From<&Table>` and `From<Table>` are implemented for `OwnedTable`.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/base/database/owned_table.rs
+++ b/crates/proof-of-sql/src/base/database/owned_table.rs
@@ -1,4 +1,4 @@
-use super::OwnedColumn;
+use super::{OwnedColumn, Table};
 use crate::base::{map::IndexMap, polynomial::compute_evaluation_vector, scalar::Scalar};
 use alloc::{vec, vec::Vec};
 use proof_of_sql_parser::Identifier;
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql/src/base/database/owned_table.rs
```diff
@@ -106,3 +106,102 @@ impl<S: Scalar> core::ops::Index<&str> for OwnedTable<S> {
...
 }
+
+impl<'a, S: Scalar> From<&Table<'a, S>> for OwnedTable<S> {
+    fn from(value: &Table<'a, S>) -> Self {
+        match OwnedTable::try_from_iter(
+            value
+                .inner_table()
+                .iter()
+                .map(|(name, column)| (*name, OwnedColumn::from(column))),
+        ) {
```
iajoiner, line 117: I'd probably simply expect here because there is no legit reason for an error.

crates/proof-of-sql/src/base/database/owned_table.rs
```diff
@@ -106,3 +106,102 @@ impl<S: Scalar> core::ops::Index<&str> for OwnedTable<S> {
...
+            Err(OwnedTableError::ColumnLengthMismatch) => {
+                panic!("Tables should not have columns with differing lengths")
+            }
+        }
+    }
+}
+
+impl<'a, S: Scalar> From<Table<'a, S>> for OwnedTable<S> {
+    fn from(value: Table<'a, S>) -> Self {
+        match OwnedTable::try_from_iter(
```
iajoiner, line 128: We can just reuse the conversion above.

crates/proof-of-sql/src/base/database/owned_table.rs
```diff
@@ -106,3 +106,102 @@ impl<S: Scalar> core::ops::Index<&str> for OwnedTable<S> {
...
+    use super::OwnedTable;
+    use crate::base::{
+        database::{owned_table_utility::*, table_utility::*},
+        scalar::test_scalar::TestScalar,
+    };
+    use bumpalo::Bump;
+    use proof_of_sql_parser::posql_time::{PoSQLTimeUnit, PoSQLTimeZone};
+
+    #[test]
+    fn test_conversion_from_table_to_owned_table() {
```
iajoiner, line 153: I'd add some examples for tables with no rows and tables with no columns

crates/proof-of-sql/src/base/database/owned_table.rs
JayWhite2357: This implementation moves rather than copies the column names. The other one would need to copy (clone when we use Ident)

iajoiner (approved):
Really thanks for this!

PR #400, opened 2024-12-03
bench: add more queries to jaeger benches

jacobtrombetta:
# Rationale for this change
The Jaeger benchmarks only measure a small number of queries with a limited amount of column types.. This PR expands the benchmarks to include more queries over most column types.

The `Decimal75` and `TimestampTZ` column types are still not supported in the benchmarks `random_util` module. They are not used in the new queries.

- The `random_util` module is expanded to support most column types.
- Additional queries are added which measure supported SQL functions inside the Jaeger benchmarks.
- The arithmetic query is updated to produce a smaller output table.

# Are these changes tested?
Yes.

### DIFF
--- a/crates/proof-of-sql/benches/scaffold/querys.rs
+++ b/crates/proof-of-sql/benches/scaffold/querys.rs
@@ -169,5 +169,5 @@ pub const QUERIES: &[(&str, &str, &[(&str, ColumnType, OptionalRandBound)])] = &
         COMPLEX_CONDITION_TITLE,
         COMPLEX_CONDITION_SQL,
         COMPLEX_CONDITION_COLUMNS,
-    ),    
+    ),
 ];
--- a/crates/proof-of-sql/benches/scaffold/querys.rs
+++ b/crates/proof-of-sql/benches/scaffold/querys.rs
@@ -86,6 +86,38 @@ const BOOLEAN_FILTER_COLUMNS: &[(&str, ColumnType, OptionalRandBound)] = &[
     ("b", ColumnType::VarChar, None),
     ("c", ColumnType::Boolean, None),
 ];
+const LARGE_COLUMN_SET_TITLE: &str = "Large Column Set";
+const LARGE_COLUMN_SET_SQL: &str = "SELECT * FROM table WHERE b = d";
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql/benches/scaffold/querys.rs
```diff
@@ -60,6 +60,21 @@ const GROUPBY_COLUMNS: &[(&str, ColumnType, OptionalRandBound)] = &[
     ),
     ("c", ColumnType::Boolean, None),
 ];
+const AGGREGATE_TITLE: &str = "Aggregate";
+const AGGREGATE_SQL: &str = "SELECT SUM(a) FROM table WHERE b = a OR c = 'ab'";
```
stuarttimwhite, line 64: NIT: Can the string be anything other than "ab", just for readability, given that there are variables a and b? Really not a big deal.

crates/proof-of-sql/benches/scaffold/querys.rs
```diff
@@ -118,6 +118,27 @@ const LARGE_COLUMN_SET_COLUMNS: &[(&str, ColumnType, OptionalRandBound)] = &[
     ("g", ColumnType::VarChar, None),
     ("h", ColumnType::Scalar, None),
 ];
+const COMPLEX_CONDITION_TITLE: &str = "Complex Condition";
+const COMPLEX_CONDITION_SQL: &str =
+    "SELECT * FROM table WHERE (a > c * c AND b < c + 10) OR (d = 'abc')";
```
stuarttimwhite, line 123: NIT: Same thing as before.

crates/proof-of-sql/benches/scaffold/querys.rs
jacobtrombetta: The motivation for `ab` and not something like `a` was to limit the size of the result table. Maybe there is a better way to limit the result table size?

crates/proof-of-sql/benches/scaffold/querys.rs
jacobtrombetta: `abc` follows the same motivation as the last comment.

PR #408, opened 2024-12-04
fix!: normalize posqltimezone

JayWhite2357:
# Rationale for this change

The `PoSQLTimeZone` type has multiple representations for the same value, leading to complex upstream usage.

See the commit messages.

# Are these changes tested?
Yes, by existing, refactored, tests.

### DIFF
--- a/crates/proof-of-sql/src/sql/parse/query_context_builder.rs
+++ b/crates/proof-of-sql/src/sql/parse/query_context_builder.rs
@@ -272,10 +272,20 @@ impl<'a> QueryContextBuilder<'a> {
     }
 }
 
-/// TODO: add docs
+/// Checks if the binary operation between the left and right data types is valid.
+///
+/// # Arguments
+///
+/// * `left_dtype` - The data type of the left operand.
+/// * `right_dtype` - The data type of the right operand.
+/// * `binary_operator` - The binary operator to be applied.
+///
+/// # Returns
+///
+/// * `true` if the operation is valid, `false` otherwise.
 pub(crate) fn type_check_binary_operation(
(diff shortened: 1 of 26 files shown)


crates/proof-of-sql/src/sql/parse/query_context_builder.rs
```diff
@@ -274,8 +274,8 @@ impl<'a> QueryContextBuilder<'a> {
 
 /// TODO: add docs
```
Dustin-Ray, line 275: NIT: While we're here in this file, a doc comment might be nice

crates/proof-of-sql/src/sql/parse/query_context_builder.rs
```diff
@@ -274,8 +274,8 @@ impl<'a> QueryContextBuilder<'a> {
...
 pub(crate) fn type_check_binary_operation(
-    left_dtype: &ColumnType,
-    right_dtype: &ColumnType,
```
Dustin-Ray, line 278: NIT: Its not clear that these changes are related to the timestamp changes. This might be a better change in a separate PR.

JayWhite2357: The changes to timestamp made `ColumnType` a smaller type, which made clippy request these changes.

Dustin-Ray: Ok, feel free to resolve this NIT then.

Dustin-Ray:
LGTM - Thank you

Dustin-Ray (approved):
LGTM

PR #412, opened 2024-12-05
bench: initial add memory tracing to jaeger benches

jacobtrombetta:
# Rationale for this change
Initial memory logging has been added to Jaeger benchmark tracing. Memory logging does add execution time to the overall benchmark, so it is placed at `TRACE` level and requires benchmarks to be run with the environment variable `RUST_LOG=trace` to execute.

The logging function does not impact performance when code runs at `DEBUG` level or above.

This benchmark is prior to the change 
image
vs with the memory logging
image

- A `log` module is added to the `utils` crate and the function `log_memory_usage` is implemented.
- Memory logging is added to instrumented sections of code.
- Memory logging is placed at the `TRACE` level, requiring benchmarks to be run as `RUST_LOG=trace` in order to be viewed in the Jaeger UI.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_gpu.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_gpu.rs
@@ -9,17 +9,13 @@ use crate::{
 use blitzar::compute::ElementP2;
 #[cfg(feature = "rayon")]
 use rayon::iter::{IntoParallelIterator, ParallelIterator};
+#[cfg(feature = "log-memory-usage")]
 use sysinfo::{System, SystemExt};
-use tracing::{debug, span, Level};
+#[cfg(feature = "log-memory-usage")]
+use tracing::debug;
+use tracing::{span, Level};
 
-fn log_memory_start_usage() {
-    log_memory_usage("Start");
-}
-
-fn log_memory_end_usage() {
-    log_memory_usage("End");
-}
-
+#[cfg(feature = "log-memory-usage")]
 #[allow(clippy::cast_precision_loss)]
 fn log_memory_usage(name: &str) {
     if tracing::level_enabled!(Level::DEBUG) {
@@ -126,7 +122,8 @@ pub(super) fn compute_dynamic_dory_commitments(
         });
     span.exit();
 
-    log_memory_end_usage();
+    #[cfg(feature = "log-memory-usage")]
+    log_memory_usage("End");
 
     ddc
 }
--- a/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_gpu.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_gpu.rs
@@ -38,7 +38,7 @@ pub(super) fn compute_dynamic_dory_commitments(
     setup: &ProverSetup,
 ) -> Vec<DynamicDoryCommitment> {
     log::log_memory_usage("Start");
-    
+
     if committable_columns.is_empty() {
         return vec![];
     }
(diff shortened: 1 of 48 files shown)


crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_gpu.rs
```diff
@@ -27,13 +45,18 @@ use tracing::{span, Level};
...
-    skip_all
+    skip_all,
+    fields(total_memory, used_memory)
 )]
 pub(super) fn compute_dynamic_dory_commitments(
     committable_columns: &[CommittableColumn],
     offset: usize,
     setup: &ProverSetup,
 ) -> Vec<DynamicDoryCommitment> {
+    if tracing::level_enabled!(Level::DEBUG) {
```
JayWhite2357, line 56: Would be more compact to put this if within `log_memory_usage()`.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_gpu.rs
```diff
@@ -27,13 +45,18 @@ use tracing::{span, Level};
 #[tracing::instrument(
     name = "compute_dynamic_dory_commitments (gpu)",
     level = "debug",
-    skip_all
+    skip_all,
+    fields(total_memory, used_memory)
```
JayWhite2357, line 49: Percent used was a nice metric to have.
Can we have `start_` and `end_` versions of these as well?
In other words, put a `log_memory_start_usage()` and `log_memory_end_usage()`.

That way we can see the changed that occurred within a single span just by looking at that span.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_gpu.rs
jacobtrombetta: Agreed. Moved.

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_gpu.rs
jacobtrombetta: Updated, but I'm not sure if it fits with what you were thinking?

JayWhite2357: Does this work? Does it show all 6 things in the span inside jaeger?

I was expecting `fields(available_memory_start, used_memory_start, percentage_memory_used_start, available_memory_end, used_memory_end, percentage_memory_used_end)` or something. But, if this works, I like this better.

jacobtrombetta: Here is what it looks like with just `fields(available_memory, used_memory, percentage_memory_used)`, from [this commit](https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/412/commits/5a5c190376f228d2866f7bc87b9e4450eb9c2d34). image

crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_commitment_helper_gpu.rs
```diff
@@ -7,7 +7,28 @@ use crate::base::{commitment::CommittableColumn, if_rayon, slice_ops::slice_cast
...
+        let used_memory = system.used_memory() as f64 / 1024.0;
+        let percentage_memory_used = (used_memory / (used_memory + available_memory)) * 100.0;
+
+        trace!(
+            "{} Available memory: {:.2} MB, Used memory: {:.2} MB, Percentage memory used: {:.2}%",
+            name,
+            available_memory,
+            used_memory,
+            percentage_memory_used
+        );
```
JayWhite2357, line 29: I'm a bit confused. The screenshot looks correct, but the code looks wrong. The `sysinfo` documentation states:
> Returns the amount of available RAM in bytes.

Dividing by 1024 gives KiB, not MiB. So, I'm thinking that the number on the screenshot is KiB, not MB.

But, if that's true, the screenshot implies that there is only 317440.51 KiB = 30 MiB total memory.

JayWhite2357 (approved):

PR #413, opened 2024-12-06
feat: efficient range check

Dustin-Ray:
# Rationale for this change

Currently, ranges on scalar values are verified in a bitwise decomposition for each bit. This requires a proportionally large amount of work to verify. A more efficient range check can leverage a lookup table and protocol with a verifying party to efficiently argue range and uniqueness against columns of scalar values.

Adds an unsound, efficient, byte-oriented range check protocol via lookup table with logarithmic derivative arguments

# Are these changes tested?
- [x] A demo RangeCheckExpr based off a ProofPlan trait implementation so that the prover and verifier builder structs can be accessed

# Split

- Introduce threading to heavy matrix operations
- Maybe dont use ProofPlan as the test expression, ProofExpr might be better
- Handle panics? or introduce error handling
- Make the protocol sound


Dustin-Ray:
closing in favor of #414

PR #414, opened 2024-12-06
feat: unsound efficient range check

Dustin-Ray:
# Rationale for this change

Currently, ranges on scalar values are verified in a bitwise decomposition for each bit. This requires a proportionally large amount of work to verify, since each bit must be checked. A more efficient range check can leverage a lookup table and protocol with a verifying party to efficiently argue range and uniqueness against columns of scalar values.

Adds an unsound, efficient, byte-oriented range check protocol via lookup table with logarithmic derivative arguments

# Are these changes tested?
- [x] A demo RangeCheckExpr based off of a ProofPlan trait implementation so that the prover and verifier builder structs can be accessed

# TODO

- [ ] Introduce threading to heavy matrix operations
- [ ] Maybe dont use ProofPlan as the test expression, ProofExpr might be better
- [ ] Handle panics? or introduce error handling
- [ ] Make the protocol sound
- [ ] configurable, generic word sizes

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
@@ -1,13 +1,106 @@
-use crate::base::{scalar::Scalar, slice_ops};
-use alloc::vec::Vec;
+//! Implements a cryptographic range check using logarithmic derivatives to decompose a column of scalars
+//! into a matrix of words. This method leverages the properties of logarithmic derivatives to efficiently
+//! verify range proofs in a zero-knowledge setting by performing word-wise decompositions, intermediate MLEs,
+//! and modular inversions.
+//!
+//! The approach builds on the techniques outlined in the paper "Multivariate Lookups Based on Logarithmic
+//! Derivatives" [ePrint 2022/1530](https://eprint.iacr.org/2022/1530.pdf), which characterizes the use of
+//! logarithmic derivatives to perform multivariate lookups in cryptographic protocols.
+//!
+//! ## Key Steps:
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -23,44 +116,313 @@ fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
...
+    word_columns: &mut [&mut [u8]],
     alpha: S,
+    table_length: usize,
     inverted_word_columns: &mut [&mut [S]],
 ) {
-    // Iterate over each column
-    for (i, byte_column) in byte_columns.iter().enumerate() {
-        // Convert bytes to field elements and add alpha
-        let mut terms_to_invert: Vec<S> = byte_column.iter().map(|w| S::from(w) + alpha).collect();
+    for i in 0..31 {
```
Dustin-Ray, line 158: This shouldnt be hardcoded

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -23,44 +116,313 @@ fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
...
+    // Allocate row_sums from the bump allocator, ensuring it lives as long as 'a
+    let row_sums = alloc.alloc_slice_fill_with(scalars.len(), |_| S::ZERO);
+
+    // Iterate over each column and sum up the corresponding row values
+    for column in inverted_word_columns {
+        // Iterate over each scalar in the column
+        for (i, inv_word) in column.iter().enumerate() {
+            row_sums[i] += *inv_word;
+        }
+    }
```
Dustin-Ray, line 304: this can be optimized

PR #416, opened 2024-12-06
refactor!: make result serialization Solidity compatible

JayWhite2357:
# Rationale for this change

`postcard` does not port well to other language (e.g. Solidity). So, we need a simpler way to use the transcript.

See the individual commits.

* chore: add and implement `Transcript::extend_as_be_from_refs`
* chore: add `OwnedColumn::try_coerce_scalar_to_numeric`
* chore: add `OwnedTable::try_coerce_with_fields`
* chore: `impl From<TableCoercionError> for QueryError`
* chore: add `extend_transcript_with_owned_table` method
* refactor!: use `OwnedTable` to serialize onto transcript - This is the primary commit, and the motivation behind the others

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/proof/query_result.rs
+++ b/crates/proof-of-sql/src/sql/proof/query_result.rs
@@ -1,5 +1,5 @@
 use crate::base::{
-    database::{OwnedTable, OwnedTableError},
+    database::{ColumnCoercionError, OwnedTable, OwnedTableError, TableCoercionError},
     proof::ProofError,
     scalar::Scalar,
 };
@@ -37,6 +37,21 @@ pub enum QueryError {
     /// The number of columns in the table was invalid.
     #[snafu(display("Invalid number of columns"))]
     InvalidColumnCount,
+}
+
+impl From<TableCoercionError> for QueryError {
+    fn from(error: TableCoercionError) -> Self {
+        match error {
+            TableCoercionError::ColumnCoercionError {
+                source: ColumnCoercionError::Overflow,
+            } => QueryError::Overflow,
+            TableCoercionError::ColumnCoercionError {
+                source: ColumnCoercionError::InvalidTypeCoercion,
+            } => ProofError::InvalidTypeCoercion.into(),
+            TableCoercionError::NameMismatch => ProofError::FieldNamesMismatch.into(),
+            TableCoercionError::ColumnCountMismatch => ProofError::FieldCountMismatch.into(),
+        }
+    }
 }
 
 /// The verified results of a query along with metadata produced by verification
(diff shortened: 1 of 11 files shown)


crates/proof-of-sql/src/sql/proof/query_result.rs
```diff
@@ -39,6 +39,30 @@ pub enum QueryError {
...
+                source: ColumnCoercionError::InvalidTypeCoercion,
+            } => ProofError::VerificationError {
+                error: "Invalid Type Coercion",
+            }
+            .into(),
+            TableCoercionError::NameMismatch => ProofError::VerificationError {
+                error: "Field Names Mismatch",
+            }
+            .into(),
+            TableCoercionError::ColumnCountMismatch => ProofError::VerificationError {
```
iajoiner, line 58: This is evidence that we really should fix `ProofError` by making `TableCoercionError` an enum variant

crates/proof-of-sql/src/sql/proof/verifiable_query_result_test_utility.rs
```diff
@@ -33,20 +27,20 @@ pub fn exercise_verification(
...
 ) {
-    let verification_result = res.clone().verify(expr, accessor, &());
-    assert!(
-        verification_result.is_ok(),
-        "Verification failed: {:?}",
-        verification_result.err()
-    );
+    res.clone()
+        .verify(expr, accessor, &())
+        .expect("Verification failed");
```
iajoiner, line 32: I'd keep the assert above so that we print the error message which is its real purpose.

iajoiner (commented):
Really thanks for the PR! Please address the comments. Are we getting rid of custom serialization and `ProvableQueryResult` entirely in the near future?

crates/proof-of-sql/src/sql/proof/query_result.rs
JayWhite2357: In this case `TableCoercionError` shouldn't be a direct variant since some of the errors are verification errors, and some are not. We want a hard distinction between the two. So, even if add variants to `ProofError`, we still need a custom convert here.

JayWhite2357:
> Really thanks for the PR! Please address the comments. Are we getting rid of custom serialization and `ProvableQueryResult` entirely in the near future?

Perhaps. I'm not going to do that now since it's not blocking, though.

crates/proof-of-sql/src/sql/proof/verifiable_query_result_test_utility.rs
iajoiner: @JayWhite2357 Sure but the point is precisely to print more details about the error, not "Verification fails".

iajoiner (approved):

iajoiner (approved):

PR #417, opened 2024-12-06
feat: allow `FirstRoundBuilder` to produce MLEs

iajoiner:

# Rationale for this change
For range check and joins we need to have pre-challenge committing of columns. As a result we need to add such functionalities to `FirstRoundBuilder`.

- add functionalities that allow `FirstRoundBuilder` to produce MLEs.
- add `consume_first_round_mle``

# Are these changes tested?

Will be.


crates/proof-of-sql/src/sql/proof/query_proof.rs
```diff
@@ -268,7 +290,7 @@ impl<CP: CommitmentEvaluationProof> QueryProof<CP> {
         // add the commitments and bit disctibutions to the proof
```
Dustin-Ray, line 290:
```suggestion
        // add the commitments and bit distributions to the proof
```

JayWhite2357 (approved):

PR #421, opened 2024-12-09
feat: efficient range check

Dustin-Ray:
# Rationale for this change

Currently, ranges on scalar values are verified in a bitwise decomposition for each bit. This requires a proportionally large amount of work to verify, since each bit must be checked. A more efficient range check can leverage a lookup table and protocol with a verifying party to efficiently argue range and uniqueness against columns of scalar values.

Adds an efficient, byte-oriented range check protocol via lookup table with logarithmic derivative arguments

# Are these changes tested?
- [x] A demo RangeCheckExpr based off of a ProofPlan trait implementation so that the prover and verifier builder structs can be accessed

# Split

- Introduce threading to heavy matrix operations
- Configurable, generic word sizes
- Configurable range
- Dynamic column sizes
- Convert from ProofPlan to ProofExpr, need ProofExpr to be able to provide verifier challenge


crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -23,44 +116,315 @@ fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
...
+    alloc: &'a Bump,
+    scalars: &[S],
+    alpha: S,
+    table_length: usize,
+    builder: &mut FinalRoundBuilder<'a, S>,
+) {
+    // Allocate from 0 to 255 and pertrub with verifier challenge
+    let word_values_plus_alpha: &mut [S] =
+        alloc.alloc_slice_fill_with(max(256, scalars.len()), |i| S::from(&(i as u8)) + alpha);
+    builder.produce_intermediate_mle(word_values_plus_alpha as &[_]);
```
JayWhite2357, line 244: Since the prover is going to compute the commitment of this, it would be far more efficient to simply commit to the words, rather than the word+alpha. This is because the words are small (8 bits), but words+alpha are large (~250 bits).

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -23,44 +116,315 @@ fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
...
+        // Iterate over each scalar in the column
+        for (i, inv_word) in column.iter().enumerate() {
+            row_sums[i] += *inv_word;
+        }
+    }
 
-        // Invert all the terms in the column at once
-        slice_ops::batch_inversion(&mut terms_to_invert);
+    // Pass the row_sums reference with the correct lifetime to the builder
+    builder.produce_intermediate_mle(row_sums as &[_]);
```
JayWhite2357, line 307: We do not need to commit to this because we can derive the evaluation of the sum as the sum of the evaluations.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -23,44 +116,315 @@ fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
...
+            ),
+        ],
+    );
+}
 
-        // Assign the inverted values back to the inverted_word_columns
-        inverted_word_columns[i].copy_from_slice(&terms_to_invert);
+/// Verify that the prover claim is correct.
+#[allow(clippy::missing_panics_doc)]
+pub fn verifier_evaluate_range_check<C>(
```
JayWhite2357, line 337: There needs to be a check that
```
sum w_eval[i] * 2^i == scalars_eval
```

JayWhite2357 (changes requested):

crates/proof-of-sql/src/sql/proof_exprs/range_check_test_expr.rs
```diff
@@ -0,0 +1,125 @@
...
+    sql::proof::{
+        CountBuilder, FinalRoundBuilder, FirstRoundBuilder, ProofPlan, ProverEvaluate,
+        VerificationBuilder,
+    },
+};
+use bumpalo::Bump;
+use serde::Serialize;
+
+#[derive(Debug, Serialize)]
+pub struct RangeCheckTestExpr {
```
iajoiner, line 20: Since it is actually a `ProofPlan` you want to name it something like `RangeCheckTestPlan`

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -1,13 +1,106 @@
...
+#[allow(clippy::missing_panics_doc)]
+pub fn final_round_evaluate_range_check<'a, S: Scalar + 'a>(
+    builder: &mut FinalRoundBuilder<'a, S>,
+    scalars: &[S],
+    table_length: usize,
+    alloc: &'a Bump,
+) {
+    // Create 31 columns, each will collect the corresponding word from all scalars.
+    // 31 because a scalar will only ever have 248 bits of data set.
+    let mut word_columns: Vec<&mut [u8]> = (0..31)
```
iajoiner, line 46: What about using `repeat` instead?

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -1,13 +1,106 @@
...
+        .map(|_| alloc.alloc_slice_fill_with(scalars.len(), |_| 0))
+        .collect();
+
+    // Allocate space for the eventual inverted word columns.
+    let mut inverted_word_columns: Vec<&mut [S]> = (0..31)
```
iajoiner, line 51: Same

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -23,44 +116,315 @@ fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
...
-        let mut terms_to_invert: Vec<S> = byte_column.iter().map(|w| S::from(w) + alpha).collect();
+    for i in 0..31 {
+        let byte_column = &mut word_columns[i];
+
+        // Allocate words_plus_alpha
+        let words_plus_alpha: &mut [S] =
+            alloc.alloc_slice_fill_with(byte_column.len(), |j| S::from(&byte_column[j]) + alpha);
+
+        // Allocate words_plus_alpha_inv
+        let words_plus_alpha_inv: &mut [S] =
```
iajoiner, line 166: You can just copy the slice above.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
```diff
@@ -23,44 +116,315 @@ fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
...
+            alloc.alloc_slice_fill_with(byte_column.len(), |j| S::from(&byte_column[j]) + alpha);
+        slice_ops::batch_inversion(&mut words_plus_alpha_inv[..]);
```
iajoiner, line 168:
```rust
slice_ops::batch_inversion(words_plus_alpha_inv);
```

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
Dustin-Ray: This comment is unclear. I understand you are saying it is better to commit to the words vs words plus alpha, what is your suggestion here? Are you saying to not commit to word values at all? Commit to word values only and not values + alpha? (this causes a verification failure), Remove this line entirely?

To be clear, this is not a column of words as instantiated elsewhere in this code, this is a column of word values that the words take on, which in the case of byte sized words, are the values from 0..255. Originally we had this committed to as an anchored MLE because the verifier trivially has access to the values 0..255 if I remember correctly. Not sure if it makes sense to go back to that.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
Dustin-Ray: Does this increase the cost of verification?

crates/proof-of-sql/src/sql/proof_plans/range_check.rs
```diff
@@ -0,0 +1,698 @@
...
+pub fn final_round_evaluate_range_check<'a, S: Scalar + 'a>(
+    builder: &mut FinalRoundBuilder<'a, S>,
+    scalars: &[S],
+    table_length: usize,
+    alloc: &'a Bump,
+) {
+    // Create 31 columns, each will collect the corresponding word from all scalars.
+    // 31 because a scalar will only ever have 248 bits of data set.
+    let mut word_columns: Vec<&mut [u8]> = repeat(())
+        .take(31)
```
Dustin-Ray, line 47: All occurrences of this value will be replaced by a dynamic size that is determined from a chosen word size

Dustin-Ray:
> In general, reduce the usage of mutability. It makes reasoning difficult.

It would be helpful if you could please point to specific cases that you think should change. We are attempting to perform as many of the operations in place as possible, and this requires mutability.

crates/proof-of-sql/src/sql/proof_exprs/range_check.rs
Dustin-Ray: We understand we are trying to satisfy the following:

```SUM (I0+I1+I2+I3-(C*IN))=0```

and that the verifier has access to

```SUM (I0+I1+I2+I3) ```

but it doesnt appear that we currently have access to the ```(C*IN)``` part, maybe the prover supplies it, but its unclear how they might do that

File: crates/proof-of-sql/src/sql/proof_exprs/range_check.rs (hunk 3 above)
Comment (Dustin-Ray) [reply crates/proof-of-sql/src/sql/proof_exprs/range_check.rs]: So far I have this:

```rust
#[derive(Debug, Serialize)]
pub struct RangeCheckTestPlan {
    pub column: ColumnRef,
}

impl ProofPlan for RangeCheckTestPlan {
    #[doc = " Form components needed to verify and proof store into `VerificationBuilder`"]
    fn verifier_evaluate<S: Scalar>(
        &self,
        builder: &mut VerificationBuilder<S>,
        accessor: &IndexMap<ColumnRef, S>,
        _result: Option<&OwnedTable<S>>,
        one_eval_map: &IndexMap<TableRef, S>,
    ) -> Result<TableEvaluation<S>, ProofError> {
        let scalar_eval = accessor[&self.column];
        verifier_evaluate_range_check(builder, one_eval_map, scalar_eval);

        Ok(TableEvaluation::new(
            vec![accessor[&self.column]],
            one_eval_map[&self.column.table_ref()],
        ))
    }
```

So we have ```scalar_eval``` set to this: 

``` let scalar_eval = accessor[&self.column];``` 

What is ```sum w_eval[i]```?

Is it derived this way?

```rust
/// Verify that the prover claim is correct.
#[allow(clippy::missing_panics_doc)]
pub fn verifier_evaluate_range_check<C>(
    builder: &mut VerificationBuilder<'_, C>,
    one_eval_map: &IndexMap<TableRef, C>,
    scalar_eval: C,
) where
    C: Scalar,
{
    let alpha = builder.consume_post_result_challenge();
    let mut w_plus_alpha_inv_evals: Vec<_> = Vec::with_capacity(31);

    let mut sum = C::ZERO;

    // Consume the (wᵢⱼ)  and (wᵢⱼ + α)⁻¹ MLEs
    for i in 0..31 {
        let one_eval = one_eval_map
            .values()
            .next()
            .expect("one_eval_map should have at least one value");

        let w_eval = builder.consume_mle_evaluation();
        let words_plus_alpha_inv = builder.consume_mle_evaluation();

        // Store the evaluations of (wᵢⱼ + α)⁻¹
        w_plus_alpha_inv_evals.push(words_plus_alpha_inv);

        // Verify that:
        // (wᵢⱼ + α)⁻¹ * wᵢⱼ + α * (wᵢⱼ + α)⁻¹ - 1 = 0
        let word_eval = words_plus_alpha_inv * w_eval + (alpha * words_plus_alpha_inv);
        sum += word_eval * C::from((2_i32).pow(i));

        builder.produce_sumcheck_subpolynomial_evaluation(
            SumcheckSubpolynomialType::Identity,
            word_eval - *one_eval,
        );
    }

    assert_eq!(sum, scalar_eval);
```

crates/proof-of-sql/src/sql/proof_plans/range_check.rs
```diff
@@ -0,0 +1,699 @@
...
+        .iter_mut()
+        .map(|column| alloc.alloc_slice_fill_with(column.len(), |_| S::ZERO))
+        .collect();
+
+    // Initialize a vector to count occurrences of each byte (0-255).
+    // The vector has 256 elements padded with zeros to match the length of the word columns
+    // The size is the larger of 256 or the number of scalars.
+    let word_counts: &mut [i64] = alloc.alloc_slice_fill_with(max(256, scalars.len()), |_| 0);
+
+    decompose_scalar_to_words(scalars, &mut word_columns, word_counts);
```
iajoiner, line 59: Why not let `decompose_scalar_to_words` return `word_columns` and `word_counts` instead?

crates/proof-of-sql/src/sql/proof_plans/range_check.rs
```diff
@@ -0,0 +1,699 @@
...
+
+/// Get a count of the intermediate MLEs, post-result challenges, and subpolynomials.
+/// The counts depend directly on the word size and should be adjusted dynamically based
+/// on the word size.
+pub fn count_range_check(builder: &mut CountBuilder<'_>) {
+    builder.count_intermediate_mles(65);
+    builder.count_post_result_challenges(1);
+    builder.count_degree(3);
+    builder.count_subpolynomials(33);
+}
```
iajoiner, line 407: This should be delete

iajoiner (approved):

PR #424, opened 2024-12-10
refactor: make dynamic dory helper code generic

stuarttimwhite:

# Rationale for this change

There will be a new Hyrax commitment evaluation scheme which will use much existing common code from dynamic dory. This is preliminary work to make that code centralized.

- New directory in proof_primitive for common hyrax and dynamic dory code.
- Removal of dynamic dory helper code to this new directory

# Are these changes tested?
Existing dynamic dory tests are modified to use the generic code.


crates/proof-of-sql/src/proof_primitive/dory/dynamic_build_vmv_state.rs
```diff
@@ -1,7 +1,8 @@
 use super::{
-    dynamic_dory_helper::{compute_dynamic_v_vec, compute_dynamic_vecs},
-    DeferredGT, G1Affine, VMVProverState, VMVVerifierState, F,
+    dynamic_dory_helper::compute_dynamic_v_vec, DeferredGT, DoryScalar, G1Affine, VMVProverState,
+    VMVVerifierState, F,
 };
+use crate::proof_primitive::dynamic_dory_and_hyrax_common_utils::standard_basis_helper::compute_dynamic_vecs;
```
Dustin-Ray, line 5: NIT: Since we may have additional commitment schemes in the future, which might also end up sharing this code, it might be better to rename the utils module something more abstract like

```
commitment_scalar_common_utils
```
or
```
scalar_common_utils
```
or something else that makes sense to you.

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
```diff
@@ -1,18 +1,56 @@
...
 
-#[allow(dead_code)]
-/// This method produces evaluation vectors from a point. This is a helper method for generating a Vector-Matrix-Vector product in the dynamic dory scheme.
+/// This method produces evaluation vectors from a point. This is a helper method for generating a Vector-Matrix-Vector product in the dynamic dory and hyrax schemes.
 ///
 /// The ith element of the `lo_vec` is essentially the ith monomial basis element (lexicographically).
 /// The ith element of the `hi_vec` is essentially the jth monomial basis element where `j = row_start_index(i)`.
 ///
 /// NOTE: the `lo_vec` and `hi_vec` are scaled by `lo_vec[0]` and `hi_vec[0]` respectively.
 /// NOTE: `lo_vec` and `hi_vec` should otherwise consist entirely of zeros in order to ensure correct output.
```
Dustin-Ray, line 45: I think comments on non-pub functions are ```//```, not ```///```, but you may want to double check that

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
```diff
@@ -97,33 +138,35 @@ fn build_partial_second_half_standard_basis_vecs(
     }
 }
 
-#[allow(dead_code)]
-pub(super) fn fold_dynamic_standard_basis_tensors(
-    point: &[F],
-    alphas: &[F],
-    alpha_invs: &[F],
-) -> (F, F) {
+pub(crate) fn fold_dynamic_standard_basis_tensors<S: Scalar, I: Into<S> + From<S> + Copy>(
```
Dustin-Ray, line 141: This needs a doc comment please

crates/proof-of-sql/src/proof_primitive/mod.rs
```diff
@@ -1,4 +1,6 @@
 //! TODO: add docs
 pub mod dory;
+/// Central location for any code that is common to the hyrax and dynamic dory schemes.
+pub(super) mod dynamic_dory_and_hyrax_common_utils;
```
Dustin-Ray, line 4: NIT: See comment above about naming this module something more abstract to accommodate future commitment schemes

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
```diff
@@ -133,49 +176,80 @@ pub(super) fn fold_dynamic_standard_basis_tensors(
...
+                    compute_dynamic_standard_basis_vecs, fold_dynamic_standard_basis_tensors,
+                },
+                structure::{row_and_column_from_index, row_start_index},
+            },
+        },
+    };
+    use ark_bls12_381::Fr as F;
+    use ark_ff::AdditiveGroup;
+
+    pub fn naive_fold<S: std::ops::MulAssign + for<'a> std::ops::AddAssign<&'a mut S> + Copy>(
```
Dustin-Ray, line 207: Does this need to be pub? If so, please add a doc comment. If not, please change the visibility accordingly.

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
```diff
@@ -97,33 +138,35 @@ fn build_partial_second_half_standard_basis_vecs(
...
+    let alphas = alphas.iter().map(|i| (*i).into()).collect::<Vec<S>>();
+    let alpha_invs = alpha_invs.iter().map(|i| (*i).into()).collect::<Vec<S>>();
     let nu = point.len() / 2 + 1;
     debug_assert_eq!(alphas.len(), nu);
     debug_assert_eq!(alpha_invs.len(), nu);
     let lo_fold = if point.is_empty() {
         alphas[0]
     } else {
-        point.iter().zip(alphas).map(|(v, a)| v + a).product()
+        point.iter().zip(alphas).map(|(v, a)| *v + a).product()
```
Dustin-Ray, line 155: I think you can remove the clone on line 169 with 
```rust
.zip(alpha_invs.clone())
```

by making this change here:

```rust
point.iter().zip(&alphas).map(|(v, a)| *v + *a).product()
```

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
```diff
@@ -97,33 +138,35 @@ fn build_partial_second_half_standard_basis_vecs(
...
+                    (acc, S::ZERO)
                 } else if i == 1 {
-                    (acc * alpha_invs[1] + p * alpha_invs[0], F::ZERO)
+                    (acc * alpha_invs[1] + p * alpha_invs[0], S::ZERO)
                 } else if i % 2 == 0 {
                     let partial = (i / 2 + 1..i)
-                        .zip(alpha_invs)
+                        .zip(alpha_invs.clone())
                         .map(|(k, a)| point[k] + a)
                         .product();
```
Dustin-Ray, line 171: Then change this:
```rust
.zip(&alpha_invs)
.map(|(k, a)| point[k] + *a)
.product();
```

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
```diff
@@ -30,22 +68,25 @@ pub(super) fn compute_dynamic_standard_basis_vecs(point: &[F], lo_vec: &mut [F],
...
-    point: &[F],
-    lo_vec: &mut [F],
-    hi_vec: &mut [F],
+fn build_partial_second_half_standard_basis_vecs<
+    S: Scalar,
+    I: Into<S> + From<S> + Copy + Mul<Output = I>,
+>(
+    point: &[I],
+    lo_vec: &mut [I],
+    hi_vec: &mut [I],
```
Dustin-Ray, line 81: NIT: Its relatively easy to infer what F might be a symbol for, but not as much for I. It would be nice to maybe add a brief comment on this for clarity's sake.

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
Dustin-Ray: Also, we are changing a pub function to non pub, which I believe makes this PR a breaking change, so please update the pr title to refactor!

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/mod.rs
```diff
@@ -0,0 +1,4 @@
+/// Code for working with vectors that interface with the hyrax and dynamic dory matrix.
+pub mod standard_basis_helper;
+/// Code for working with the matrix that hyrax and dynamic dory share.
+pub mod structure;
```
Dustin-Ray, line 4: NIT: the naming of this could be improved, ```structure``` is quite generic. If the module has something to do with a shared matrix between hyrax and dory, you could consider naming it something like 
```
matrix_helper
```
or something else that might make sense to you

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
```diff
@@ -1,18 +1,56 @@
...
+        .skip(nu)
+        .map(|b| S::ONE - b)
+        .product::<S>()
+        .into();
+    let standard_basis_point: Vec<I> = point
+        .map(|b| I::from((S::ONE - b).inv().expect("Values in point cannot be 1.") - S::ONE))
+        .collect_vec();
+    compute_dynamic_standard_basis_vecs::<S, _>(&standard_basis_point, &mut lo_vec, &mut hi_vec);
+    (lo_vec, hi_vec)
+}
```
Dustin-Ray, line 37: Please mitigate these clones that this change has introduced:

```suggestion
pub(crate) fn compute_dynamic_vecs<
    S: Scalar,
    I: Into<S> + From<S> + Copy + core::ops::MulAssign + core::ops::Mul<Output = I>,
>(
    point: &[I],
) -> (Vec<I>, Vec<I>) {
    let point: Vec<S> = point.iter().map(|i| (*i).into()).collect();
    let nu = point.len() / 2 + 1;
    let mut lo_vec = vec![I::from(S::ZERO); 1 << nu];
    let mut hi_vec = vec![I::from(S::ZERO); 1 << nu];
    lo_vec[0] = point
        .iter()
        .take(nu)
        .map(|&b| S::ONE - b)
        .product::<S>()
        .into();
    hi_vec[0] = point
        .iter()
        .skip(nu)
        .map(|&b| S::ONE - b)
        .product::<S>()
        .into();
    let standard_basis_point: Vec<I> = point
        .iter()
        .map(|&b| I::from((S::ONE - b).inv().expect("Values in point cannot be 1.") - S::ONE))
        .collect();
    compute_dynamic_standard_basis_vecs::<S, _>(&standard_basis_point, &mut lo_vec, &mut hi_vec);
    (lo_vec, hi_vec)
}


```

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
Dustin-Ray: > Also, we are changing a pub function to non pub, which I believe makes this PR a breaking change, so please update the pr title to refactor!

- edit: disregard this comment, this does not affect the public api

crates/proof-of-sql/src/proof_primitive/dory/dynamic_build_vmv_state.rs
JayWhite2357: "dynamic" and "matrix" are the words that come to my mine

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/mod.rs
JayWhite2357: `matrix_structure`?

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
Dustin-Ray: Further, I'm not sure why you need the I generic here. I think you can remove it entirely from this function as well as the ```compute_dynamic_standard_basis_vecs``` function below. If you really really need it for something, please explain your reasoning, otherwise, i think you can safely drop it. This makes it so that you dont have an empty generic hanging around in your calls to this function, so this:

```rust
compute_dynamic_vecs::<DoryScalar, _>(&point)
compute_dynamic_standard_basis_vecs::<DoryScalar, _>(&point, &mut lo_vec, &mut hi_vec);
```

becomes this:

```rust
compute_dynamic_vecs::<DoryScalar>(&point)
compute_dynamic_standard_basis_vecs::<DoryScalar>(&point, &mut lo_vec, &mut hi_vec);
```

Which is easier to read, maintain, and use in practice. This change will break some of your tests, but all you need to do is remove the second generic from the function call as illustrated above, and then in tests where you have type mismatch errors from this change, you can do something like this:

```rust
    #[test]
    fn we_can_compute_dynamic_vecs_for_length_1_point() {
        let point = vec![F::from(2)];
        let expected_lo_vec = vec![F::from(1), F::from(0)];
        let expected_hi_vec = vec![F::from(1), F::from(1)];
        let point_converted: Vec<DoryScalar> = point.iter().map(|&x: &F| x.into()).collect();
        let (lo_vec, hi_vec) = compute_dynamic_vecs::<DoryScalar>(&point_converted);
        assert_eq!(
            expected_lo_vec
                .iter()
                .map(|&x| x.into())
                .collect::<Vec<DoryScalar>>(),
            lo_vec
        );
        assert_eq!(
            expected_hi_vec
                .iter()
                .map(|&x| x.into())
                .collect::<Vec<DoryScalar>>(),
            hi_vec
        );
    }
```

I would argue this approach is preferred because it is much more readable and explicit about what you are trying to accomplish. If you reallyyyyy need this second I parameter in these two functions, maybe for a future change somewhere else, please document this explicitly, otherwise I think we can safely drop it.

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
```diff
@@ -1,18 +1,56 @@
...
 /// NOTE: the `lo_vec` and `hi_vec` are scaled by `lo_vec[0]` and `hi_vec[0]` respectively.
 /// NOTE: `lo_vec` and `hi_vec` should otherwise consist entirely of zeros in order to ensure correct output.
-pub(super) fn compute_dynamic_standard_basis_vecs(point: &[F], lo_vec: &mut [F], hi_vec: &mut [F]) {
+fn compute_dynamic_standard_basis_vecs<
+    S: Scalar,
+    I: Into<S> + From<S> + Copy + core::ops::MulAssign + core::ops::Mul<Output = I>,
+>(
+    point: &[I],
+    lo_vec: &mut [I],
+    hi_vec: &mut [I],
```
JayWhite2357, line 52: I'm not sure why I did two mutable slices rather than returning two `Vec`s. Perhaps because `(Vec<_>, Vec<_>)` is ugly. Perhaps some other reason. But, mutating slices is also not great. I wonder if there is a better solution here.

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
stuarttimwhite: That's not working for me. Was that working for you locally?

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
stuarttimwhite: This is not a new function. I'm not really sure what the point of this is.

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
stuarttimwhite: That's a good point. None of these functions are externally public, and they get limited use outside unit tests.

crates/proof-of-sql/src/proof_primitive/dynamic_dory_and_hyrax_common_utils/standard_basis_helper.rs
stuarttimwhite: It seems like [all doc comments are meant to be ///](https://doc.rust-lang.org/reference/comments.html#doc-comments). This is a doc comment. I could delete it, since this function is no longer public, but I figured that would be a waste.

Dustin-Ray (approved):
LGTM - Thank you!

crates/proof-of-sql/src/proof_primitive/dory/dynamic_build_vmv_state.rs
```diff
@@ -11,7 +16,12 @@ pub(super) fn build_dynamic_vmv_prover_state(
     T_vec_prime: Vec<G1Affine>,
     nu: usize,
 ) -> VMVProverState {
-    let (lo_vec, hi_vec) = compute_dynamic_vecs(b_point);
+    let (lo_vec, hi_vec) =
+        compute_dynamic_vecs(&b_point.iter().copied().map(MontScalar).collect_vec());
```
JayWhite2357, line 20:
```suggestion
        compute_dynamic_vecs(bytemuck::TransparentWrapper::wrap_slice(b_point));
```

JayWhite2357 (approved):

PR #427, opened 2024-12-11
refactor!: drop counts from verifier

JayWhite2357:
# Rationale for this change

In order to simplify the verifier, we wish to only need to traverse the ProofPlan once. The `counts` method requires a second pass in addition to `verifier_evaluate`. This PR removes the need for this.

See the individual commits. At a high level
* The methods in `VerificationBuilder` are made fallible so that the verifier will catch incorrect proof sizes just in time.
* The use of counts is dropped.

# Are these changes tested?
Yes


crates/proof-of-sql/src/sql/proof_exprs/sign_expr_test.rs
```diff
@@ -33,20 +32,20 @@ fn prover_evaluation_generates_the_bit_distribution_of_a_negative_constant_colum
...
 
-#[test]
-fn count_fails_if_a_bit_distribution_is_out_of_range() {
-    let dists = [BitDistribution::new::<Curve25519Scalar, _>(&[
-        Curve25519Scalar::from(3) * Curve25519Scalar::from(u128::MAX),
-    ])];
-    let mut builder = CountBuilder::new(&dists);
-    assert!(count_sign(&mut builder).is_err());
-}
+// #[test]
```
iajoiner, line 35: Shall this test be removed?

JayWhite2357: Yes. I removed it and added another to replace it.

iajoiner (approved):

PR #432, opened 2024-12-13
feat: EVM compatible proof plan serialization 

JayWhite2357:
# Rationale for this change

We need Solidity to be able to read Proof Plans.

An initial implementation of serialization for Proof Plan is added.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/evm_compatibility/serialize_query_expr.rs
+++ b/crates/proof-of-sql/src/evm_compatibility/serialize_query_expr.rs
@@ -1,0 +1,158 @@
+use super::dyn_proof_plan_serializer::{DynProofPlanSerializer, ProofPlanSerializationError};
+use crate::{
+    base::scalar::Scalar,
+    sql::{parse::QueryExpr, proof::ProofPlan},
+};
+use alloc::vec::Vec;
+
+/// Serializes a `QueryExpr` into a vector of bytes.
+///
+/// This function takes a `QueryExpr` and attempts to serialize it into a vector of bytes.
+/// The serialization is done in a manner that is compatible with efficient proof verification
+/// on the EVM.
+///
+/// # Arguments
+///
+/// * `query_expr` - A reference to the `QueryExpr` to be serialized.
+///
+/// # Returns
+///
+/// * `Ok(Vec<u8>)` - A vector of bytes representing the serialized query expression.
+/// * `Err(ProofPlanSerializationError)` - An error indicating why the serialization failed.
+///
+/// # Errors
+///
+/// This function returns a `ProofPlanSerializationError::NotSupported` error if the query
+/// expression contains postprocessing steps or if the proof plan cannot be serialized.
+pub fn serialize_query_expr<S: Scalar>(
+    query_expr: &QueryExpr,
+) -> Result<Vec<u8>, ProofPlanSerializationError> {
+    let plan = query_expr
+        .postprocessing()
+        .is_empty()
+        .then(|| query_expr.proof_expr())
+        .ok_or(ProofPlanSerializationError::NotSupported)?;
+    let bytes = DynProofPlanSerializer::<S>::try_new(
+        plan.get_table_references(),
+        plan.get_column_references(),
+    )?
+    .serialize_dyn_proof_plan(plan)?
+    .into_bytes();
+    Ok(bytes)
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        base::{
+            database::{ColumnRef, ColumnType, LiteralValue},
+            map::indexset,
+            scalar::test_scalar::TestScalar,
+        },
+        evm_compatibility::{
+            dyn_proof_plan_serializer::{
+                constants::*, DynProofPlanSerializer, ProofPlanSerializationError,
+            },
+            serialize_query_expr,
+        },
+        sql::{
+            parse::QueryExpr,
+            postprocessing::{OwnedTablePostprocessing, SlicePostprocessing},
+            proof_exprs::{
+                AliasedDynProofExpr, ColumnExpr, DynProofExpr, EqualsExpr, LiteralExpr, TableExpr,
+            },
+            proof_plans::{DynProofPlan, EmptyExec, FilterExec},
+        },
+    };
+    use core::iter;
+    use itertools::Itertools;
+
+    #[test]
+    fn we_can_generate_serialized_proof_plan_for_query_expr() {
+        let table_ref = "namespace.table".parse().unwrap();
+        let identifier_alias = "alias".parse().unwrap();
+
+        let plan = DynProofPlan::Filter(FilterExec::new(
+            vec![AliasedDynProofExpr {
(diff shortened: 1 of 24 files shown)


crates/proof-of-sql/src/evm_compatibility/dyn_proof_plan_serializer.rs
```diff
@@ -0,0 +1,314 @@
...
+pub enum SerializationError {
+    #[snafu(display("Not supported"))]
+    /// A SQL expression that is not supported by the serialization
+    NotSupported,
+    #[snafu(display("Too many results"))]
+    /// The number of results in a filter expression is too large
+    TooManyResults,
+    #[snafu(display("Too many tables"))]
+    /// The number of tables in a plan is too large
+    TooManyTables,
```
iajoiner, line 38: I think we should put the exact limit in the error message or people won't even know how many are too many

crates/proof-of-sql/src/evm_compatibility/dyn_proof_plan_serializer.rs
```diff
@@ -0,0 +1,314 @@
...
+    #[snafu(display("Too many columns"))]
+    /// The number of columns in a plan is too large
+    TooManyColumns,
```
iajoiner, line 41: Same. It's actually plausible to have 256 columns in web2

iajoiner (commented):
Hmm are you sure we don't want to use an existing serialization library such as ethabi, rlp or ssz for this?

JayWhite2357:
> Hmm are you sure we don't want to use an existing serialization library such as ethabi, rlp or ssz for this?

I originally tried the abi, but found that this format is magnitudes cheaper when actually passing over the AST.

Although, I'm now considering changing this to postorder rather than preorder. The idea is that the data is in the exact order that we need it as we do the verification. Saves arithmetic and having to jump around. I'm not sure if it would work as well, though.

Not too familiar with rlp or ssz. I don't think they are Solidity/EVM formats, but are used by Ethereum for other things.

crates/proof-of-sql/src/evm_compatibility/serialize_query_expr.rs
```diff
@@ -0,0 +1,158 @@
...
+            .chain([FILTER_EXEC_NUM, 0, 1]) // filter expr, table number, result count
+            .chain([COLUMN_EXPR_NUM, 0]) // column expr, column a (#0)
+            .chain([EQUALS_EXPR_NUM]) // equals expr
+            .chain([COLUMN_EXPR_NUM, 1]) // column expr, column b (#1)
+            .chain([LITERAL_EXPR_NUM, BIGINT_TYPE_NUM]) // literal expr, literal type
+            .chain([0; 31]) // leading 0s of literal value
+            .chain([5]) // literal value
+            .collect_vec();
+        assert_eq!(bytes, expected_bytes);
+    }
```
JayWhite2357, line 157: This is the "integration test": the top level example of what is happening in this PR.

iajoiner (approved):

crates/proof-of-sql/src/evm_compatibility/serialize_query_expr.rs
waelsy123: let expected_bytes = iter::empty::<u8>()
            .chain([FILTER_EXEC_NUM, 0, 1]) // filter expr, table number, result count
            .chain([COLUMN_EXPR_NUM, 0]) // column expr, column a (#0)
            .chain([EQUALS_EXPR_NUM]) // equals expr
            .chain([LITERAL_EXPR_NUM, BIGINT_TYPE_NUM]) // literal expr, literal type
            .chain([0; 31]) // leading 0s of literal value
            .chain([5]) // literal value
            .collect_vec();
            
            
           @JayWhite2357  what would that be?

waelsy123 (dismissed):

crates/proof-of-sql/src/evm_compatibility/serialize_query_expr.rs
```diff
@@ -0,0 +1,158 @@
...
+            proof_plans::{DynProofPlan, EmptyExec, FilterExec},
+        },
+    };
+    use core::iter;
+    use itertools::Itertools;
+
+    #[test]
+    fn we_can_generate_serialized_proof_plan_for_query_expr() {
+        let table_ref = "namespace.table".parse().unwrap();
+        let identifier_alias = "alias".parse().unwrap();
```
waelsy123, line 73: what is `identifier_alias` used for?

crates/proof-of-sql/src/evm_compatibility/serialize_query_expr.rs
JayWhite2357: > could be a basic question but I understand we expect the filter expression to be `where a = 5` so we do we have `.chain([COLUMN_EXPR_NUM, 1]) // column expr, column b (#1)` in line 151 ?

I flipped the a and b in the comments. It should be the other way. I'll fix it.

JayWhite2357: > ```
>  let expected_bytes = iter::empty::<u8>()
>         .chain([FILTER_EXEC_NUM, 0, 1]) // filter expr, table number, result count
>         .chain([COLUMN_EXPR_NUM, 0]) // column expr, column a (#0)
>         .chain([EQUALS_EXPR_NUM]) // equals expr
>         .chain([LITERAL_EXPR_NUM, BIGINT_TYPE_NUM]) // literal expr, literal type
>         .chain([0; 31]) // leading 0s of literal value
>         .chain([5]) // literal value
>         .collect_vec();
>         
>         
>        @JayWhite2357  what would that be? 
> ```

This is invalid. Would be something like `SELECT a FROM table WHERE 5 = `

crates/proof-of-sql/src/evm_compatibility/serialize_query_expr.rs
JayWhite2357: Nothing at the moment. Eventually, it will need to be used to validate the result names, if desired. But, it isn't going to be used by the EVM proof verifier.

crates/proof-of-sql/src/evm_compatibility/serialize_query_expr.rs
waelsy123: so the serialized query expression is agnostic to column names. right? I see the query plan create incremental reference to column by their mentioning order

waelsy123: @JayWhite2357 how about `result count` does that refer to rows count? and needed for verification? because I don't see it needed within AST

iajoiner (approved):

crates/proof-of-sql/src/evm_compatibility/serialize_query_expr.rs
JayWhite2357: That refers to the number of results in the filter.
e.g., `SELECT a,b,c,d,e,f,g FROM table where a=5` would have 7 as the result count.

waelsy123 (approved):

PR #440, opened 2024-12-16
fix: make range check protocol sound by moving word value evaluation into sumcheck

Dustin-Ray:
# Rationale for this change

Currently, the prover in the range check protocol establishes a column of values from 0 to the max value that a word can be represented by, in the case of bytes, 256. This is unsound because the prover could send commitments to any values and the protocol would successfully verify, when it is crucial that the word values be the sequentially represented values that fit inside of the word (in the case of bytes, this is a column containing all values from 0..256). This PR moves this step of word value calculation into the actual sumcheck machinery, ensuring that the correct values are produced during protocol execution.

- [x] Add a new ```rho_256_evaluation``` field to the ```SumcheckMleEvaluations``` struct
- [x] Alter ```SumcheckMleEvaluations``` constructor to produce the ```rho_256_evaluation``` evaluations
- [x] Alter the range check verifier to consume the ```rho_256_evaluation```, assuming that the ```FirstRoundBuilder``` correctly executes the ```produce_one_evaluation_length``` function in a gadget or test expression.

# Are these changes tested?
Yes, by existing tests


crates/proof-of-sql/src/sql/proof_gadgets/range_check_test.rs
```diff
@@ -27,6 +27,7 @@ impl ProverEvaluate for RangeCheckTestPlan {
         table_map: &IndexMap<TableRef, Table<'a, S>>,
     ) -> Table<'a, S> {
         builder.request_post_result_challenges(1);
+        builder.produce_one_evaluation_length(256);
```
iajoiner, line 30: @Dustin-Ray Is this really just so that we update `range_length`? Looks like this is the right time to split the one evaluation length queue from the `range_length` updates. Do you want to do this or do you want me to?

iajoiner (commented):
Really thanks for the improvement. I'd love to see range checks used though.

crates/proof-of-sql/src/sql/proof_gadgets/range_check_test.rs
Dustin-Ray: @iajoiner if its a small change without much of a diff then I would be happy to do it in this PR, if its more involved I can definitely do it in another PR

Dustin-Ray:
> I'd love to see range checks used though.

Here are [complete round trip tests](https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/crates/proof-of-sql/src/sql/proof_gadgets/range_check_test.rs#L93-L159) on columns of scalars, with an assumed range of 0 <= x <= 2^248

iajoiner (approved):

PR #441, opened 2024-12-17
refactor!: identifier to ident and update docs

varshith257:

# Rationale for this change

To update docs to reflect the migration of Identifier -> Ident

- Updated docs in the proof-of-sql crate with Ident

# Are these changes tested?

Part of  #235

### DIFF
--- a/crates/proof-of-sql/README.md
+++ b/crates/proof-of-sql/README.md
@@ -90,7 +90,7 @@ Parsing Query... 1.870256ms
 Generating Proof... 467.45371ms
 Verifying Proof... 7.106864ms
 Valid proof!
-Query result: OwnedTable { table: {Identifier { name: "b" }: VarChar(["hello", "world"])} }
+Query result: OwnedTable { table: {Ident { value: "b", quote_style: None }: VarChar(["hello", "world"])} }
 ```
 
 For a detailed explanation of the example and its implementation, refer to the [README](https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/crates/proof-of-sql/examples/hello_world/README.md) and source code in [hello_world/main.rs](https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/crates/proof-of-sql/examples/hello_world/main.rs).
--- a/crates/proof-of-sql/benches/scaffold/random_util.rs
+++ b/crates/proof-of-sql/benches/scaffold/random_util.rs
@@ -10,7 +10,6 @@ pub type OptionalRandBound = Option<fn(usize) -> i64>;
 /// # Panics
 ///
 /// Will panic if:
-/// - The provided identifier cannot be parsed into an `Identifier` type.
 /// - An unsupported `ColumnType` is encountered, triggering a panic in the `todo!()` macro.
 #[allow(clippy::cast_sign_loss, clippy::cast_possible_truncation)]
 pub fn generate_random_columns<'a, S: Scalar>(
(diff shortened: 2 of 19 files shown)


crates/proof-of-sql/README.md
```diff
@@ -90,7 +90,7 @@ Parsing Query... 1.870256ms
 Generating Proof... 467.45371ms
 Verifying Proof... 7.106864ms
 Valid proof!
-Query result: OwnedTable { table: {Identifier { name: "b" }: VarChar(["hello", "world"])} }
+Query result: OwnedTable { table: {Ident { name: "b" }: VarChar(["hello", "world"])} }
```
iajoiner, line 93: @varshith257 Please do run this example and put the results here. I don't think this is simply a string replacement here.

crates/proof-of-sql/examples/hello_world/README.md
```diff
@@ -30,5 +30,5 @@ Parsing Query... 1.870256ms
 Generating Proof... 467.45371ms
 Verifying Proof... 7.106864ms
 Valid proof!
-Query result: OwnedTable { table: {Identifier { name: "b" }: VarChar(["hello", "world"])} }
+Query result: OwnedTable { table: {Ident { name: "b" }: VarChar(["hello", "world"])} }
```
iajoiner, line 33: Same

crates/proof-of-sql/src/base/database/owned_table_utility.rs
```diff
@@ -55,7 +55,7 @@ pub fn owned_table<S: Scalar>(
 /// ]);
 ///```
 /// # Panics
-/// - Panics if `name.parse()` fails to convert the name into an `Identifier`.
+/// - Panics if `name.parse()` fails to convert the name into an `Ident`.
```
iajoiner, line 58: This shouldn't panic any more. Also we can't run `name.parse()` any more either. Same for other cases.

crates/proof-of-sql/src/base/database/table_utility.rs
```diff
@@ -22,7 +22,7 @@ use bumpalo::Bump;
 use proof_of_sql_parser::posql_time::{PoSQLTimeUnit, PoSQLTimeZone};
 use sqlparser::ast::Ident;
 
-/// Creates an [`Table`] from a list of `(Identifier, Column)` pairs.
+/// Creates an [`Table`] from a list of `(Ident, Column)` pairs.
```
iajoiner, line 25: What I mentioned about `owned_table_utility.rs` applies here as well.

iajoiner (commented):
Really thanks for your PR!
General comments:
1. You don't need to remove the word "identifier" when it refers to the general concept of identifiers rather than `proof_of_sql_parser::Identifier`
2. .The comments on the two table util files need to be fixed in the entire files rather than in just the two instances I mentioned.

crates/proof-of-sql/benches/scaffold/random_util.rs
```diff
@@ -10,7 +10,7 @@ pub type OptionalRandBound = Option<fn(usize) -> i64>;
 /// # Panics
 ///
 /// Will panic if:
-/// - The provided identifier cannot be parsed into an `Identifier` type.
+/// - The provided identifier cannot be parsed into an `Ident` type.
```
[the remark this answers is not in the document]
iajoiner: Yes

crates/proof-of-sql/examples/hello_world/README.md
iajoiner: @varshith257 Well then please use this here and in the other example with such comments as well.

iajoiner (approved):

iajoiner (changes requested):
Please remove unnecessary panic docs from `owned_table_utility.rs` and `table_utility.rs`

iajoiner:
@varshith257 Which issues?

iajoiner (approved):
Thanks again for the PR!

JayWhite2357 (approved):

PR #443, opened 2024-12-17
fix: add sumcheck range length field

Dustin-Ray:

# Rationale for this change

# Are these changes tested?


crates/proof-of-sql/src/sql/proof/query_proof.rs
```diff
@@ -104,9 +104,9 @@ impl<CP: CommitmentEvaluationProof> QueryProof<CP> {
         let query_result = expr.first_round_evaluate(&mut first_round_builder, &alloc, &table_map);
         let owned_table_result = OwnedTable::from(&query_result);
         let provable_result = query_result.into();
-        let one_evaluation_lengths = first_round_builder.one_evaluation_lengths();
+        let range_evaluation_lengths = first_round_builder.sumcheck_range_lengths();
```
iajoiner, line 107: @Dustin-Ray No we need both one evaluation lengths and sumcheck range lengths but for different reasons
The former is for one evaluation and the latter is for computing `range_length`

PR #445, opened 2024-12-18
fix: separate ones eval length and range lengths 

Dustin-Ray:
# Rationale for this change

Currently, range length is determined from ```one_evaluation_lengths```, when this should actually be a value that is set upon construction of the first round builder, and can be updated if needed. This PR introduces support for a configurable range length for the first round builder.

- [x] adds ```range_length``` field to ```FirstRoundBuilder``` with constructor parameter and accessor function
- [x] updates usages of ```FirstRoundBuilder``` to start from a range of zero
- [x] updates range check to set the range using these changes

# Are these changes tested?
yes


crates/proof-of-sql/src/sql/proof/first_round_builder.rs
```diff
@@ -8,19 +8,34 @@ pub struct FirstRoundBuilder {
...
+    /// The range length used in the proof.
+    range_length: usize,
 }
 
 impl Default for FirstRoundBuilder {
     fn default() -> Self {
-        Self::new()
+        Self::new(0)
     }
 }
```
iajoiner, line 19: Actually it was only here because of Clippy. It doesn't need to.

crates/proof-of-sql/src/sql/proof/first_round_builder.rs
```diff
@@ -8,19 +8,34 @@ pub struct FirstRoundBuilder {
...
+            range_length: initial_range_length,
+        }
+    }
+
+    /// Get the range length used in the proof.
+    pub(crate) fn range_length(&self) -> usize {
+        self.range_length
+    }
+
+    /// Update the range length used in the proof.
```
iajoiner, line 35: Please note in the comment that we only change it if the input is larger

crates/proof-of-sql/src/sql/proof_plans/filter_exec_test.rs
```diff
@@ -192,7 +192,7 @@ fn we_can_get_an_empty_result_from_a_basic_filter_on_an_empty_table_using_first_
             ColumnType::Decimal75(Precision::new(75).unwrap(), 0),
         ),
     ];
-    let first_round_builder = &mut FirstRoundBuilder::new();
+    let first_round_builder = &mut FirstRoundBuilder::new(0);
```
iajoiner, line 195: You should start from the input table length instead of 0

crates/proof-of-sql/src/sql/proof_plans/projection_exec_test.rs
```diff
@@ -174,7 +174,7 @@ fn we_can_get_an_empty_result_from_a_basic_projection_on_an_empty_table_using_fi
             ColumnType::Decimal75(Precision::new(75).unwrap(), 0),
         ),
     ];
-    let first_round_builder = &mut FirstRoundBuilder::new();
+    let first_round_builder = &mut FirstRoundBuilder::new(0);
```
iajoiner, line 177: Same here

crates/proof-of-sql/src/sql/proof_plans/slice_exec_test.rs
```diff
@@ -102,7 +102,7 @@ fn we_can_get_an_empty_result_from_a_slice_on_an_empty_table_using_first_round_e
             ColumnType::Decimal75(Precision::new(75).unwrap(), 0),
         ),
     ];
-    let first_round_builder = &mut FirstRoundBuilder::new();
+    let first_round_builder = &mut FirstRoundBuilder::new(0);
     let res: OwnedTable<Curve25519Scalar> = ProvableQueryResult::from(expr.first_round_evaluate(
```
iajoiner, line 106: Same

crates/proof-of-sql/src/sql/proof_plans/union_exec_test.rs
```diff
@@ -272,7 +272,7 @@ fn we_can_get_result_from_union_using_first_round_evaluate() {
         ],
         fields.clone(),
     );
-    let first_round_builder = &mut FirstRoundBuilder::new();
+    let first_round_builder = &mut FirstRoundBuilder::new(0);
```
iajoiner, line 275: Take the max of input table lengths please

crates/proof-of-sql/src/sql/proof/first_round_builder.rs
```diff
@@ -8,19 +8,34 @@ pub struct FirstRoundBuilder {
     num_post_result_challenges: usize,
     /// The extra one evaluation lengths used in the proof.
     one_evaluation_lengths: Vec<usize>,
+    /// The range length used in the proof.
```
iajoiner, line 11: Please mention that it is the max of all the lengths used in sumcheck

crates/proof-of-sql/src/sql/proof_plans/filter_exec_test.rs
Dustin-Ray: these tables are empty so the lengths should be zero right

crates/proof-of-sql/src/sql/proof_plans/projection_exec_test.rs
Dustin-Ray: these tables are empty so the lengths should be zero right

crates/proof-of-sql/src/sql/proof_plans/slice_exec_test.rs
Dustin-Ray: these tables are empty so the lengths should be zero right

crates/proof-of-sql/src/sql/proof/first_round_builder.rs
```diff
@@ -8,19 +8,28 @@ pub struct FirstRoundBuilder {
     num_post_result_challenges: usize,
     /// The extra one evaluation lengths used in the proof.
     one_evaluation_lengths: Vec<usize>,
-}
-
-impl Default for FirstRoundBuilder {
-    fn default() -> Self {
-        Self::new()
-    }
+    /// The max of all `range_lengths` used in sumcheck.
```
iajoiner, line 11: Sorry let's just make it `// The `range_length` used in sumcheck which is max of all possible ones.`

crates/proof-of-sql/src/sql/proof_plans/union_exec.rs
```diff
@@ -120,6 +120,7 @@ impl ProverEvaluate for UnionExec {
         let res = table_union(&inputs, alloc, self.schema.clone()).expect("Failed to union tables");
         builder.request_post_result_challenges(2);
         builder.produce_one_evaluation_length(res.num_rows());
+        builder.update_range_length(res.num_rows());
```
JayWhite2357, line 123: Might be nice for `produce_one_evaluation_length` to internally/automatically call `update_range_length`. That way, it's more failsafe.

JayWhite2357 (commented):
Thanks! This is nice and clean.
I'll let @iajoiner handle final approval.

crates/proof-of-sql/src/sql/proof_plans/union_exec.rs
iajoiner: @JayWhite2357 `update_range_length` is really only used for the few situations where `range_length` can potentially increase, mostly multi-table nodes (joins and unions) as well as range checks. So after the split `one_eval_lengths` and `range_length` no longer have much to do with each other.

crates/proof-of-sql/src/sql/proof_plans/union_exec.rs
```diff
@@ -121,6 +121,7 @@ impl ProverEvaluate for UnionExec {
...
```
iajoiner, line 124: Repetition

iajoiner (approved):
@Dustin-Ray This is really neat!

PR #457, opened 2025-01-06
refactor: move range check word decomposition to first round builder

Dustin-Ray:
# Rationale for this change

Currently, the range check word decomposition is handled entirely in the ```FinalRound```. It is necessary to move this step into the ```FirstRound``` so that it is not possible for any prover to gain a malicious advantage by performing a decomposition while the  challenge alpha is known. Since alpha can only be provided in the ```FinalRound```, moving word decomposition and commitment into the first round effectively erases any possibility that the prover could cheat by gaining knowledge of the verifier challenge in advance.

- [x] move word decomposition into the FirstRound of the range check protocol
- [x] update commitment procedures in verifier accordingly

# Are these changes tested?
Existing testing framework is successfully updated

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
+++ b/crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
@@ -32,8 +32,32 @@ use bytemuck::cast_slice;
 use core::{cmp::max, iter::repeat};
 
 /// Update the max range length for the range check.
-pub fn first_round_evaluate_range_check<'a, S: Scalar + 'a>(builder: &mut FirstRoundBuilder<S>) {
+pub fn first_round_evaluate_range_check<'a, S: Scalar + 'a>(
+    builder: &mut FirstRoundBuilder<'a, S>,
+    scalars: &[S],
+    alloc: &'a Bump,
+) {
     builder.update_range_length(256);
+
+    // Create 31 columns, each will collect the corresponding word from all scalars.
+    // 31 because a scalar will only ever have 248 bits of data set.
+    let mut word_columns: Vec<&mut [u8]> = (0..31)
+        .map(|_| alloc.alloc_slice_fill_copy(scalars.len(), 0))
+        .collect();
+    // Initialize a vector to count occurrences of each byte (0-255).
+    // The vector has 256 elements padded with zeros to match the length of the word columns
+    // The size is the larger of 256 or the number of scalars.
+    let word_counts: &mut [i64] = alloc.alloc_slice_fill_copy(max(256, scalars.len()), 0);
+
+    decompose_scalar_to_words(scalars, &mut word_columns, word_counts);
+
+    for byte_column in &mut word_columns {
+        // Allocate words
+        let words = alloc.alloc_slice_fill_with(byte_column.len(), |j| S::from(&byte_column[j]));
+
+        // Produce an MLE over words
+        builder.produce_intermediate_mle(words as &[_]);
+    }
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
```diff
@@ -91,6 +117,82 @@ pub fn final_round_evaluate_range_check<'a, S: Scalar + 'a>(
     );
 }
 
+/// Verify that the prover claim is correct.
```
iajoiner, line 120: Please do not move the function or make the move a separate commit. Thanks!

crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
```diff
@@ -32,8 +32,34 @@ use bytemuck::cast_slice;
...
+    alloc: &'a Bump,
+) {
     builder.update_range_length(256);
+
+    // Create 31 columns, each will collect the corresponding word from all scalars.
+    // 31 because a scalar will only ever have 248 bits of data set.
+    let mut word_columns: Vec<&mut [u8]> = repeat(())
+        .take(31)
+        .map(|()| alloc.alloc_slice_fill_with(scalars.len(), |_| 0))
+        .collect();
```
iajoiner, line 47: I think we can use `core::iter::repeat_with` instead and simplify the code.

crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
```diff
@@ -32,8 +32,34 @@ use bytemuck::cast_slice;
...
+
+    // Initialize a vector to count occurrences of each byte (0-255).
+    // The vector has 256 elements padded with zeros to match the length of the word columns
+    // The size is the larger of 256 or the number of scalars.
+    let word_counts: &mut [i64] = alloc.alloc_slice_fill_with(max(256, scalars.len()), |_| 0);
```
iajoiner, line 51: `alloc.alloc_slice_fill_copy(max(256, scalars.len()), 0)`

crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
```diff
@@ -91,6 +117,82 @@ pub fn final_round_evaluate_range_check<'a, S: Scalar + 'a>(
     );
 }
 
+/// Verify that the prover claim is correct.
+///
+/// # Panics
+///
+/// if a column contains values outside of the selected range.
+pub fn verifier_evaluate_range_check<C>(
```
iajoiner: Let's use `S` for `Scalar` to be aligned with the rest of the project. `C` normally refers to commitments.

crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
Dustin-Ray:
```rust
    let mut word_columns: Vec<&mut [u8]> =
        core::iter::repeat_with(|| alloc.alloc_slice_fill_with(scalars.len(), |_| 0))
            .take(31)
            .collect();
```

doesnt seem that much simpler to me if this is what youre suggesting, but we can do this:

```rust
let mut word_columns: Vec<&mut [u8]> = (0..31)
    .map(|_| alloc.alloc_slice_fill_with(scalars.len(), |_| 0))
    .collect();
```

and this is a little better

iajoiner (approved):

PR #459, opened 2025-01-08
refactor!: sign expr redesign

stuarttimwhite:

# Rationale for this change

Sign_expr currently has the lead bit of its unsigned 256 bit masks indicating the sign of the value. This causes some branching logic that we would rather avoid. Instead, values shall be shifted to the right by 2^255, thereby reducing branching logic and still allowing these values to be represented by unsigned 256 bit masks.

1. TBD

# Are these changes tested?
TBD


crates/proof-of-sql/src/sql/proof_exprs/inequality_expr.rs
```diff
@@ -108,13 +100,7 @@ impl ProofExpr for InequalityExpr {
         let equals_zero = prover_evaluate_equals_zero(table.num_rows(), builder, alloc, diff);
```
JayWhite2357, line 100: We should be able to remove this and `prover_evaluate_or` as well.

crates/proof-of-sql/src/base/bit/bit_distribution.rs
```diff
@@ -1,70 +1,87 @@
...
 
-    #[allow(clippy::missing_panics_doc)]
-    pub fn sign_bit(&self) -> bool {
-        assert!(!self.has_varying_sign_bit());
-        self.or_all[3] & (1 << 63) != 0
+    /// # Panics
+    ///
+    /// Panics if lead bit varies but `bit_evals` is empty
+    pub fn leading_bit_eval<S: ScalarExt>(&self, bit_evals: &[S], one_eval: S) -> S {
+        if U256::from(self.vary_mask) & (U256::ONE << 255) != U256::ZERO {
```
JayWhite2357, line 71: NIT:
I just realized this is supported. `shl` is const.
```suggestion
        if U256::from(self.vary_mask) & 256::ONE.shl(255) != U256::ZERO {
```

crates/proof-of-sql/src/base/bit/bit_distribution.rs
```diff
@@ -73,92 +90,24 @@ impl BitDistribution {
...
-    where
-        F: FnMut(usize, usize),
-    {
-        for i in 0..4 {
-            let bitset = self.vary_mask[i];
-            for pos in BitIter::from(bitset) {
-                f(i, pos);
-            }
-        }
+        (self.leading_bit_inverse_mask() >> 128) == (U256::MAX >> 129)
```
JayWhite2357, line 98: NIT: See above.
```suggestion
        (self.leading_bit_inverse_mask() >> 128) == U256::MAX.shr(129)
```

crates/proof-of-sql/src/base/bit/bit_distribution.rs
```diff
@@ -73,92 +90,24 @@ impl BitDistribution {
...
-        }
-        for i in (0..3).rev() {
-            let mask = self.or_all[i];
-            if mask != 0 {
-                return 64 - (mask.leading_zeros() as usize) - 1 + 64 * i;
-            }
-        }
-        panic!("no bits are set");
+    ///
+    /// The panic shouldn't be mathematically possible
```
JayWhite2357, line 105: NIT: If this is true, then use `#[allow(clippy::missing_panic_doc)]` to ignore the lint.
You only need to document panics if you can actually panic.

PR #465, opened 2025-01-10
refactor!: inequality expr redefinition

stuarttimwhite:

# Rationale for this change

To simplify the logic for inequality expr, it is convenient to redefined inequality expr to be strictly < or >, rather than <= or >=.

Inequality_expr is redefined. Code is changed as needed to keep everything functional.

# Are these changes tested?
There are existing tests that have been modified.


crates/proof-of-sql/src/base/database/column_comparison_operation.rs
```diff
@@ -299,7 +299,10 @@ impl ComparisonOp for GreaterThanOrEqualOp {
         S: Scalar,
         T: Copy + Debug + Ord + Zero + Into<S>,
     {
-        ge_decimal_columns(lhs, rhs, left_column_type, right_column_type)
```
iajoiner, line 302: We need to rename `ge_decimal_columns` to `gt_decimal_columns` and `le_decimal_columns` to `lt_decimal_columns`

stuarttimwhite: Here "greater than or equal to" was replaced with "not less than or equal to" = "greater than". Are you suggesting that we rename `ge_decimal_columns` to `gt_decimal_columns` and redefine those functions accordingly? Either approach is fine with me. I think the results will be the same.

iajoiner: Sorry. I think this is fine.

iajoiner (approved):
Really thanks! This simplifies things.

JayWhite2357 (approved):

JayWhite2357:
You could do const generics like this: https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/503.
It's fine the way it is, though.

PR #467, opened 2025-01-10
fix: bump Rust to 1.84.0 and related fixes

iajoiner:

# Rationale for this change
udeps stop working after an update. Hence let's bump Rust to get CI to work again.

- bump Rust to 1.84.0
- clippy fixes

# Are these changes tested?

Existing tests should pass.

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/scalar_and_i256_conversions.rs
+++ b/crates/proof-of-sql/src/base/arrow/scalar_and_i256_conversions.rs
@@ -16,7 +16,7 @@ pub fn convert_scalar_to_i256<S: Scalar>(val: &S) -> i256 {
     let abs_scalar = if is_negative { -*val } else { *val };
     let limbs: [u64; 4] = abs_scalar.into();
 
-    let low = (limbs[0] as u128) | ((limbs[1] as u128) << 64);
+    let low = u128::from(limbs[0]) | (u128::from(limbs[1]) << 64);
     let high = i128::from(limbs[2]) | (i128::from(limbs[3]) << 64);
 
     let abs_i256 = i256::from_parts(low, high);
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -48,7 +48,7 @@ pub enum CommittableColumn<'a> {
     RangeCheckWord(&'a [u8]),
 }
 
-impl<'a> CommittableColumn<'a> {
+impl CommittableColumn<'_> {
     /// Returns the length of the column.
     #[must_use]
     pub fn len(&self) -> usize {
--- a/crates/proof-of-sql/src/base/database/owned_table_utility.rs
+++ b/crates/proof-of-sql/src/base/database/owned_table_utility.rs
@@ -169,7 +169,6 @@ pub fn int128<S: Scalar>(
 ///     scalar("a", [1, 2, 3]),
 /// ]);
 /// ```
-
 pub fn scalar<S: Scalar>(
     name: impl Into<Ident>,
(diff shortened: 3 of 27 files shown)


crates/proof-of-sql/src/base/commitment/table_commitment.rs
```diff
@@ -187,10 +186,9 @@ impl<C: Commitment> TableCommitment<C> {
     }
 
     /// Returns a [`TableCommitment`] to the provided table with the given row offset.
-    #[allow(
-        clippy::missing_panics_doc,
-        reason = "since OwnedTables cannot have columns of mixed length or duplicate idents"
-    )]
+    ///
+    /// # Panics
+    /// Shall never panic since [`OwnedTables`] cannot have columns of mixed length or duplicate idents
```
tlovell-sxt, line 191: NIT:
```suggestion
    /// Shall never panic since [`OwnedTable`]s cannot have columns of mixed length or duplicate idents
```

tlovell-sxt, line 191: I feel like it's a bit awkward that we're having to add panic docs to functions that we intend not to panic. Maybe we should `#[expect(clippy::missing_panic_doc)]` instead?

crates/proof-of-sql/src/sql/proof_exprs/equals_expr.rs
```diff
@@ -102,10 +102,10 @@ impl ProofExpr for EqualsExpr {
...
 }
 
-#[allow(
-    clippy::missing_panics_doc,
-    reason = "table_length is guaranteed to match lhs.len()"
-)]
+/// Evaluate if the result is zero
+///
+/// # Panics
+/// Panics if `table_length` does not match the length of `lhs`. Shouldn't happen in practice if the code is used as intended.
```
tlovell-sxt, line 108: Yeah this one is weird. this one seems like it should have a panic doc since there's nothing preventing callers from providing invalid input, even if the only "callers" are internal to proof-of-sql
```suggestion
/// Panics if `table_length` does not match the length of `lhs`.
```

crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
```diff
@@ -32,11 +32,10 @@ pub(crate) fn add_subtract_literals<S: Scalar>(
...
 }
 
-#[allow(
-    clippy::missing_panics_doc,
-    reason = "lhs and rhs are guaranteed to have the same length by design, ensuring no panic occurs"
-)]
 /// Add or subtract two columns together.
+///
+/// # Panics
+/// Panics if `lhs` and `rhs` are not of the same length which shouldn't ever happen.
```
tlovell-sxt, line 38: Same NIT here
```suggestion
/// Panics if `lhs` and `rhs` are not of the same length.
```

crates/proof-of-sql/src/sql/proof_exprs/or_expr.rs
```diff
@@ -89,10 +89,11 @@ impl ProofExpr for OrExpr {
...
 
-#[allow(
-    clippy::missing_panics_doc,
-    reason = "table_length matches lhs and rhs lengths, ensuring no panic occurs"
-)]
+/// Result evaluate `OrExpr`
+///
+/// # Panics
+/// Panics if `table_length` does not match the length of `lhs` or `rhs`.
+/// In practice, this should never happen.
```
tlovell-sxt, line 96: These too. I guess I'll just leave it as a blanket comment that I don't like claiming that something "shouldn't happen" if there's no type guarantees, even if it's only `pub(crate)`

tlovell-sxt (approved):

Issue #464, opened 2025-01-10
IPA seems to break down during range check proof while Dory seems to be ok


Dustin-Ray:
# Bug Report: Proof Breakdown in InnerProductProof System for RangeCheck

## Describe the Bug

The bug is found on [this branch](https://github.com/spaceandtimelabs/sxt-proof-of-sql/tree/feat/range-check-arbitrary-word-columns)

Go directly to the failing test [here](https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/bb7d8c8addafbbb906b6be199a5a2bf1e6b333da/crates/proof-of-sql/src/sql/proof_gadgets/range_check_test.rs#L329-L362)

The `InnerProductProof` system fails to verify a column of scalars when the range includes values up to `10241`, while the same proof system works correctly for ranges up to `10240`. Notably, the `DoryProof` system verifies the same range (`0..10241`) without any issues.

### Minimal Example
Here is a minimal test that reproduces the issue in the `InnerProductProof` system:

```rust
#[test]
fn ipa_proof_breaks_down_on_range_check() {
    // Create a column of scalars from 0 to 10240
    let data = owned_table::<Curve25519Scalar>([scalar("a", 0..10240)]);
    let t = "sxt.t".parse().unwrap();
    let accessor = OwnedTableTestAccessor::<InnerProductProof>::new_from_table(t, data, 0, ());
    let ast = RangeCheckTestPlan {
        column: ColumnRef::new(t, "a".into(), ColumnType::Scalar),
    };
    let verifiable_res = VerifiableQueryResult::<InnerProductProof>::new(&ast, &accessor, &());
    let res: Result<QueryData<MontScalar<ark_curve25519::FrConfig>>, QueryError> =
        verifiable_res.verify(&ast, &accessor, &());
    assert!(res.is_ok());

    // Now create a column of scalars from 0 to 10241
    let data = owned_table::<Curve25519Scalar>([scalar("a", 0..10241)]);
    let t = "sxt.t".parse().unwrap();
    let accessor = OwnedTableTestAccessor::<InnerProductProof>::new_from_table(t, data, 0, ());
    let ast = RangeCheckTestPlan {
        column: ColumnRef::new(t, "a".into(), ColumnType::Scalar),
    };
    let verifiable_res = VerifiableQueryResult::<InnerProductProof>::new(&ast, &accessor, &());
[...]

PR #473, opened 2025-01-13
feat: range check arbitrary size columns 

Dustin-Ray:
# Rationale for this change

Currently, it is possible to argue the correct ranges of scalars up to the word value size, but anything below or beyond this value (currently 256 for byte sized words) will fail to verify. This PR expands range check on a scalar column to arbitrary lengths.

- [x] arbitrary lengths beyond maximum word value
- [x] arbitrary lengths below maximum word value
- benches were carried out in criterion and jaeger
- refactored inner functions to make the loops more readable, will move back to idiomatic representations before marking this ready for merge

# Are these changes tested?

- [x] test can range check up to 2^248 boundary
- [x] test cannot range check past 2^ 248 boundary
- [x] test cannot range check below 0
- [x] test can range check below the maximum word size value
- [x] range check with dory
- [x] range check with ipa - currently failing as certain column lengths for unknown reason, see #464


crates/proof-of-sql/src/sql/proof_gadgets/mod.rs
```diff
@@ -5,6 +5,7 @@ use bitwise_verification::{verify_constant_abs_decomposition, verify_constant_si
 mod bitwise_verification_test;
 mod sign_expr;
 pub(crate) use sign_expr::{prover_evaluate_sign, result_evaluate_sign, verifier_evaluate_sign};
+#[allow(clippy::needless_range_loop)] // keep the loop for readability for now, refactor later
```
Dustin-Ray, line 8: I will change this back immediately in the next PR, this was super important to keep for now so that we can understand the loops in the range check functions

crates/proof-of-sql/src/sql/proof_gadgets/range_check_test.rs
```diff
@@ -137,8 +146,28 @@ mod tests {
     }
 
     #[test]
-    fn we_can_prove_a_range_check_with_range_0_to_256() {
-        let data = owned_table([scalar("a", 0..256)]);
+    #[allow(clippy::cast_sign_loss)]
```
Dustin-Ray, line 149: this is added because the range used to index the scalars while creating the owned table can be negative but here it doesnt matter because its testing positive ranges only and we cant create a Scalar from but Uints

crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
```diff
@@ -134,31 +134,36 @@ fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
...
+        }
+    }
+
+    // Count the occurrences of each byte
+    for byte_index in 0..31 {
+        for i in 0..scalars.len() {
+            let byte = word_columns[byte_index][i];
+            byte_counts[byte as usize] += 1;
+        }
+    }
```
iajoiner, line 154: Hmm there should be a simpler and more functional way to write this

PR #474, opened 2025-01-14
refactor!: update scalar challenges to use mask

JayWhite2357:
# Rationale for this change

The current mechanism for drawing a challenge is a bit opaque. This simplifies it.

When drawing a challenge from the transcript, it is 32 bytes long. Instead of using `[u64; 4]` to coerce it into a `Scalar`, it is more straightforward to simply drop the first few bits. This can be accomplished by masking the challenged by a `CHALLENGE_MASK` before converting to a `Scalar`.

Additionally, this PR updates the `Scalar` implementations so that we have a blanket implementation of `impl<T> Scalar for MontScalar<T>`. This removes the need to manually figure out the constants for each Scalar type.

# Are these changes tested?

Yes.

### DIFF
--- a/crates/proof-of-sql/src/base/proof/transcript_core.rs
+++ b/crates/proof-of-sql/src/base/proof/transcript_core.rs
@@ -56,7 +56,11 @@ impl<T: TranscriptCore> Transcript for T {
         self.extend_as_be::<[u64; 4]>(messages.into_iter().map(RefInto::ref_into));
     }
     fn scalar_challenge_as_be<S: Scalar>(&mut self) -> S {
-        receive_challenge_as_be::<[u64; 4]>(self).into()
+        let mut challenge = receive_challenge_as_be::<[u64; 4]>(self);
+        for (c, m) in challenge.iter_mut().zip(S::CHALLENGE_MASK) {
+            *c &= m;
+        }
+        challenge.into()
     }
     fn challenge_as_le(&mut self) -> [u8; 32] {
         self.raw_challenge()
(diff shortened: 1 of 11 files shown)


crates/proof-of-sql/src/base/proof/transcript_core.rs
```diff
@@ -56,7 +56,11 @@ impl<T: TranscriptCore> Transcript for T {
         self.extend_as_be::<[u64; 4]>(messages.into_iter().map(RefInto::ref_into));
     }
     fn scalar_challenge_as_be<S: Scalar>(&mut self) -> S {
-        receive_challenge_as_be::<[u64; 4]>(self).into()
+        let mut challenge = receive_challenge_as_be::<[u64; 4]>(self);
```
stuarttimwhite, line 59: Maybe we could use bnum::U256 here? Avoid the zip. Feel free to resolve if that's not your preference.

crates/proof-of-sql/src/base/scalar/mont_scalar.rs
```diff
@@ -432,17 +433,17 @@ where
...
     const TWO: Self = Self(Fp::new(ark_ff::BigInt([2, 0, 0, 0])));
     const TEN: Self = Self(Fp::new(ark_ff::BigInt([10, 0, 0, 0])));
-    const CHALLENGE_MASK: [u64; 4] = {
+    const CHALLENGE_MASK: U256 = {
         assert!(
             T::MODULUS.0[3].leading_zeros() < 64,
             "modulus expected to be larger than 1 << (64*3)"
         );
-        [
+        U256::from_digits([
```
stuarttimwhite, line 441: I think this would be the same as `U256::MAX >> (T::MODULUS.0[3].leading_zeros() + 1)`

JayWhite2357: That was my initial goal, but `>>` is non-const.

stuarttimwhite: That's too bad.

stuarttimwhite (approved):

PR #476, opened 2025-01-14
refactor: separate column evaluations from final round mle evaluations

JayWhite2357:
# Rationale for this change

The source of the data/commitments for referenced columns is different from final round mles. They are also handle very differently. So, they should be separated in the proof as well.

See rationale and title.

# Are these changes tested?

Yes, by existing tests.


Dustin-Ray (approved):
LGTM - Thanks

PR #477, opened 2025-01-15
feat: add rho eval

iajoiner:

# Rationale for this change
For joins it is necessary to have not only one evaluations but also evaluations of rho = (0, 1, 2, \cdots, n-1) vectors. Hence we need to add them to our proofs.

- an algorithm to compute rho evaluations

# Are these changes tested?

Yes.

### DIFF
--- a/crates/proof-of-sql/src/base/polynomial/lagrange_basis_evaluation_test.rs
+++ b/crates/proof-of-sql/src/base/polynomial/lagrange_basis_evaluation_test.rs
@@ -1,7 +1,7 @@
 use crate::base::{
     polynomial::{
-        compute_evaluation_vector, compute_truncated_lagrange_basis_inner_product,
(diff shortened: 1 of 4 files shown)


crates/proof-of-sql/src/base/polynomial/lagrange_basis_evaluation_test.rs
```diff
@@ -108,46 +154,46 @@ fn compute_truncated_lagrange_basis_sum_gives_correct_values_with_3_variables()
 }
 
 #[test]
-fn compute_truncated_lagrange_basis_sum_gives_correct_values_with_3_variables_using_dalek_scalar() {
```
iajoiner, line 111: Since the test was formerly a duplicate of the test above I just modified it so that we check that the rho eval is correct with 3 variables.

JayWhite2357 (approved):

PR #481, opened 2025-01-15
perf: enhance range check operations

Dustin-Ray:
# Rationale for this change

The current range check gadget was implemented with a focus on clarity of the multiple cryptographic steps involved in forming the argument. Now that the gadget is functional, we can focus on improving it's performance.

- [x] Commit to u8-sized words where possible, so anything that doesnt require inversion should just be a simple u8, this should reduce commitment size and times to some extent
- [x] clean up suppressed clippy lints
- [x] add jaeger tracing spans
- [x] initial pass at reducing mutability of range check inner functions and data structures, will follow up with the rest of this in the next PR

# Are these changes tested?
Covered by existing tests

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -25,6 +25,8 @@ use proof_of_sql_parser::posql_time::{PoSQLTimeUnit, PoSQLTimeZone};
 pub enum CommittableColumn<'a> {
     /// Borrowed Bool column, mapped to `bool`.
     Boolean(&'a [bool]),
+    /// Borrowed `Byte` column, mapped to `u8`.
+    Uint8(&'a [u8]),
     /// Borrowed `TinyInt` column, mapped to `i8`.
     TinyInt(&'a [i8]),
     /// Borrowed `SmallInt` column, mapped to `i16`.
@@ -43,9 +45,6 @@ pub enum CommittableColumn<'a> {
     VarChar(Vec<[u64; 4]>),
     /// Borrowed Timestamp column with Timezone, mapped to `i64`.
     TimestampTZ(PoSQLTimeUnit, PoSQLTimeZone, &'a [i64]),
-    /// Borrowed byte column, mapped to `u8`. This is not a `PoSQL`
-    /// type, we need this to commit to words in the range check.
-    RangeCheckWord(&'a [u8]),
 }
 
 impl CommittableColumn<'_> {
@@ -53,6 +52,7 @@ impl CommittableColumn<'_> {
     #[must_use]
     pub fn len(&self) -> usize {
         match self {
+            CommittableColumn::Uint8(col) => col.len(),
             CommittableColumn::TinyInt(col) => col.len(),
             CommittableColumn::SmallInt(col) => col.len(),
             CommittableColumn::Int(col) => col.len(),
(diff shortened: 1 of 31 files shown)


crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
```diff
@@ -184,45 +232,42 @@ fn decompose_scalar_to_words<'a, S: Scalar + 'a>(
...
 
-        // Provide the inverted column to the builder
         builder.produce_intermediate_mle(words_inv as &[_]);
 
-        // Copy the inverted values into the user-provided `inverted_word_columns`
         inv_column.copy_from_slice(words_inv);
 
-        // Prepare a column of "true" (1-bit flags) to use in the final polynomial check
-        let input_ones = alloc.alloc_slice_fill_copy(table_length, true);
+        let input_ones = alloc.alloc_slice_fill_copy(inverted_word_columns[0].len(), true);
```
Dustin-Ray, line 269: maybe theres a cleaner way to do this

crates/proof-of-sql/src/base/commitment/committable_column.rs
```diff
@@ -77,6 +77,33 @@ impl CommittableColumn<'_> {
     pub fn column_type(&self) -> ColumnType {
         self.into()
     }
+
+    /// For range check words, we need to know if the column is signed or not.
```
iajoiner, line 81: Shall we add `RangeCheckWord` to `ColumnType` instead?

Dustin-Ray: Yeah, thats a cleaner solution. See #498

Dustin-Ray:
This PR now depends on #498 which gives full u8 support as a column type.

Dustin-Ray:
> This PR now depends on #498 which gives full u8 support as a column type.

#498 Successfully merged

crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
```diff
@@ -86,24 +84,60 @@ pub(crate) fn final_round_evaluate_range_check<'a, S: Scalar + 'a>(
...
     let alpha = builder.consume_post_result_challenge();
 
+    // avoids usize to u8 cast
+    let mut word_value_table = [0u8; 256];
+    let mut inv_word_values_plus_alpha_table = [S::ZERO; 256];
+    for i in 0u8..=255 {
+        word_value_table[i as usize] = i;
+        inv_word_values_plus_alpha_table[i as usize] = S::from(&i);
+    }
+    let word_val_table: &mut [u8] =
```
iajoiner, line 112: Is this needed, may I ask?

crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
```diff
@@ -86,24 +84,59 @@ pub(crate) fn final_round_evaluate_range_check<'a, S: Scalar + 'a>(
...
         alpha,
-        table_length,
         &mut inverted_word_columns,
+        inv_word_vals_plus_alpha_table,
     );
+    span.exit();
 
     // Produce an MLE over the word values
-    prove_word_values(alloc, alpha, builder);
+    prove_word_values(
```
iajoiner, line 133: NIT: Actually in the future maybe we want to put all the proof components together in the same function just like how we handle `proof_expr` and `proof_plan`.

iajoiner (approved):
I think this one is fine. Please rebase and merge.

crates/proof-of-sql/src/sql/proof_gadgets/range_check.rs
Dustin-Ray: With this gadget, I really do not want to take that route yet. In a future PR after mutability is reduced, I will commit to revisiting this, but right now there is simply too much code and too much complexity for my own comfort to inline everything into the same function.

PR #488, opened 2025-01-17
refactor: move examples out of check script, create scripts for testing and examples

Dustin-Ray:
# Rationale for this change

Currently, we have shell scripts to run the CI jobs locally before submitting commits to a PR, to ensure that remote CI jobs are successful with as few runs as possible. The current check job includes several computationally expensive examples that should not be included in in a check workflow. Making this refactor makes local CI simulation more reliable and reduces load of the remote CI runners.

- [x] moves examples into distinct block in lint-and-test yaml for CI jobs
- [x] creates testing and examples scripts for separating the local CI jobs

# Are these changes tested?
- [x] Test that check script correctly identifies all clippy lints in lint-and-release yaml
- [x] test that check script only runs checks
- [x] test that examples script only runs examples
- [x] test that test script only runs test

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -146,7 +146,6 @@ jobs:
         - name: Run posql_db example (Without Blitzar)
           run: bash crates/proof-of-sql/examples/posql_db/run_example.sh --no-default-features --features="rayon"
 
-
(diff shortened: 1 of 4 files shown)


iajoiner (approved):
Really thanks for this!

PR #489, opened 2025-01-18
refactor!: use only serde to extend transcript in QueryProof

JayWhite2357:
# Rationale for this change

The QueryProof creation currently has a lot of custom transcript handling that differs from the QueryProof's serialization format. While the transcript handling aligns with EVM verification, it requires custom serialization. This PR makes it so that the two are identical.

* The QueryProof is broken up into smaller parts, each of which is put onto the transcript in aggregate. This simplifies the code.
* Each message uses serde to serialize the message and put it on the transcript. The underlying transcript dictates the actual serialization format.

# Are these changes tested?
They are tested by existing tests. Future PRs will have tests showing EVM compatibility.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof/query_proof.rs
+++ b/crates/proof-of-sql/src/sql/proof/query_proof.rs
@@ -6,21 +6,20 @@ use super::{
 use crate::{
     base::{
         bit::BitDistribution,
-        commitment::{Commitment, CommitmentEvaluationProof},
+        commitment::CommitmentEvaluationProof,
         database::{
-            ColumnRef, CommitmentAccessor, DataAccessor, MetadataAccessor, OwnedColumn, OwnedTable,
-            Table, TableRef,
+            ColumnRef, CommitmentAccessor, DataAccessor, MetadataAccessor, OwnedTable, Table,
+            TableRef,
         },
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/sql/proof/query_proof.rs
```diff
@@ -46,37 +45,51 @@ fn get_index_range<'a>(
...
-/// The proof for a query.
-///
-/// Note: Because the class is deserialized from untrusted data, it
-/// cannot maintain any invariant on its data members; hence, they are
-/// all public so as to allow for easy manipulation for testing.
 #[derive(Clone, Serialize, Deserialize)]
-pub(super) struct QueryProof<CP: CommitmentEvaluationProof> {
-    /// Bit distributions
-    pub bit_distributions: Vec<BitDistribution>,
+pub struct QueryProofMessage1<C> {
```
iajoiner, line 49: Maybe we should use `QueryProofMessageFirstRound` and `QueryProofMessageFinalRound` instead?

JayWhite2357: Good idea.

iajoiner (approved):

PR #490, opened 2025-01-18
feat: simplify serialization of proof plans for EVM compatibility

JayWhite2357:
# Rationale for this change

Previously I merged a custom EVM serializer for the query plan. This incurs a maintenance cost. This PR leverages serde instead, at the cost of a bit less flexibility.

The `DynProofPlan` is converted to a simplified plan. The main change is removing any table and column references.
This simplified plan can be serialized using serde (in particular, bincode+bigendian+fixint gives an easy EVM format).
Ultimately, this is used by an `EVMProofPlan` that is a newtype wrapper around `DynProofPlan` that serializes through this simplified plan. This allows `EVMProofPlan` to implement `ProofPlan` while still having a serialization different from the more complex `DynProofPlan`s.

# Are these changes tested?
Yes. Later PRs will demonstrate EVM compatibility.


crates/proof-of-sql/src/sql/evm_proof_plan/proof_plan.rs
```diff
@@ -0,0 +1,103 @@
...
+        let columns = column_refs
+            .into_iter()
+            .map(|column_ref| {
+                table_refs
+                    .get_index_of(&column_ref.table_ref())
+                    .map(|table_index| (table_index, column_ref.column_id().to_string()))
+            })
+            .collect::<Option<_>>()
+            .ok_or(Error::TableNotFound)
+            .map_err(serde::ser::Error::custom)?;
```
Dustin-Ray, line 53: NIT: You _could_ potentially write this as:

```rust
let columns = column_refs
    .into_iter()
    .map(|column_ref| {
        let table_index = table_refs
            .get_index_of(&column_ref.table_ref())
            .ok_or(Error::TableNotFound)?;
        Ok((table_index, column_ref.column_id().to_string()))
    })
    .collect::<Result<Vec<(usize, String)>, Error>>() // Error specified here
    .map_err(serde::ser::Error::custom)?;  
```

Which may or may not be more readable depending on personal preference, but the original is fine too.

crates/proof-of-sql/src/sql/evm_proof_plan/proof_plan.rs
```diff
@@ -0,0 +1,103 @@
...
+pub struct EVMProofPlan(pub DynProofPlan);
+
+impl Serialize for EVMProofPlan {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        #[derive(Serialize)]
+        struct CompactPlan {
+            tables: Vec<String>,
+            columns: Vec<(usize, String)>,
+            plan: Plan,
+        }
```
Dustin-Ray, line 37: NIT: Its interesting seeing this struct in the method vs the module, as long as this never gets used anywhere else I think its ok. Does it need any doc comments at all? Maybe not as important if its just serialized and sent over the wire

crates/proof-of-sql/src/sql/evm_proof_plan/proof_plan.rs
```diff
@@ -0,0 +1,103 @@
...
+            .verifier_evaluate(builder, accessor, result, one_eval_map)
+    }
+    fn get_column_result_fields(&self) -> Vec<ColumnField> {
+        self.0.get_column_result_fields()
+    }
+    fn get_column_references(&self) -> IndexSet<ColumnRef> {
+        self.0.get_column_references()
+    }
+    fn get_table_references(&self) -> IndexSet<TableRef> {
+        self.0.get_table_references()
```
Dustin-Ray, line 83: NIT: named fields are always more readable compared to tuple indexing

crates/proof-of-sql/src/sql/evm_proof_plan/exprs.rs
```diff
@@ -0,0 +1,82 @@
...
+        database::{ColumnRef, LiteralValue},
+        map::IndexSet,
+    },
+    sql::proof_exprs::{self, DynProofExpr},
+};
+use alloc::boxed::Box;
+use serde::Serialize;
+
+#[derive(Serialize)]
+pub enum Expr {
```
Dustin-Ray, line 13: A small doc comment here might be nice

crates/proof-of-sql/src/sql/evm_proof_plan/exprs.rs
```diff
@@ -0,0 +1,82 @@
...
+    Column(ColumnExpr),
+    Equals(EqualsExpr),
+    Literal(LiteralExpr),
+}
+impl Expr {
+    pub fn try_from_proof_expr(
```
Dustin-Ray, line 19: Same here, a doc comment on anything pub would be nice, unless theres a good reason not to do so for this method.

Dustin-Ray:
Overall it's looking good, theres some inconsistency with the comments on ```pub``` methods and enums, but maybe theres a good reason for that here.

crates/proof-of-sql/src/sql/evm_proof_plan/proof_plan.rs
JayWhite2357: I put it inside the method since I wanted to make it very clear that this should only ever be use to serialize it. (De-serialization isn't possible since this is lossy.)

crates/proof-of-sql/src/sql/evm_proof_plan/proof_plan.rs
JayWhite2357: See my fixup commit: https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/490/commits/4111da94b5f7fef22ea929ec3b4b707ffa7526bb.

crates/proof-of-sql/src/sql/evm_proof_plan/proof_plan.rs
JayWhite2357: See my fixup commit: https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/490/commits/4111da94b5f7fef22ea929ec3b4b707ffa7526bb.

crates/proof-of-sql/src/sql/evm_proof_plan/exprs.rs
JayWhite2357: See my fixup commit: https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/490/commits/4111da94b5f7fef22ea929ec3b4b707ffa7526bb.

crates/proof-of-sql/src/sql/evm_proof_plan/exprs.rs
JayWhite2357: See my fixup commit: https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/490/commits/4111da94b5f7fef22ea929ec3b4b707ffa7526bb.

PR #492, opened 2025-01-19
fix: type conversion in expression aggregation

offensif:

# Rationale for this change

The original code uses a .into() call after the Expression::Aggregation { ... }. However, the expr field type in AliasedResultExpr expects Box<Expression>, and the additional conversion via .into() may be redundant. The correct way is to directly create the Box<Expression> without using .into().


iajoiner (approved):
@offensif Really thanks for your contribution! This looks good!

PR #502, opened 2025-01-22
feat: add increasing gadget

iajoiner:
Please be sure to look over the pull request guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md#submit-pr.

Depends on #487

# Please go through the following checklist
- [x] The PR title and commit messages adhere to guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md. In particular `!` is used if and only if at least one breaking change has been introduced.
- [x] I have run the ci check script with `source scripts/run_ci_checks.sh`.

# Rationale for this change
This is a very simple gadget that is a variant of the uniqueness one (please merge that first) which can be used for ordering.

# What changes are included in this PR?
- add increasing gadget.

# Are these changes tested?

Yes.

PR #503, opened 2025-01-22
feat: add monotonic gadget

iajoiner:
Replaces #487 and #502

# Rationale for this change
In order to do inner joins and ordering we need to be able to prove that a single column is increasing, strictly or not.

- add monotonic gadget.

# Are these changes tested?

Yes.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_gadgets/monotonic.rs
+++ b/crates/proof-of-sql/src/sql/proof_gadgets/monotonic.rs
@@ -1,0 +1,143 @@
+//! Prove that a column is increasing or decreasing, strictly or non-strictly.
+use super::{
+    final_round_evaluate_shift, first_round_evaluate_shift, prover_evaluate_sign,
+    verifier_evaluate_sign, verify_shift,
+};
+use crate::{
+    base::{proof::ProofError, scalar::Scalar},
+    sql::proof::{FinalRoundBuilder, FirstRoundBuilder, VerificationBuilder},
+};
+use alloc::vec;
+use bumpalo::Bump;
+
+/// Perform first round evaluation of monotonicity.
+pub(crate) fn first_round_evaluate_monotonic<S: Scalar>(
+    builder: &mut FirstRoundBuilder<'_, S>,
+    num_rows: usize,
+) {
+    builder.produce_one_evaluation_length(num_rows + 1);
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/sql/proof_gadgets/monotonic.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Perform final round evaluation of monotonicity.
+#[allow(clippy::too_many_arguments)]
+pub(crate) fn final_round_evaluate_monotonic<'a, S: Scalar>(
+    builder: &mut FinalRoundBuilder<'a, S>,
+    alloc: &'a Bump,
+    alpha: S,
+    beta: S,
+    column: &'a [S],
+    strictness: bool,
```
JayWhite2357, line 30: Can we make this a const generic?

crates/proof-of-sql/src/sql/proof_gadgets/monotonic.rs
```diff
@@ -0,0 +1,132 @@
...
+    let sign_eval = verifier_evaluate_sign(builder, ind_eval, shifted_one_eval)?;
+    let singleton_one_eval = builder.mle_evaluations.singleton_one_evaluation;
+    let allowed_evals = if strictness {
+        vec![
+            one_eval,
+            shifted_one_eval - singleton_one_eval,
+            one_eval - singleton_one_eval,
+        ]
+    } else {
+        vec![
```
JayWhite2357, line 120:
```suggestion
        [
            one_eval,
            shifted_one_eval - singleton_one_eval,
            one_eval - singleton_one_eval,
        ]
    } else {
        [
```

crates/proof-of-sql/src/sql/proof_gadgets/monotonic.rs
iajoiner: @JayWhite2357 You mean make strictness a const parameter?

JayWhite2357: Also, I think it would be pretty easy to add an `increasing` parameter as well so that we can make this any of
* strictly increasing
* non-strictly increasing
* strictly decreasing
* non-strictly decreasing

JayWhite2357 (approved):
Looks great! Thanks.

PR #508, opened 2025-01-24
feat: evm verifier

JayWhite2357:
# Rationale for this change

EVM verifier

# What changes are included in this PR?

EVM verifier

# Are these changes tested?

Yes

PR #511, opened 2025-01-29
feat: add Yul Transcript and MathUtil libraries

JayWhite2357:
# Rationale for this change

This PR is the first of many breaking #508 into manageable pieces.

The following Yul functions are added, along with testing around them. They will be imported in future PRs.
* `log2_up`
* `draw_challenge`
* `draw_challenges`
* `append_calldata`

# Are these changes tested?
Yes

### DIFF
--- a/solidity/src/base/MathUtil.sol
+++ b/solidity/src/base/MathUtil.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: UNLICENSED
+// This is licensed under the Cryptographic Open Software License 1.0
+pragma solidity ^0.8.28;
+
+/// @title Math Utilities Library
+/// @notice Provides functions to perform various math operations
+library MathUtil {
+    /// @notice Computes `max(1,ceil(log_2(value)))`
+    /// @dev The smallest integer greater than or equal to the base 2 logarithm of a number.
+    /// If the number is less than 2, the result is 1.
+    /// @param value0 The input value for which to compute the logarithm
+    /// @return exponent0 The computed logarithm value
+    function log2Up(uint256 value0) internal pure returns (uint256 exponent0) {
+        assembly {
+            function log2_up(value) -> exponent {
+                if value { value := sub(value, 1) }
+                exponent := 1
+                for {} shr(exponent, value) {} { exponent := add(exponent, 1) }
+            }
+            exponent0 := log2_up(value0)
+        }
+    }
+}
--- a/solidity/src/base/Transcript.sol
+++ b/solidity/src/base/Transcript.sol
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: UNLICENSED
+// This is licensed under the Cryptographic Open Software License 1.0
+pragma solidity ^0.8.28;
+
+// assembly only constants
+// solhint-disable-next-line no-unused-import
+import {FREE_PTR, MODULUS_MASK, WORD_SIZE} from "./Constants.sol";
+
+/// @title Transcript Library
+/// @notice Provides functions to manage a simple public coin transcript
+/// @dev The transcript is a sequence of messages and challenges, where each challenge is a scalar (less than MODULUS).
+library Transcript {
+    /// @notice Draw a challenge from the transcript, and update the state of the transcript.
+    /// @dev The result should be a scalar value (less than MODULUS).
+    /// The challenges is a masked version of the current state of the transcript.
+    /// The new state of the transcript is the hash of the current state.
+    /// @param transcript0 The current state of the transcript
+    /// @return result0 The drawn challenge
+    function drawChallenge(uint256[1] memory transcript0) internal pure returns (uint256 result0) {
+        assembly {
+            function draw_challenge(transcript_ptr) -> result {
+                result := and(mload(transcript_ptr), MODULUS_MASK)
+                mstore(transcript_ptr, keccak256(transcript_ptr, WORD_SIZE))
+            }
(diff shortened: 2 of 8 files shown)


solidity/src/base/MathUtil.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: UNLICENSED
+// This is licensed under the Cryptographic Open Software License 1.0
+pragma solidity ^0.8.28;
+
+/// A library for math utilities.
+library MathUtil {
+    /// Computes `max(1,ceil(log_2(value)))`,
```
yosriady, line 7: For docstrings, consider following the NatSpec format https://docs.soliditylang.org/en/latest/natspec-format.html

solidity/src/base/MathUtil.sol
```diff
@@ -0,0 +1,55 @@
...
+            function log2_up(value) -> exponent {
+                if value { value := sub(value, 1) }
+                exponent := 1
+                for {} shr(exponent, value) {} { exponent := add(exponent, 1) }
+            }
+            exponent_ := log2_up(value_)
+        }
+    }
+
+    function testWeCanComputeLog2Up() public pure {
```
yosriady, line 22: Move test functions outside of the libraries e.g. to `/solidity/test/MathUtil.t.sol`

solidity/src/base/MathUtil.sol
```diff
@@ -0,0 +1,55 @@
...
+        for (uint256 i = 9; i <= 16; ++i) {
+            assert(log2Up(i) == 4);
+        }
+        for (uint256 i = 17; i <= 32; ++i) {
+            assert(log2Up(i) == 5);
+        }
+        /* solhint-enable gas-strict-inequalities */
+    }
+
+    function testFuzzLog2Up(uint256 value) public pure {
```
yosriady, line 42: Nice, forge should run this as a property based test

https://book.getfoundry.sh/forge/fuzz-testing

yosriady (commented):
I recommend moving unit tests to its own file so it's not imported to the production contracts and become callable in the public contracts

Also consider adding Github Actions to run the following test and coverage commands in each PR:

https://book.getfoundry.sh/reference/forge/forge-test
https://book.getfoundry.sh/reference/forge/forge-coverage

You can see examples of this in Wael's zkpay repo

solidity/src/base/MathUtil.sol
JayWhite2357: Happy to do this, but here's some context.
This library will never actually get used. Ultimately, the Yul will get extracted and moved to a different file via a preprocessing script.

The library is entirely wrapper functionality in order to test and hold the Yul itself.

See https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/solidity/src/Placeholder.pre.sol and https://github.com/spaceandtimelabs/sxt-proof-of-sql/tree/main/solidity/scripts as well as #512.

So, everything in this file is not really going to be production code. So, perhaps it all belongs in `test`. On the other hand, the Yul snippets are production code.

JayWhite2357:
@yosriady We do have with tests here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/ae5788930933489f73588f0623d2c9d46895fc42/.github/workflows/lint-and-test.yml#L249.
This calls this script https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/solidity/scripts/test.sh.

I'll add coverage as well.

solidity/src/base/MathUtil.sol
yosriady: I see. In that case it makes sense to move it all to `test`
Otherwise, external auditors will ask the same question

solidity/src/base/Transcript.sol
```diff
@@ -0,0 +1,202 @@
...
+    /// The result should be a scalar value (less than MODULUS).
+    /// The challenges is a masked version of the current state of the transcript.
+    /// The new state of the transcript is the hash of the current state.
+    ///
+    /// In particular, the critical property is that the challenges are deterministic and uniformly distributed,
+    /// but with no collisions.
+    function drawChallenge(uint256[1] memory transcript_) private pure returns (uint256 result_) {
+        assembly {
+            function draw_challenge(transcript_ptr) -> result {
+                result := and(mload(transcript_ptr), MODULUS_MASK)
```
waelsy123, line 23: since that MODULUS_MASK is always ones, do we still need to apply AND operation? 

result := mload(transcript_ptr)

waelsy123 (approved):
looks great

solidity/test/base/Transcript.t.sol
```diff
@@ -0,0 +1,127 @@
+// SPDX-License-Identifier: UNLICENSED
+// This is licensed under the Cryptographic Open Software License 1.0
+pragma solidity ^0.8.28;
+
+import {Transcript} from "../../src/base/Transcript.sol";
+import {MODULUS, WORD_SIZE, FREE_PTR} from "../../src/base/Constants.sol";
+
+library TranscriptTest {
+    function testWeCanDrawChallenge() public pure {
+        uint256[1] memory transcriptAPtr = [uint256(12345)];
```
waelsy123, line 10: you can do something like: 
using Transcript for uint256[1];

then: 
uint256 challengeA1 = transcriptAPtr.drawChallenge();

waelsy123 (approved):

iajoiner (approved):

PR #512, opened 2025-01-30
feat: add ECPrecompiles Yul library

JayWhite2357:
# Rationale for this change

This PR is the one of many breaking #508 into manageable pieces. In particular, this provides methods for elliptic curve operations.

The following Yul functions are added, along with some tests around them. They will be imported in future PRs.
* `ec_add`
* `ec_mul`
* `ec_pairing_x2`
* `ec_mul_assign`
* `calldata_ec_add_assign`
* `calldata_ec_mul_add_assign`

In addition to these methods, there is a lot of work done to preprocessing and CI to enable code coverage and tests to work properly.

# Are these changes tested?
Somewhat. It is difficult do any kind of thorough testing because the only mechanism that solidity has of computing ec operations is the precompiles.

### DIFF
--- a/solidity/scripts/preprocess_yul_imports.sh
+++ b/solidity/scripts/preprocess_yul_imports.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 if [ $# -ne 1 ] || [ ! -d "$1" ]; then
@@ -12,7 +12,8 @@ while IFS= read -r -d '' file; do
     outfile="${file%.pre.sol}.post.sol"
     echo "Processing $file -> $outfile"
     
-    if ! awk '
+    if ! gawk '
+    # Handle Yul imports
     /\/\/ IMPORT-YUL/ {
         import_path = $3
         # Resolve actual file path for reading
@@ -34,7 +35,8 @@ while IFS= read -r -d '' file; do
         
         # Extract function name from the complete signature
         match(signature, /function[[:space:]]+([[:alnum:]_]+)/, m)
-        print "        // IMPORTED-YUL " import_path "::" m[1]
+        print "            // IMPORTED-YUL " import_path "::" m[1]
+        print "            function exclude_coverage_start_" m[1] "() {} // solhint-disable-line no-empty-blocks"
         
         # Skip original function if not empty
         if (!(signature ~ /{[[:space:]]*}/)) {
@@ -68,12 +70,25 @@ while IFS= read -r -d '' file; do
                 break
             }
         }
+
+        print "            function exclude_coverage_stop_" m[1] "() {} // solhint-disable-line no-empty-blocks"
         close(import_file)
         
         if (!found) {
             print "Error: Function " m[1] " not found in " import_file > "/dev/stderr"
             exit 1
         }
+        next
+    }
+    # Handle Solidity imports
+    /^import/ {
+        import_line = $0
+        while (import_line !~ /;[[:space:]]*$/) {
+            getline
+            import_line = import_line "\n" $0
+        }
+        gsub(/\.pre\.sol/, ".post.sol", import_line)
+        print import_line
         next
     }
     { print }
--- a/solidity/src/base/Constants.sol
+++ b/solidity/src/base/Constants.sol
@@ -2,11 +2,51 @@
 // This is licensed under the Cryptographic Open Software License 1.0
 pragma solidity ^0.8.28;
 
-// This is the modulus of the bn254 scalar field.
+/// @dev The modulus of the bn254 scalar field.
 uint256 constant MODULUS = 0x30644e72_e131a029_b85045b6_8181585d_2833e848_79b97091_43e1f593_f0000001;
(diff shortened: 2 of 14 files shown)


yosriady (commented):
Like the other PR #511, I highly recommend moving the test functions out to a separate file

To test library functions you may need to create mock test contracts that imports the library

solidity/src/base/Constants.sol
```diff
@@ -0,0 +1,34 @@
...
+uint256 constant WORDX10_SIZE = 0x140;
+uint256 constant WORDX11_SIZE = 0x160;
+uint256 constant WORDX12_SIZE = 0x180;
+
+uint256 constant ECADD_ADDRESS = 0x06;
+uint256 constant ECMUL_ADDRESS = 0x07;
+uint256 constant ECPAIRING_ADDRESS = 0x08;
+uint256 constant ECADD_GAS = 150;
+uint256 constant ECMUL_GAS = 6000;
+uint256 constant ECPAIRINGX2_GAS = 113000;
```
waelsy123, line 23: isn't the gas for ecPairing = 45000 ?

also, why do we need 2x amount the gas?

waelsy123: is it because we pass 12 words in line 70?

JayWhite2357: It's 45000 + 34000 * number of pairings. So, this is the gas for 2 pairings.

waelsy123 (dismissed):

waelsy123:
it's worth to mention here that zksync they have their patched version of foundry to support their different zkEVM. so I thought it could be a good idea to use their foundry in ci especially if we want to deploy the verifier to zkEVM (zksync rollups) : https://foundry-book.zksync.io/getting-started/installation#ci-installation-with-github-actions

solidity/scripts/preprocess_yul_imports.sh
```diff
@@ -34,7 +35,8 @@ while IFS= read -r -d '' file; do
         
         # Extract function name from the complete signature
         match(signature, /function[[:space:]]+([[:alnum:]_]+)/, m)
-        print "        // IMPORTED-YUL " import_path "::" m[1]
+        print "            // IMPORTED-YUL " import_path "::" m[1]
+        print "            function exclude_coverage_start_" m[1] "() {} // solhint-disable-line no-empty-blocks"
```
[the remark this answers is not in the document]
JayWhite2357: Rather than requiring a developer to put the marker in themselves, it's just automatically part of the generated `.post.sol` file on the imported blocks.

A developer doesn't need to know about it. (And probably shouldn't manually use it.)

Only the `IMPORTED-YUL` blocks are ignored. The original block should not be ignored and is what needs to be covered.
If an imported block is not covered, we don't really care since it was already covered in the original location.

JayWhite2357:
> it's worth to mention here that zksync they have their patched version of foundry to support their different zkEVM. so I thought it could be a good idea to use their foundry in ci especially if we want to deploy the verifier to zkEVM (zksync rollups) : https://foundry-book.zksync.io/getting-started/installation#ci-installation-with-github-actions

Good idea. Let's do that in another PR. It might get painful with all of the preprocessing.

waelsy123 (approved):

iajoiner (approved):

solidity/src/base/Constants.sol
```diff
@@ -2,11 +2,51 @@
...
+/// @dev Error code for when ECADD inputs are invalid.
+uint256 constant INVALID_EC_ADD_INPUTS = 0x765bcba0_00000000_00000000_00000000_00000000_00000000_00000000_00000000;
+/// @dev Error code for when ECMUL inputs are invalid.
+uint256 constant INVALID_EC_MUL_INPUTS = 0xe32c7472_00000000_00000000_00000000_00000000_00000000_00000000_00000000;
+/// @dev Error code for when ECPAIRING inputs are invalid.
+uint256 constant INVALID_EC_PAIRING_INPUTS = 0x4385b511_00000000_00000000_00000000_00000000_00000000_00000000_00000000;
+
+/// @title Errors library
+/// @notice Library containing custom error definitions.
+library Errors {
```
waelsy123, line 45: consider having errors in separated file

solidity/test/base/ECPrecompiles.t.pre.sol
```diff
@@ -0,0 +1,136 @@
...
+    function testECAdd() public view {
+        uint256[4] memory argsPtr = [uint256(1), 2, 1, 2];
+        ECPrecompiles.ecAdd(argsPtr);
+        assert(argsPtr[0] == 0x030644e72e131a029b85045b68181585d97816a916871ca8d3c208c16d87cfd3);
+        assert(argsPtr[1] == 0x15ed738c0e0a7c92e7845f96b2ae9c0a68a6a449e3538fc7ff3ebf7a5a18a2c4);
+        // scratch space
+        assert(argsPtr[2] == 1);
+        assert(argsPtr[3] == 2);
+
+        argsPtr = [uint256(0), 0, 1, 2];
```
waelsy123, line 16: @JayWhite2357  as ecADD and ecMUL uses alt_bn128 curve. how could add a point (0,0) which it doesn't belong on the curve?

PR #515, opened 2025-02-01
docs: Fix case mismatch in import statement

defiberrys:

# Rationale for this change

On case-sensitive file systems (e.g., Linux), the import statement in `Contract.pre.sol` references `YulUtils.pre.sol`, but the actual file is named `yul_utils.sol`. This causes an import failure. Updating the import to match the correct file name ensures compatibility across different environments.

- Fixed case mismatch in the import statement for `YulUtils.pre.sol`.
- Ensured the import references the correct filename `yul_utils.sol`.

# Are these changes tested?

Yes, verified that the contract compiles correctly after the fix.

### DIFF
--- a/solidity/scripts/README.md
+++ b/solidity/scripts/README.md
@@ -63,7 +63,7 @@ contract YulUtils {
 
 Main contract (`Contract.pre.sol`):
 ```solidity
-import {YulUtils} from "../libs/YulUtils.pre.sol";
+import {YulUtils} from "../libs/yul_utils.sol";
 
 contract MyContract {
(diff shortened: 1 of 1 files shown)


JayWhite2357 (changes requested):
Follow the solidity style guide please.
See https://docs.soliditylang.org/en/latest/style-guide.html#contract-and-library-names.

defiberrys:
> JayWhite2357

[JayWhite2357](https://github.com/JayWhite2357) yes, I made a mistake. Please review [this](https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/518) pull request instead. I’m closing this one.

PR #522, opened 2025-02-03
chore: Fixed regex to correctly match job names with hyphens

defiberrys:

---

**# Rationale for this change**

The regular expression `^  [A-Za-z0-9_]+:` in the current script correctly captures lines containing job names, but it is limited to letters, digits, and underscores, which doesn't account for job names with hyphens (e.g., `deploy-prod` or `build-linux`). This could lead to issues if such job names are used in the workflow YAML files.

**# What changes are included in this PR?**

I’ve updated the regex in the script to allow job names containing hyphens as well, making it more flexible. The updated regex now correctly matches job names with letters, numbers, underscores, and hyphens:

```bash
sed -n -E '/^  examples:/,/^  [[:space:]]+[a-zA-Z0-9_-]+:/p' "$YAML_FILE"
```

This ensures that the script can handle job names like `deploy-prod` or `build-linux`, making it more robust.

**# Are these changes tested?**

Yes, the changes have been tested by running the script with a variety of job names (including those with hyphens) and verifying that it works as expected without issues.

### DIFF
--- a/scripts/run_posql_examples.sh
+++ b/scripts/run_posql_examples.sh
@@ -26,7 +26,7 @@ fi
 # 2) Then find lines containing `cargo` commands and strip off 'run:'.
 ################################################################################
 cargo_commands=$(
-  sed -n -E '/^  examples:/,/^  [A-Za-z0-9_]+:/p' "$YAML_FILE" \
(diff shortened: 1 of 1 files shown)


iajoiner (approved):
@defiberrys Really thanks for your work! This looks good. @Dustin-Ray ?

Dustin-Ray (approved):
LGTM - Thanks for this

PR #538, opened 2025-02-11
feat: support for variable length binary

Dustin-Ray:
# Rationale for this change

Introduces support for arbitrary length binary columns. This first pass introduces initial support for the type as per standard procedure for adding new PoSQL data types.

- [x] initial pass for var length binary

# Are these changes tested?
- [x] commitment tests
- [x] tests first pass functionality
- [x] tests arrow conversions

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/owned_and_arrow_conversions.rs
+++ b/crates/proof-of-sql/src/base/arrow/owned_and_arrow_conversions.rs
@@ -22,9 +22,9 @@ use crate::base::{
 use alloc::sync::Arc;
 use arrow::{
     array::{
-        ArrayRef, BooleanArray, Decimal128Array, Decimal256Array, Int16Array, Int32Array,
-        Int64Array, Int8Array, StringArray, TimestampMicrosecondArray, TimestampMillisecondArray,
-        TimestampNanosecondArray, TimestampSecondArray, UInt8Array,
+        ArrayRef, BinaryArray, BooleanArray, Decimal128Array, Decimal256Array, Int16Array,
+        Int32Array, Int64Array, Int8Array, StringArray, TimestampMicrosecondArray,
+        TimestampMillisecondArray, TimestampNanosecondArray, TimestampSecondArray, UInt8Array,
     },
     datatypes::{i256, DataType, Schema, SchemaRef, TimeUnit as ArrowTimeUnit},
     error::ArrowError,
@@ -105,6 +105,9 @@ impl<S: Scalar> From<OwnedColumn<S>> for ArrayRef {
             }
             OwnedColumn::Scalar(_) => unimplemented!("Cannot convert Scalar type to arrow type"),
             OwnedColumn::VarChar(col) => Arc::new(StringArray::from(col)),
+            OwnedColumn::VarBinary(col) => {
+                Arc::new(BinaryArray::from_iter_values(col.iter().map(Vec::as_slice)))
+            }
             OwnedColumn::TimestampTZ(time_unit, _, col) => match time_unit {
                 PoSQLTimeUnit::Second => Arc::new(TimestampSecondArray::from(col)),
                 PoSQLTimeUnit::Millisecond => Arc::new(TimestampMillisecondArray::from(col)),
@@ -234,6 +237,15 @@ impl<S: Scalar> TryFrom<&ArrayRef> for OwnedColumn<S> {
                     .unwrap()
                     .iter()
                     .map(|s| s.unwrap().to_string())
+                    .collect(),
+            )),
+            DataType::Binary => Ok(Self::VarBinary(
+                value
+                    .as_any()
+                    .downcast_ref::<BinaryArray>()
+                    .unwrap()
+                    .iter()
+                    .map(|s| s.map(<[u8]>::to_vec).unwrap())
                     .collect(),
             )),
             DataType::Timestamp(time_unit, timezone) => match time_unit {
--- a/crates/proof-of-sql/src/base/database/column.rs
+++ b/crates/proof-of-sql/src/base/database/column.rs
@@ -52,6 +52,8 @@ pub enum Column<'a, S: Scalar> {
     /// - the second element maps to a timezone
     /// - the third element maps to columns of timeunits since unix epoch
     TimestampTZ(PoSQLTimeUnit, PoSQLTimeZone, &'a [i64]),
+    /// Variable length binary columns
+    VarBinary((&'a [&'a [u8]], &'a [S])),
 }
 
 impl<'a, S: Scalar> Column<'a, S> {
@@ -72,6 +74,7 @@ impl<'a, S: Scalar> Column<'a, S> {
             Self::TimestampTZ(time_unit, timezone, _) => {
                 ColumnType::TimestampTZ(*time_unit, *timezone)
             }
+            Self::VarBinary(..) => ColumnType::VarBinary,
         }
     }
     /// Returns the length of the column.
@@ -87,6 +90,10 @@ impl<'a, S: Scalar> Column<'a, S> {
             Self::Int(col) => col.len(),
             Self::BigInt(col) | Self::TimestampTZ(_, _, col) => col.len(),
             Self::VarChar((col, scals)) => {
+                assert_eq!(col.len(), scals.len());
+                col.len()
+            }
+            Self::VarBinary((col, scals)) => {
                 assert_eq!(col.len(), scals.len());
                 col.len()
             }
@@ -176,6 +183,17 @@ impl<'a, S: Scalar> Column<'a, S> {
                     alloc.alloc_slice_copy(scalars.as_slice()),
                 ))
             }
+            OwnedColumn::VarBinary(col) => {
+                let scalars = col
+                    .iter()
+                    .map(|b| S::from_byte_slice_via_hash(b))
+                    .collect::<Vec<_>>();
+                let bytes = col.iter().map(|s| s as &'a [u8]).collect::<Vec<_>>();
+                Column::VarBinary((
+                    alloc.alloc_slice_clone(&bytes),
+                    alloc.alloc_slice_copy(scalars.as_slice()),
+                ))
+            }
             OwnedColumn::TimestampTZ(tu, tz, col) => Column::TimestampTZ(*tu, *tz, col.as_slice()),
         }
     }
@@ -260,6 +278,14 @@ impl<'a, S: Scalar> Column<'a, S> {
         }
     }
 
+    /// Returns the column as a slice of strings and a slice of scalars if it is a varchar column. Otherwise, returns None.
+    pub(crate) fn as_varbinary(&self) -> Option<(&'a [&'a [u8]], &'a [S])> {
+        match self {
+            Self::VarBinary((col, scals)) => Some((col, scals)),
+            _ => None,
+        }
+    }
+
     /// Returns the column as a slice of i64 if it is a timestamp column. Otherwise, returns None.
     pub(crate) fn as_timestamptz(&self) -> Option<&'a [i64]> {
         match self {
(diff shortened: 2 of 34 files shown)


JayWhite2357 (commented):
Generally looks good.
Still needs tests, but I don't see anything obviously wrong.

crates/proof-of-sql/src/base/scalar/scalar.rs
```diff
@@ -32,6 +32,8 @@ pub trait Scalar:
     + for<'a> core::convert::From<&'a i64> // Required for `Column` to implement `MultilinearExtension`
     + for<'a> core::convert::From<&'a i128> // Required for `Column` to implement `MultilinearExtension`
     + for<'a> core::convert::From<&'a u8> // Required for `Column` to implement `MultilinearExtension`
+    + for<'a> core::convert::From<&'a [u8]> // Required for `Column` to implement `MultilinearExtension`
+    + for<'a>core::convert::From<&'a Vec<u8>> // Required for `Column` to implement `MultilinearExtension`
```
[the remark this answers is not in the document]
Dustin-Ray: ok will do

crates/proof-of-sql/src/base/arrow/record_batch_utility.rs
```diff
@@ -23,6 +23,19 @@ impl ToArrow for Vec<bool> {
...
+    }
+
+    fn to_array(self) -> arrow::array::ArrayRef {
+        use arrow::array::BinaryArray;
+        use Arc;
+        let array = BinaryArray::from_iter_values(self.iter().map(std::vec::Vec::as_slice));
+        Arc::new(array)
+    }
+}
+
```
JayWhite2357, line 38: Do we need this?
I want to eliminate this entire file.

crates/proof-of-sql/src/base/arrow/owned_and_arrow_conversions.rs
```diff
@@ -105,6 +105,10 @@ impl<S: Scalar> From<OwnedColumn<S>> for ArrayRef {
             }
             OwnedColumn::Scalar(_) => unimplemented!("Cannot convert Scalar type to arrow type"),
             OwnedColumn::VarChar(col) => Arc::new(StringArray::from(col)),
+            OwnedColumn::VarBinary(col) => {
+                let arr = BinaryArray::from_iter_values(col.iter().map(std::vec::Vec::as_slice));
+                Arc::new(arr)
```
JayWhite2357, line 110: Don't use `std`.
Instead add `use alloc::vec::Vec`.

NIT:
```suggestion
                Arc::new(BinaryArray::from_iter_values(col.iter().map(Vec::as_slice)))
```

crates/proof-of-sql/src/base/database/column.rs
```diff
@@ -176,6 +183,17 @@ impl<'a, S: Scalar> Column<'a, S> {
...
+                    .iter()
+                    .map(|b| S::from_le_bytes_mod_order(b))
+                    .collect::<Vec<_>>();
+                let bytes = col.iter().map(|s| s as &'a [u8]).collect::<Vec<_>>();
+                Column::VarBinary((
+                    alloc.alloc_slice_clone(&bytes),
+                    alloc.alloc_slice_copy(scalars.as_slice()),
+                ))
+            }
             OwnedColumn::TimestampTZ(tu, tz, col) => Column::TimestampTZ(*tu, *tz, col.as_slice()),
```
JayWhite2357, line 197: This is insecure because it is really easy to find collisions.
[0] is the same as [0,0,0,0,0] is the same as the le bytes of the field modulus as a "`u256`"  is the same as the le bytes of 2x the field modulus as a "`u256`".

Instead, we need to hash.

JayWhite2357 (changes requested):
Not finished reviewing, just wanted to put some of these comments out.

crates/proof-of-sql/src/base/scalar/scalar_ext.rs
```diff
@@ -30,9 +30,46 @@ pub trait ScalarExt: Scalar {
     fn into_u256_wrapping(self) -> U256 {
         U256::from(Into::<[u64; 4]>::into(self))
     }
+
+    /// Converts a slice of bytes into a `Scalar` by interpreting the bytes as
+    /// a little-endian number and taking it modulo the field’s modulus.
+    fn from_le_bytes_mod_order(bytes: &[u8]) -> Self;
```
JayWhite2357, line 36: For this PR, we need to have a method that is
```rust
fn from_byte_slice_via_hash(bytes: &[u8]) -> Self {
    let mut hasher = tiny_keccak::Keccak::v256();
    hasher.update(bytes);
    let mut hashed_bytes = [0u8; 32];
    hasher.finalize(&mut hashed_bytes);
    // mask hashed_bytes with CHALLENGE_MASK
    // convert hashed_bytes to S
}
```

Alternatively, create the method and just copy the code from
```rust
impl<T: MontConfig<4>> From<&[u8]> for MontScalar<T>
```
but I don't know if that will be any easier.

crates/proof-of-sql/src/base/database/group_by_util.rs
```diff
@@ -199,7 +202,7 @@ pub(crate) fn max_aggregate_column_by_index_counts<'a, S: Scalar>(
         }
         Column::Scalar(col) => max_aggregate_slice_by_index_counts(alloc, col, counts, indexes),
         // The following should never be reached because the `MAX` function can't be applied to varchar.
-        Column::VarChar(_) => {
+        Column::VarChar(_) | Column::VarBinary(_) => {
             unreachable!("MAX can not be applied to varchar")
```
JayWhite2357, line 206: Error message should be updated/not reused.

crates/proof-of-sql/src/base/arrow/record_batch_utility.rs
Dustin-Ray: Its used in arrow conversion tests. Removing this file is outside of the scope of this PR but can be done another time. Seems like its tightly integrated into arrow conversion testing.

crates/proof-of-sql/src/base/scalar/scalar_ext.rs
```diff
@@ -30,6 +32,29 @@ pub trait ScalarExt: Scalar {
...
+        // Convert to little-endian:
+        let limbs_le = [
+            u64::from_le(limbs_native[0]),
+            u64::from_le(limbs_native[1]),
+            u64::from_le(limbs_native[2]),
+            u64::from_le(limbs_native[3]),
+        ];
+
+        Self::from(limbs_le)
+    }
```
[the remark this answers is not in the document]
JayWhite2357: We also need to mask the hash so that it fits within the scalar. See https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/538#discussion_r1955135141, https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/468d6306778fee62b34dfcd0cb75b409d19d413d/crates/proof-of-sql/src/base/scalar/mont_scalar.rs#L144, and https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/468d6306778fee62b34dfcd0cb75b409d19d413d/crates/proof-of-sql/src/base/proof/transcript_core.rs#L64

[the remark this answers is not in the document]
JayWhite2357: Also, I like the idea of doing something like this:
https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/468d6306778fee62b34dfcd0cb75b409d19d413d/crates/proof-of-sql/src/base/scalar/mont_scalar.rs#L138-L140

Otherwise the empty slice maps to `0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`, which isn't really a "default".

crates/proof-of-sql/src/base/database/column_repetition_operation.rs
```diff
@@ -181,4 +203,60 @@ mod tests {
...
+
+    #[test]
+    fn test_column_repetition_op_varbinary() {
+        let bump = Bump::new();
+
+        let bytes = vec![b"foo".as_ref(), b"bar".as_ref()];
+        let scalars: Vec<TestScalar> = bytes
+            .iter()
+            .map(|b| TestScalar::from_le_bytes_mod_order(b))
+            .collect();
```
JayWhite2357, line 215: NIT: this should probably be a hash to reflect what is actually happining in practice.

Or, just do this
```suggestion
        let scalars = vec![TestScalar::from(1), TestScalar::from(2)];
```

crates/proof-of-sql/src/base/scalar/mont_scalar.rs
```diff
@@ -649,3 +649,13 @@ where
...
+
+impl<T> From<&Vec<u8>> for MontScalar<T>
+where
+    T: MontConfig<4>,
+    MontScalar<T>: Scalar,
+{
+    fn from(value: &Vec<u8>) -> Self {
+        MontScalar::from_le_bytes_mod_order(value.as_slice())
+    }
+}
```
JayWhite2357, line 661: We should remove this. There are multiple reasonable embeddings from bytes to a scalar. We shouldn't pick one of them as the "default".

crates/proof-of-sql/src/base/scalar/mod.rs
```diff
@@ -14,7 +14,7 @@ pub mod test_scalar;
 #[cfg(test)]
 mod test_scalar_test;
 
-mod scalar_ext;
+pub(crate) mod scalar_ext;
```
JayWhite2357, line 17: Why is this needed?

crates/proof-of-sql/src/base/slice_ops/inner_product.rs
```diff
@@ -26,3 +27,11 @@ where
         .map(|(a, b)| a.into() * *b)
         .sum()
 }
+
+/// Cannot use blanket impls for `Vec<u8>` because bytes might have different embeddings as scalars
+pub fn inner_product_with_bytes<S: ScalarExt>(a: &[Vec<u8>], b: &[S]) -> S {
```
JayWhite2357, line 32: It would be slightly nicer to be able to pass a function in for the conversion so you could do something like

```rust
inner_product_with(a, b, ScalarExt::from_byte_slice_via_hash)
```
that way, you can _probably_ make `inner_product_ref_cast` just be
```rust
inner_product_with(a, b, Into::into)
```
or something similar, resulting in less duplicate code.

JayWhite2357: Also, there are no tests for this.

JayWhite2357 (changes requested):

JayWhite2357 (changes requested):
Almost good to go.

crates/proof-of-sql/src/base/arrow/record_batch_utility.rs
Dustin-Ray: > I thought I had a comment on this... 

It was left as a review comment that was not resolvable and so I accidentally deleted it, my bad

crates/proof-of-sql/src/base/scalar/scalar_ext.rs
```diff
@@ -30,6 +31,26 @@ pub trait ScalarExt: Scalar {
...
+        if bytes.is_empty() {
+            return Self::zero();
+        }
+
+        let mut hasher = tiny_keccak::Keccak::v256();
+        hasher.update(bytes);
+        let mut hashed_bytes = [0u8; 32];
+        hasher.finalize(&mut hashed_bytes);
+        let hashed_val =
+            U256::from_le_slice(&hashed_bytes).expect("32 bytes => guaranteed to parse as U256");
```
JayWhite2357, line 50: NIT: we can avoid the expect (and the presumably slower checking required in `from_le_slice`)
```suggestion
        let mut hashed_bytes = [0u64; 4];
        hasher.finalize(hashed_bytes.as_mut_bytes());
        let hashed_val = U256::from(hashed_bytes );
```

JayWhite2357 (approved):

PR #546, opened 2025-02-13
feat(sol): add HyperKZG MSM helpers

JayWhite2357:
# Rationale for this change

The HyperKZG verifier can be broken down into smaller methods in order to facilitate more granular unit testing. This is the second (of two) PRs creating helper methods.

2 helper methods:
* `univariate_group_evaluation` to compute

$$ \sum_{i=0}^{m-1} e^i \mathbf{g}_i $$

* `compute_gl_msm` to compute

$$ g_L := \left(d^2 + d + 1\right)\cdot \left(\mathbf{C} + \sum_{i=0}^{m-1} q^{i+1}\cdot \mathbf{com}_i\right) + r\cdot\mathbf{w}_0 - dr \cdot\mathbf{w}_1 + (dr)^2\cdot\mathbf{w}_2 + b\cdot (-\mathbf{G}) $$

# Are these changes tested?
Yes

### DIFF
--- a/solidity/.solhint.json
+++ b/solidity/.solhint.json
@@ -6,6 +6,7 @@
         "no-global-import": "off",
         "no-inline-assembly": "off",
         "one-contract-per-file": "off",
-        "ordering": "off"
+        "ordering": "off",
+        "function-max-lines": "off"
     }
 }
--- a/solidity/src/hyperkzg/HyperKZGHelpers.pre.sol
+++ b/solidity/src/hyperkzg/HyperKZGHelpers.pre.sol
@@ -173,4 +173,81 @@ library HyperKZGHelpers {
             check_v_consistency(__v.offset, __r, __x, __y)
         }
     }
+
+    /// @notice Perform a multi-scalar multiplication of EC points using an evaluation scalar
+    /// @custom:as-yul-wrapper
+    /// #### Wrapped Yul Function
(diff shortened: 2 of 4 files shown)


iajoiner (dismissed):

solidity/.solhint.json
```diff
@@ -6,6 +6,7 @@
         "no-global-import": "off",
         "no-inline-assembly": "off",
         "one-contract-per-file": "off",
-        "ordering": "off"
+        "ordering": "off",
+        "function-max-lines": "off"
```
waelsy123, line 10: this could be useful to reduce the stake so we avoid --via-ir

waelsy123 (approved):

solidity/.solhint.json
JayWhite2357: You mean the stack?
Yes and no. The reason I turned it off is that there will be a decent number of methods that end up having more than 50% of the assembly block simply be import boilerplate.
For example, https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/546/files#diff-aa3fe37b7bb1dfcabed7a7cb222af92433c92df7a65ca4a690cf0fce32dc8c0eR285-R361 is huge.
It also happens to be nearing the stack limit because there are a ton of parameters, but the actual code is only 16 lines long.

PR #556, opened 2025-02-17
feat: add `logical_plans::Expr`

iajoiner:

# Rationale for this change
This PR is inspired by `DataFusion`.

We do expect to have proof and postprocessing nodes coexisting for a long time. Hence it is necessary to introduce `Expr` and `LogicalPlan` to make it easier to detect schema errors (e.g. `select not_a_column from tab` and `select * from tab where not_a_bool_col`) and separate the process of processing a query into logical nodes from routing them to proof and postprocessing nodes. 

- add `sql/logical_plans` directory
- add `logical_plans::Expr`
- add `LogicalPlanError`

# Are these changes tested?

Will be.

### DIFF
--- a/crates/proof-of-sql/src/sql/logical_plans/error.rs
+++ b/crates/proof-of-sql/src/sql/logical_plans/error.rs
@@ -1,13 +1,12 @@
-use alloc::string::String;
 use snafu::Snafu;
 
 /// Errors encountered during the process of converting `sqlparser::ast::Statement` to `LogicalPlan`
 #[derive(Debug, Snafu)]
 pub enum LogicalPlanError {
-    #[snafu(display("Unsupported feature: {:?}", message))]
-    /// Used when a feature parseable in `sqlparser` is not supported
-    Unsupported {
-        /// The unsupported feature
-        message: String,
+    #[snafu(display("Unsupported binary operator: {:?}", op))]
+    /// Used when a binary operator is not supported
+    UnsupportedBinaryOperator {
+        /// The unsupported binary operator
+        op: sqlparser::ast::BinaryOperator,
     },
 }
--- a/crates/proof-of-sql/src/sql/logical_plans/error.rs
+++ b/crates/proof-of-sql/src/sql/logical_plans/error.rs
@@ -1,0 +1,13 @@
+use alloc::string::String;
+use snafu::Snafu;
+
+/// Errors encountered during the process of converting `sqlparser::ast::Statement` to `LogicalPlan`
+#[derive(Debug, Snafu)]
+pub enum LogicalPlanError {
+    #[snafu(display("Unsupported feature: {:?}", message))]
+    /// Used when a feature parseable in `sqlparser` is not supported
+    Unsupported {
+        /// The unsupported feature
+        message: String,
+    },
+}
--- a/crates/proof-of-sql/src/sql/logical_plans/expr.rs
+++ b/crates/proof-of-sql/src/sql/logical_plans/expr.rs
@@ -1,0 +1,116 @@
+use super::LogicalPlanError;
+use crate::base::database::{ColumnRef, LiteralValue};
+use alloc::{boxed::Box, format};
+use serde::{Deserialize, Serialize};
+use sqlparser::ast::BinaryOperator as SqlBinaryOperator;
+
+/// Enum of column expressions that are either provable or supported in postprocessing
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum Expr {
+    /// Column
+    Column(ColumnRef),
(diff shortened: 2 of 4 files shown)


crates/proof-of-sql/src/sql/logical_plans/expr.rs
```diff
@@ -0,0 +1,116 @@
...
+use alloc::{boxed::Box, format};
+use serde::{Deserialize, Serialize};
+use sqlparser::ast::BinaryOperator as SqlBinaryOperator;
+
+/// Enum of column expressions that are either provable or supported in postprocessing
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum Expr {
+    /// Column
+    Column(ColumnRef),
+    /// Provable CONST expression
```
JayWhite2357, line 12: NIT:
```suggestion
    /// A constant expression
```

crates/proof-of-sql/src/sql/logical_plans/expr.rs
```diff
@@ -0,0 +1,116 @@
...
+            SqlBinaryOperator::Gt => Ok(Self::Gt),
+            SqlBinaryOperator::Lt => Ok(Self::Lt),
+            SqlBinaryOperator::GtEq => Ok(Self::Gte),
+            SqlBinaryOperator::LtEq => Ok(Self::Lte),
+            SqlBinaryOperator::And => Ok(Self::And),
+            SqlBinaryOperator::Or => Ok(Self::Or),
+            SqlBinaryOperator::Plus => Ok(Self::Add),
+            SqlBinaryOperator::Minus => Ok(Self::Sub),
+            SqlBinaryOperator::Multiply => Ok(Self::Mul),
+            SqlBinaryOperator::Divide => Ok(Self::Div),
```
JayWhite2357, line 72: Can we make all of these names the same? Is there any reason to name things differently from `sqlparser`?

(For the record, I do think `Plus` and `Minus` are weird names, but the upsides of consistency outweigh that I think.)

crates/proof-of-sql/src/sql/logical_plans/expr.rs
```diff
@@ -0,0 +1,116 @@
...
+            (SqlBinaryOperator::Or, BinaryOperator::Or),
+            (SqlBinaryOperator::Plus, BinaryOperator::Add),
+            (SqlBinaryOperator::Minus, BinaryOperator::Sub),
+            (SqlBinaryOperator::Multiply, BinaryOperator::Mul),
+            (SqlBinaryOperator::Divide, BinaryOperator::Div),
+        ];
+
+        for (sql_op, expected) in test_cases {
+            let result = BinaryOperator::try_from(sql_op).unwrap();
+            assert_eq!(result, expected,);
```
JayWhite2357, line 105: NIT:
```suggestion
            assert_eq!(result, expected);
```

crates/proof-of-sql/src/sql/logical_plans/expr.rs
```diff
@@ -0,0 +1,116 @@
+use super::LogicalPlanError;
+use crate::base::database::{ColumnRef, LiteralValue};
+use alloc::{boxed::Box, format};
+use serde::{Deserialize, Serialize};
+use sqlparser::ast::BinaryOperator as SqlBinaryOperator;
```
JayWhite2357, line 5: This could get confusing. Can you do this and use `ast::BinaryOperator` everywhere instead? It's almost the same length, and it avoids the `as` confusion.
```suggestion
use sqlparser::ast;
```

crates/proof-of-sql/src/sql/logical_plans/error.rs
```diff
@@ -0,0 +1,13 @@
...
+
+/// Errors encountered during the process of converting `sqlparser::ast::Statement` to `LogicalPlan`
+#[derive(Debug, Snafu)]
+pub enum LogicalPlanError {
+    #[snafu(display("Unsupported feature: {:?}", message))]
+    /// Used when a feature parseable in `sqlparser` is not supported
+    Unsupported {
+        /// The unsupported feature
+        message: String,
+    },
```
JayWhite2357, line 12: I'd much prefer to avoid `String`s in errors. `String`s require `alloc` at least. Pure `enum`s are cheaper and better. I'm rather of the opinion that an error variant should be used in exactly one place.
If needed, we can eventually created a sub-error that is called `Unsupported`, but that's excessive right now.

```suggestion
    UnsupportedBinaryOperator {
        /// The unsupported operator
        op: sqlparser::ast::BinaryOperator,
    },
```

JayWhite2357 (changes requested):

JayWhite2357 (approved):

PR #562, opened 2025-02-19
feat: add `logical_plans::LogicalPlan`

iajoiner:

Depends on the following PRs
- [x] #556 
Depends on #556
# Rationale for this change
Add `LogicalPlan`.

- add `LogicalPlan`

# Are these changes tested?

Will be.

### DIFF
--- a/crates/proof-of-sql/src/sql/logical_plans/plan.rs
+++ b/crates/proof-of-sql/src/sql/logical_plans/plan.rs
@@ -7,9 +7,9 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
 pub enum LogicalPlan {
     /// Empty
-    Empty,
+    Empty(Empty),
     /// Table scan
-    TableScan(TableRef),
+    TableScan(TableScan),
     /// Projection
     Projection(Projection),
     /// Filter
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql/src/sql/logical_plans/plan.rs
```diff
@@ -0,0 +1,128 @@
...
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+pub struct Join {
+    /// Left input plan
+    pub left: Box<LogicalPlan>,
+    /// Right input plan
+    pub right: Box<LogicalPlan>,
+    /// Equijoin condition
+    pub on: Vec<(Expr, Expr)>,
+    /// Join type
+    pub join_type: JoinType,
```
JayWhite2357, line 106: Since we only have one join type at the moment, I would think this could be added later.
The `struct` could be renamed `InnerJoin` if you want to clarify that it is `Inner`, or just left as `Join`.

crates/proof-of-sql/src/sql/logical_plans/plan.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/// Union
+/// e.g. SELECT a, b FROM t1 UNION ALL SELECT a, b FROM t2
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+pub struct Union {
+    /// Input plans
+    pub inputs: Vec<LogicalPlan>,
+    /// Schema of the output
+    pub schema: Vec<ColumnField>,
```
JayWhite2357, line 127: Can this be inferred from the inputs?
I would rather avoid redundant information in the plan.

crates/proof-of-sql/src/sql/logical_plans/plan.rs
```diff
@@ -0,0 +1,128 @@
...
+use alloc::{boxed::Box, vec::Vec};
+use serde::{Deserialize, Serialize};
+
+/// Enum of logical plans that are either provable or supported in postprocessing
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+pub enum LogicalPlan {
+    /// Empty
+    Empty,
+    /// Table scan
+    TableScan(TableRef),
```
JayWhite2357, line 12: Give these variants actual `structs`.
Although there are not traits right now, I would rather design this so that something like `enum_dispatch` could easily be used if needed.
```suggestion
    /// Empty
    Empty(Empty),
    /// Table scan
    TableScan(TableScan),
```

JayWhite2357 (changes requested):

crates/proof-of-sql/src/sql/logical_plans/plan.rs
iajoiner: Yes since in practice nobody is going to have a union of no plans

JayWhite2357 (approved):

PR #564, opened 2025-02-20
fix: update to Rust 2024 edition / Rust 1.85.0

iajoiner:

# Rationale for this change
Rust 1.85.0 & 2024 edition was just released.

- bump Rust edition to 2024
- bump Rust version to 1.85.0
- Automatic migration through `cargo fix --edition`
- replace instances of `gen()` with `r#gen()` due to `gen` becoming a new keyword in Rust 2024 edition

# Are these changes tested?

Yes.


iajoiner:
Let's not bump Rust edition just yet due to edition and MSRV cascading downstream

Dustin-Ray:
I was going to comment that I dont really feel comfortable approving this right now, but I think you closed it for the exact reasons that I was going to enumerate. I would want to see more ownership over the downstream effects of a PR like before approving it. It would need to be done in concert with all of our other services and tbh its not even clear that we need this right now. Thanks for the effort but I dont think its a great move to just do it here and expect all of the downstream dependents to catch up, we need a comprehensive strategy for a large change like this.

PR #575, opened 2025-02-24
feat(sol): add literal expression

JayWhite2357:
# Rationale for this change

We need to add the SQL nodes to the solidity verifier.

* Added a `case_const` function to "compile-time" assert that two constants are equal. This is because Yul does not support constants in case statements.
* Added `literal_expr_evaluate` function.

# Are these changes tested?
Yes

### DIFF
--- a/solidity/src/proof_exprs/LiteralExpr.pre.sol
+++ b/solidity/src/proof_exprs/LiteralExpr.pre.sol
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: UNLICENSED
+// This is licensed under the Cryptographic Open Software License 1.0
+pragma solidity ^0.8.28;
+
+import "../base/Constants.sol";
+import "../base/Errors.sol";
+
+/// @title LiteralExpr
+/// @dev Library for handling literal expressions
+library LiteralExpr {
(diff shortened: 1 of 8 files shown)


solidity/src/proof_exprs/LiteralExpr.pre.sol
```diff
@@ -0,0 +1,72 @@
...
+    /// ##### Parameters
+    /// * `expr_ptr_in` - the calldata pointer to the beginning of the expression data
+    /// * `chi_in_eval` - the chi value for evaluation
+    /// ##### Return Values
+    /// * `expr_ptr` - the pointer to the remaining expression after consuming the literal expression
+    /// * `eval` - the evaluated result
+    /// @dev This function evaluates a literal expression by multiplying the literal value by chi_in_eval.
+    /// This is because `chi_in_eval` is the evaluation of a column of ones of the appropriate length.
+    /// @param __expr The literal expression data
+    /// @param __chiInEval The chi value for evaluation
```
iajoiner, line 31: maybe `chi_eval`? Also `expr_ptr_out` would be helpful

iajoiner (approved):

PR #576, opened 2025-02-24
feat(sol): add column expression

JayWhite2357:
# Rationale for this change

We need to add the SQL nodes to the solidity verifier.

* Added column evaluations to the verification builder. This diverges slightly from the Rust code, but reduces the stack size, which is critical in the EVM.
* Added `column_expr_evaluate`.

# Are these changes tested?
Yes

### DIFF
--- a/solidity/src/base/Constants.sol
+++ b/solidity/src/base/Constants.sol
@@ -84,7 +84,7 @@ uint256 constant G2_NEG_GEN_Y_REAL = 0x1d9befcd05a5323e6da4d435f3b617cdb3af83285
 uint256 constant G2_NEG_GEN_Y_IMAG = 0x275dc4a288d1afb3cbb1ac09187524c7db36395df7be3b99e673b13a075a65ec;
 
 /// @dev Size of the verification builder in bytes.
-uint256 constant VERIFICATION_BUILDER_SIZE = 0x20 * 9;
+uint256 constant VERIFICATION_BUILDER_SIZE = 0x20 * 10;
 /// @dev Offset of the pointer to the challenge queue in the verification builder.
 uint256 constant BUILDER_CHALLENGES_OFFSET = 0x20 * 0;
 /// @dev Offset of the pointer to the first round MLEs in the verification builder.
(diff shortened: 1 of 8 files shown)


solidity/src/base/Constants.sol
```diff
@@ -28,6 +28,11 @@ uint256 constant UINT32_SIZE = 0x04;
 /// @dev Number of bits needed to pad uint32 to 256 bits
 /// @dev This is useful for shifting a uint256 to the right to extract a uint32
 uint256 constant UINT32_PADDING_BITS = 0xE0;
+/// @dev Size of uint64 in bytes
+uint256 constant UINT64_SIZE = 0x08;
```
iajoiner, line 32: Do we need UINT64_SIZE when there is INT64_SIZE?

solidity/src/proof/VerificationBuilder.pre.sol
```diff
@@ -464,4 +465,59 @@ library VerificationBuilder {
...
+    /// ##### Signature
+    /// ```yul
+    /// builder_set_column_evaluations(builder_ptr, values_ptr)
+    /// ```
+    /// ##### Parameters
+    /// * `builder_ptr` - memory pointer to the builder struct region
+    /// * `values_ptr` - pointer to the array in memory
+    /// @param __builder The builder struct
+    /// @param __values The column evaluation values array
+    function __setColumnEvaluations(Builder memory __builder, uint256[] memory __values) internal pure {
```
iajoiner, line 481: Hmm do we intend to set evaluations for all columns for all tables at once?

JayWhite2357: Yeah. This is essentially the equivalent of this: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/ffb26c778caa4b7f12d19a069a7b2c2d50a80a24/crates/proof-of-sql/src/sql/proof/query_proof.rs#L417-L420.

iajoiner (approved):

PR #578, opened 2025-02-25
feat: support Uint16 type

Dustin-Ray:
# Rationale for this change

Introduces support for additional unsigned integer type uint 16. The main need for this is found in the range check protocol which is found under certain conditions to operate significantly better with a wider word size. This type is also useful in other database contexts.

- [x] Standard initial support procedure for u16 type

# Are these changes tested?
- [x] existing tests currently passing, more tests can be added depending on codecov report


Dustin-Ray:
@iajoiner, @JayWhite2357  says we dont need this type anymore. Let me know what to do with this. We needed the u16 type for commitable columns using u16s like in range check. It sounds like maybe we want to reduce the number of types we have eventually?

iajoiner:
Sorry we don't need this type. Closed.

PR #583, opened 2025-02-27
feat!: remove restrictions that at least one non-empty input table exists from `sql/proof`

iajoiner:

# Rationale for this change
Tableless queries such as `select 'not_a_column', 0 as def_int;` are legitimate and can have non-trivial results. However we currently force the results to be empty since we currently don't allow tableless queries in `sql/parse` any way. Now this needs to be fixed in order for such queries to be allowed and return correct results. Note that we aren't going to enable it in `sql/parse` since it is legacy code that will be removed very soon but we will in the planner.

- remove restriction that at least one non-empty input table exists

# Are these changes tested?

Existing tests should pass.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof/verifiable_query_result_test.rs
+++ b/crates/proof-of-sql/src/sql/proof/verifiable_query_result_test.rs
@@ -111,22 +111,3 @@ fn we_can_verify_queries_on_an_empty_table() {
     let expected_res = owned_table([bigint("a1", [0; 0])]);
     assert_eq!(table, expected_res);
 }
-
-#[test]
-fn empty_verification_fails_if_the_result_contains_non_null_members() {
-    let expr = EmptyTestQueryExpr {
-        columns: 1,
-        ..Default::default()
-    };
-    let accessor = OwnedTableTestAccessor::<InnerProductProof>::new_from_table(
-        TableRef::new("sxt", "test"),
-        owned_table([bigint("a1", [0_i64; 0])]),
-        0,
-        (),
-    );
-    let res = VerifiableQueryResult::<InnerProductProof> {
-        result: Some(owned_table([])),
-        proof: None,
-    };
-    assert!(res.verify(&expr, &accessor, &()).is_err());
-}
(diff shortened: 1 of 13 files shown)


crates/proof-of-sql/src/sql/proof/verifiable_query_result_test.rs
```diff
@@ -111,22 +111,3 @@ fn we_can_verify_queries_on_an_empty_table() {
     let expected_res = owned_table([bigint("a1", [0; 0])]);
     assert_eq!(table, expected_res);
 }
-
-#[test]
-fn empty_verification_fails_if_the_result_contains_non_null_members() {
```
stuarttimwhite, line 116: How difficult would it be to replace this with an appropriate passing test? Does that even make sense in this case?

crates/proof-of-sql/src/sql/proof_gadgets/membership_check_test.rs
```diff
@@ -447,10 +447,8 @@ mod tests {
             source_columns: vec![],
             candidate_columns: vec![],
         };
-        let verifiable_res = VerifiableQueryResult::<InnerProductProof>::new(&plan, &accessor, &());
-        let actual = verifiable_res.verify(&plan, &accessor, &()).unwrap().table;
-        let expected = owned_table([int128("multiplicities", [0_i128; 0])]);
-        assert_eq!(actual, expected);
+        let _verifiable_res =
+            VerifiableQueryResult::<InnerProductProof>::new(&plan, &accessor, &());
```
stuarttimwhite, line 451: Is this actual testing anything without `verify` being called?

iajoiner: Yes. It fundamentally makes no sense for membership check or permutation (used for inner joins and in the future ordering) if there are no columns.

crates/proof-of-sql/src/sql/proof/verifiable_query_result_test.rs
iajoiner: @stuarttimwhite We used to check that `proof` and `result` need to be both `Some` or `None`. Since these two fields are no longer optional and there is already a test which checks that verification works for empty tables there is no need for this test any more. @JayWhite2357 please correct me if I'm wrong.

stuarttimwhite (approved):

PR #586, opened 2025-03-03
build: explicitly install the active toolchain in ci because rustup 1.28.0 no longer does so by default

iajoiner:

# Rationale for this change
In [Rustup 1.28.0](https://blog.rust-lang.org/2025/03/02/Rustup-1.28.0.html) which was released yesterday it is no longer true that the active toolchain is automatically installed. Hence we need to install it explicitly in CI or CI breaks.

Really thanks @tlovell-sxt for finding the actual source of the problem!

- adjust CI so that relevant toolchains are manually installed

# Are these changes tested?

N/A

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -33,7 +33,7 @@ jobs:
             target/
           key: ${{ runner.os }}-cargo-check-${{ hashFiles('**/Cargo.toml') }}
       - name: Install stable toolchain
-        run: curl https://sh.rustup.rs -sSf | bash -s -- -y --profile minimal && source ~/.cargo/env
+        run: curl https://sh.rustup.rs -sSf | bash -s -- -y --profile minimal && source ~/.cargo/env && rustup toolchain install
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-and-test.yml
```diff
@@ -211,7 +211,7 @@ jobs:
         uses: actions/checkout@v3
       - name: Install stable toolchain
         run: |
-          curl https://sh.rustup.rs -sSf | bash -s -- -y --profile minimal && source ~/.cargo/env
+          curl https://sh.rustup.rs -sSf | bash -s -- -y --profile minimal && source ~/.cargo/env && rustup toolchain install
           rustup component add rustfmt
       - name: Run cargo fmt
```
tlovell-sxt, line 216: I believe udeps needs a change too

tlovell-sxt: `rustup toolchain install nightly`

tlovell-sxt (approved):
lgtm

PR #580, opened 2025-02-26
feat: add `sql_to_proof_plans`

iajoiner:
Please be sure to look over the pull request guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md#submit-pr.

# Please go through the following checklist
- [x] The PR title and commit messages adhere to guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md. In particular `!` is used if and only if at least one breaking change has been introduced.
- [x] I have run the ci check script with `source scripts/run_ci_checks.sh`.

# Rationale for this change
In order to take advantage of name resolution and logical plan from datafusion we would like to add a new planner crate.

# What changes are included in this PR?
- add `sql_to_proof_plans`

# Are these changes tested?

Will be.

PR #596, opened 2025-03-05
feat: add minimal `proof-of-sql-planner` crate && relevant CI change

iajoiner:

# Rationale for this change
In order to take advantage of name resolution and logical plan from datafusion we would like to add a new planner crate. This PR is a part of #580 since #580 is getting large.

- add `proof-of-sql-planner` crate with the following
  - `PlannerError` and `PlannerResult`
  - `PoSqlContextProvider` and `PoSqlTableSource`
  - utils that perform datafusion <-> PoSql conversions
- let CI check for `no_std` for `proof-of-sql` and `proof-of-sql-parser` crates since `proof-of-sql-planner` which uses `datafusion` requires `std`

# Are these changes tested?

Yes.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -72,10 +72,11 @@ jobs:
         run: |
           cargo check -p proof-of-sql --no-default-features --features="std"
           cargo check -p proof-of-sql --all-targets --no-default-features --features="std"
-      - name: Run cargo check (proof-of-sql-parser) with no_std target.
+      - name: Run cargo check (proof-of-sql-parser & proof-of-sql) with no_std target.
         run: |
           rustup target add thumbv7em-none-eabi
-          cargo check --target thumbv7em-none-eabi --no-default-features
+          cargo check -p proof-of-sql-parser --target thumbv7em-none-eabi --no-default-features
+          cargo check -p proof-of-sql --target thumbv7em-none-eabi --no-default-features
 
   test:
     name: Test Suite
--- a/crates/proof-of-sql-planner/Cargo.toml
+++ b/crates/proof-of-sql-planner/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "proof-of-sql-planner"
+edition.workspace = true
+exclude.workspace = true
+repository.workspace = true
+version.workspace = true
+license-file.workspace = true
+
+[dependencies]
+arrow = { workspace = true }
+bumpalo = { workspace = true }
+datafusion = '38.0.0'
+indexmap = { workspace = true }
+proof-of-sql = { path = "../proof-of-sql", features = ["arrow"] }
+snafu = { workspace = true }
+sqlparser = { workspace = true }
+
+[lints]
+workspace = true
--- a/crates/proof-of-sql-planner/Cargo.toml
+++ b/crates/proof-of-sql-planner/Cargo.toml
@@ -8,12 +8,14 @@ license-file.workspace = true
 
 [dependencies]
 arrow = { workspace = true }
-bumpalo = { workspace = true }
 datafusion = '38.0.0'
 indexmap = { workspace = true }
 proof-of-sql = { path = "../proof-of-sql", features = ["arrow"] }
 snafu = { workspace = true }
 sqlparser = { workspace = true }
+
+[dev-dependencies]
+bumpalo = { workspace = true }
 
 [lints]
 workspace = true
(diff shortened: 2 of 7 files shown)


.github/workflows/lint-and-test.yml
```diff
@@ -72,10 +72,11 @@ jobs:
         run: |
           cargo check -p proof-of-sql --no-default-features --features="std"
           cargo check -p proof-of-sql --all-targets --no-default-features --features="std"
-      - name: Run cargo check (proof-of-sql-parser) with no_std target.
+      - name: Run cargo check (proof-of-sql-parser & proof-of-sql) with no_std target.
         run: |
           rustup target add thumbv7em-none-eabi
-          cargo check --target thumbv7em-none-eabi --no-default-features
+          cargo check -p proof-of-sql-parser --target thumbv7em-none-eabi --no-default-features
+          cargo check -p proof-of-sql --target thumbv7em-none-eabi --no-default-features
```
JayWhite2357, line 79: Can you create a GitHub issue that adds a wasm target for this crate.
We can add it after the initial dev work.
It's not a big rush, so it could be good for an external contributor/bounty.

crates/proof-of-sql-planner/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "proof-of-sql-planner"
+edition.workspace = true
+exclude.workspace = true
+repository.workspace = true
+version.workspace = true
+license-file.workspace = true
+
+[dependencies]
+arrow = { workspace = true }
+bumpalo = { workspace = true }
```
JayWhite2357, line 11: This can be a dev dependency

crates/proof-of-sql-planner/src/util.rs
```diff
@@ -0,0 +1,243 @@
...
+            TableRef::from_names(Some(schema), table)
+        }
+    }
+}
+
+/// Convert a [`ScalarValue`] to a [`LiteralValue`]
+///
+/// TODO: add other types supported in `PoSQL`
+#[allow(dead_code)]
+pub(crate) fn scalar_value_to_literal_value(value: &ScalarValue) -> PlannerResult<LiteralValue> {
```
JayWhite2357, line 25: Since this is always copied/cloned, let's take ownership of it. Users of the function can clone if needed.
```suggestion
pub(crate) fn scalar_value_to_literal_value(value: ScalarValue) -> PlannerResult<LiteralValue> {
```

crates/proof-of-sql-planner/src/util.rs
```diff
@@ -0,0 +1,243 @@
...
+use arrow::datatypes::{Field, Schema};
+use datafusion::{
+    catalog::TableReference,
+    common::{Column, DFSchema, ScalarValue},
+};
+use proof_of_sql::base::database::{ColumnField, ColumnRef, ColumnType, LiteralValue, TableRef};
+
+/// Convert a [`TableReference`] to a [`TableRef`]
+///
+/// If catalog is provided it is ignored
```
JayWhite2357, line 11: What about this query: `(SELECT * FROM foo.bar.baz) UNION (SELECT * FROM qux.bar.baz)`?  That's what catelog means, right?
If a user provides a catalog, I'd rather give them an error than silently drop it, so I feel like this should return a `Result` and the `TableReference::Full` should error.

crates/proof-of-sql-planner/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! This crate converts a `DataFusion` `LogicalPlan` to a `ProofPlan` and `Postprocessing`
```
JayWhite2357, line 1: This would allow you to remove the `allow`s everywhere else, and makes it easier to deal with in the future.
```suggestion
//! This crate converts a `DataFusion` `LogicalPlan` to a `ProofPlan` and `Postprocessing`
#![cfg_attr(not(test), expect(dead_code))] // TODO: remove this when initial development work is done
#![cfg_attr(test, expect(clippy::missing_panics_doc))]
```

JayWhite2357 (changes requested):
Looks good. Only a few minor requests.

.github/workflows/lint-and-test.yml
iajoiner: Sure! https://github.com/spaceandtimelabs/sxt-proof-of-sql/issues/597

JayWhite2357 (approved):

PR #603, opened 2025-03-06
feat: convert `LogicalPlan` to `ProofPlan`

iajoiner:

# Rationale for this change
I'm adding a few basic `LogicalPlan`s and convert them to `ProofPlan`s to get end to end running ASAP.

- adds conversion for `EmptyRelation` and `TableScan` whenever there is no fetch limit
- makes `AliasedDynProofExpr`, `ProofExpr` and `TableExpr` pub

# Are these changes tested?

Yes.

### DIFF
--- a/crates/proof-of-sql-planner/src/plan.rs
+++ b/crates/proof-of-sql-planner/src/plan.rs
@@ -23,14 +23,11 @@ use proof_of_sql::{
 /// However that shouldn't be taken for granted.
 fn get_aliased_dyn_proof_exprs(
     table_ref: &TableRef,
-    projection: Option<Vec<usize>>,
+    projection: &[usize],
     input_schema: &DFSchema,
     output_schema: &DFSchema,
 ) -> PlannerResult<Vec<AliasedDynProofExpr>> {
-    let num_input_columns = input_schema.columns().len();
-    let projection_indexes =
-        projection.unwrap_or_else(|| (0..num_input_columns).collect::<Vec<_>>());
-    projection_indexes
+    projection
         .iter()
         .enumerate()
         .map(
@@ -59,7 +56,7 @@ fn get_aliased_dyn_proof_exprs(
 fn table_scan_to_projection(
     table_name: &TableReference,
     schemas: &IndexMap<TableReference, DFSchema>,
-    projection: Option<Vec<usize>>,
+    projection: &[usize],
     projected_schema: &DFSchema,
 ) -> PlannerResult<DynProofPlan> {
     // Check if the table exists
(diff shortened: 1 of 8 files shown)


crates/proof-of-sql-planner/src/plan.rs
```diff
@@ -0,0 +1,472 @@
...
+    table_ref: &TableRef,
+    projection: Option<Vec<usize>>,
+    input_schema: &DFSchema,
+    output_schema: &DFSchema,
+) -> PlannerResult<Vec<AliasedDynProofExpr>> {
+    let num_input_columns = input_schema.columns().len();
+    let projection_indexes =
+        projection.unwrap_or_else(|| (0..num_input_columns).collect::<Vec<_>>());
+    projection_indexes
+        .iter()
```
JayWhite2357, line 34: Avoid collect.
```suggestion
    let num_extra_columns = match projection {
        Some(_) => 0,
        None => input_schema.columns().len(),
    };
    let projection_indexes = projection.into_iter().flatten().chain(0..num_extra_columns);
    projection_indexes
```

crates/proof-of-sql-planner/src/plan.rs
```diff
@@ -0,0 +1,472 @@
...
+            filters,
+            fetch: None,
+            ..
+        }) if filters.is_empty() => {
+            table_scan_to_projection(table_name, schemas, projection.clone(), projected_schema)
+        }
+        // Filter but no fetch limit
+        LogicalPlan::TableScan(TableScan {
+            table_name,
+            projection,
```
JayWhite2357, line 147: Can we assume this if the LogicalPlan is resolved?
```suggestion
            projection: Some(proj),
```
If we won't ever hit the `None` case in production, I would rather not have code handling it. (Or add it later.)

crates/proof-of-sql-planner/src/plan.rs
```diff
@@ -0,0 +1,472 @@
...
+            filters,
+            fetch: None,
+            ..
+        }) if !filters.is_empty() => table_scan_to_filter(
+            table_name,
+            schemas,
+            projection.clone(),
+            projected_schema,
+            filters,
+        ),
```
JayWhite2357, line 158: I know I said the opposite before, but now that there is a helper function for the two cases, I think this is more readable.
(This is only because we handle both cases completely.)
```suggestion
        }) => {
            if filters.is_empty() {
                table_scan_to_projection(table_name, schemas, projection.clone(), projected_schema)
            } else {
                table_scan_to_filter(
                    table_name,
                    schemas,
                    projection.clone(),
                    projected_schema,
                    filters,
                )
            }
        }
```

crates/proof-of-sql-planner/src/plan.rs
```diff
@@ -0,0 +1,472 @@
...
+            },
+        )
+        .collect::<PlannerResult<Vec<_>>>()
+}
+
+/// Convert a `TableScan` without filters or fetch limit to a `DynProofPlan`
+fn table_scan_to_projection(
+    table_name: &TableReference,
+    schemas: &IndexMap<TableReference, DFSchema>,
+    projection: Option<Vec<usize>>,
```
JayWhite2357, line 62: Passing as a reference will avoid clones, and it is only ever iterated over, so ownership does nothing.
```suggestion
    projection: &Option<Vec<usize>>,
```

crates/proof-of-sql-planner/src/plan.rs
iajoiner: @JayWhite2357 Sure. There are actually two more cases I'm adding in #608 as well. Datafusion adds a `fetch` to `TableScan` if there is any `LIMIT` clause which in almost all cases very significantly restrict the size of a lot of compute. That's also why I have these in functions haha.

JayWhite2357 (approved):

PR #606, opened 2025-03-06
feat: add div and mod gadget

stuarttimwhite:

# Rationale for this change

A new gadget is needed in order to support the div and mod proof expressions.

A new div and mod gadget.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/base/database/column.rs
+++ b/crates/proof-of-sql/src/base/database/column.rs
@@ -426,9 +426,7 @@ impl ColumnType {
     /// Returns `None` if the type is not a signed integer.
     /// `sqrt_negative_min(NumericalType) = floor(sqrt(-NumericalType::MIN))`
     #[must_use]
-    #[cfg_attr(not(test), expect(dead_code))]
-    #[expect(clippy::trivially_copy_pass_by_ref)]
-    fn sqrt_negative_min(&self) -> Option<u64> {
+    pub(crate) fn sqrt_negative_min(self) -> Option<u64> {
         match self {
             ColumnType::TinyInt => Some(11),
             ColumnType::SmallInt => Some(181),
--- a/crates/proof-of-sql/src/base/database/column.rs
+++ b/crates/proof-of-sql/src/base/database/column.rs
@@ -440,7 +440,7 @@ impl ColumnType {
     }
 
     /// Returns the number of bits in the integer type if it is an integer type. Otherwise, return None.
-    fn to_integer_bits(self) -> Option<usize> {
+    pub(crate) fn to_integer_bits(self) -> Option<usize> {
         match self {
             ColumnType::Uint8 | ColumnType::TinyInt => Some(8),
             ColumnType::SmallInt => Some(16),
--- a/crates/proof-of-sql/src/base/database/column_type_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_type_operation.rs
@@ -115,8 +115,7 @@ pub fn try_multiply_column_types(
 /// Determine the output type of a divide/modulo operation if it is possible
 /// to divide and modulo the two input types. If the types are not compatible, return
 /// an error.
-#[cfg_attr(not(test), expect(dead_code))]
-fn try_divide_modulo_column_types(
+pub fn try_divide_modulo_column_types(
     lhs: ColumnType,
     rhs: ColumnType,
 ) -> ColumnOperationResult<(ColumnType, ColumnType)> {
--- a/crates/proof-of-sql/src/sql/proof_gadgets/divide_and_modulo_expr.rs
+++ b/crates/proof-of-sql/src/sql/proof_gadgets/divide_and_modulo_expr.rs
@@ -1,0 +1,189 @@
+use crate::{
+    base::{
+        database::{Column, ColumnRef, Table},
+        map::IndexMap,
+        proof::ProofError,
+        scalar::Scalar,
+    },
+    sql::{
+        proof::{FinalRoundBuilder, VerificationBuilder},
+        proof_exprs::{
+            divide_columns, modulo_columns,
+            DynProofExpr, ProofExpr,
+        },
+    },
+    utils::log,
+};
+use alloc::boxed::Box;
+use bumpalo::Bump;
+use serde::{Deserialize, Serialize};
+
+/// A gadget for proving divide and modulo expressions in tandem.
+/// They must be proved in tandem under this protocol.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct DivideAndModuloExpr {
+    pub lhs: Box<DynProofExpr>,
+    pub rhs: Box<DynProofExpr>,
+}
+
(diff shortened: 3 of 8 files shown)


crates/proof-of-sql/src/base/database/column_type_operation.rs
```diff
@@ -750,4 +777,46 @@ mod test {
...
+            eligible_columns_enumerated.clone(),
+            eligible_columns_enumerated
+        ) {
+            let remainder = try_divide_modulo_column_types(numerator, denominator).unwrap();
+            let expected_remainder = if di > ni { denominator } else { numerator };
+            assert_eq!(remainder, (numerator, expected_remainder));
+        }
+        let ineligible_columns = [
+            ColumnType::Uint8,
+            ColumnType::Scalar,
```
Dustin-Ray, line 801: I would imagine that unsigned integers and scalars would support division

crates/proof-of-sql/src/sql/proof_gadgets/divide_and_modulo_expr.rs
```diff
@@ -0,0 +1,1053 @@
...
+        quotient: &'a [S],
+        rhs: Vec<S>,
+        min_sqrt_scalar: S,
+    ) -> &'a [S] {
+        let in_range_q_or_b = alloc.alloc_slice_fill_with(quotient.len(), |_i| S::ZERO);
+        for (res, (q, b)) in in_range_q_or_b
+            .iter_mut()
+            .zip(quotient.iter().copied().zip(rhs.clone()))
+        {
+            // We do or rather than and here because scalars wrap negative values, so only one can be true at a time
```
Dustin-Ray, line 87:
```suggestion
            // We do `or` rather than `and` here because scalars wrap negative values, so only one can be true at a time
```

crates/proof-of-sql/src/sql/proof_gadgets/divide_and_modulo_expr.rs
```diff
@@ -0,0 +1,1053 @@
...
+                b
+            };
+            *res = in_range_value;
+        }
+        in_range_q_or_b
+    }
+}
+
+impl DivideAndModuloExpr {
+    #[cfg_attr(not(test), expect(dead_code))]
```
Dustin-Ray, line 100: the purpose of this attribute is unclear. If this is never going to be used for anything, please dont add it.

crates/proof-of-sql/src/sql/proof_gadgets/divide_and_modulo_expr.rs
```diff
@@ -0,0 +1,1053 @@
...
+        Self { lhs, rhs }
+    }
+
+    #[allow(clippy::missing_panics_doc)]
+    fn min_scalar<S: Scalar>(&self) -> S {
+        self.lhs.data_type().min_scalar::<S>().unwrap()
+    }
+
+    #[allow(clippy::missing_panics_doc)]
+    pub fn data_type(&self) -> (ColumnType, ColumnType) {
```
Dustin-Ray, line 111: Its not immediately clear that this needs to be pub, I think you can keep this local

Dustin-Ray (changes requested):
In general this looks really nice. Thanks for the detailed docs comments on some of the functions, and for the comments throughout the gadget in general, it greatly aids with understanding.

My overall comment is that you can avoid introducing dead code by actually constructing the gadget in a ```div_mod_test_expr.rs``` file that lives adjacent to the ```divide_and_modulo_expr.rs``` file in the gadgets module. This will help you avoid introducing dead code into the repo which are then suppressed with clippy lints, and this also follows the convention of the rest of the gadgets in that module.

This also gives you the ability to create a complete test setup which could demonstrate the usage of different commitment schemes. I dont mind the usage of mock here but it does hide a lot of testing functionality that would be nice to see.

crates/proof-of-sql/src/base/database/column_type_operation.rs
stuarttimwhite: I think that is the plan for the future, in addition to decimals, but the constraints will be different for those, so this is just a decent starting point.

crates/proof-of-sql/src/sql/proof_gadgets/divide_and_modulo_expr.rs
stuarttimwhite: This particular attribute is the same as `#[expect(dead_code)]`, except that it allows test code to call it. Any functions that are added to this gadget will be used by `DivExpr`, and `ModExpr`, which will be coming in my next PR.

crates/proof-of-sql/src/sql/proof_gadgets/divide_and_modulo_expr.rs
stuarttimwhite: This will be used in the `data_type` function for the `ProofExpr` implementation of `DivExpr` and `ModExpr`. I'm happy to change this to private for now if that's generally better practice.

Dustin-Ray: If its not needed to be pub yet, lets make it private until we need it to be pub, that way we cant forget about it and leave it exposed by accident. I think there are a few more functions in this PR that could use examination for this as well.

crates/proof-of-sql/src/base/database/column.rs
```diff
@@ -422,8 +422,22 @@ impl ColumnType {
         )
     }
 
+    /// Returns the floor of the sqrt of the negative min integer
+    /// Returns `None` if the type is not a signed integer.
```
iajoiner, line 426: You probably want to provide the formula as well as an example 
e.g. `f(NumericalType) := floor(sqrt(-NumericalType::MIN))`

crates/proof-of-sql/src/base/database/column_type_operation.rs
```diff
@@ -112,6 +112,31 @@ pub fn try_multiply_column_types(
...
+/// an error.
+pub fn try_divide_modulo_column_types(
+    lhs: ColumnType,
+    rhs: ColumnType,
+) -> ColumnOperationResult<(ColumnType, ColumnType)> {
+    if lhs.is_integer() && lhs.is_signed() && rhs.is_integer() && rhs.is_signed() {
+        Ok((
+            lhs,
+            if lhs.byte_size() > rhs.byte_size() {
+                lhs
```
iajoiner, line 126: Hmm maybe we should exchange 126 and 128? BigInt % Int = Int and Int % BigInt = Int, right?

stuarttimwhite: After thinking about it I think it'll always be the numerator/left type (for now). For this expression, `BigInt::MAX % 0` is actually evaluated as `BigInt::MAX`, for example. And `-1 % BigInt::MAX` evaluates to -1, not `BigInt::MAX - 1`. So `BigInt % Int = BigInt` and `Int % BigInt = Int`.

PR #608, opened 2025-03-06
 feat: convert `LogicalPlan` variants `Limit`, `Projection`, `Union` and `TableScan` with `fetch` to `ProofPlan`

iajoiner:

# Rationale for this change
Now that we have basic `LogicalPlan` -> `ProofPlan` we should support more features. Moreover a bug has been found in the code which is to be fixed and with a test checking for it.

- add `TableScan` with fetch
- add `Limit`, `Projection` and `Union`
- fix one error in `table_scan_to_projection` which accidentally uses output rather than input `ColumnField`s

# Are these changes tested?

Will be.

### DIFF
--- a/crates/proof-of-sql-planner/src/plan.rs
+++ b/crates/proof-of-sql-planner/src/plan.rs
@@ -1,15 +1,18 @@
-use super::{expr_to_proof_expr, table_reference_to_table_ref, PlannerError, PlannerResult};
+use super::{
+    df_schema_to_column_fields, expr_to_proof_expr, table_reference_to_table_ref, PlannerError,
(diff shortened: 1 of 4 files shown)


crates/proof-of-sql-planner/src/plan.rs
```diff
@@ -130,15 +128,57 @@ pub fn logical_plan_to_proof_plan(
             projection: Some(projection),
             projected_schema,
             filters,
-            fetch: None,
+            fetch,
```
JayWhite2357, line 131: Is this ever `Some`? Could we configure the optimizer to not push down limits?

PR #609, opened 2025-03-06
test(sol): fix flaky tests

JayWhite2357:
# Rationale for this change

Some of the fuzz tests rely on random sampling not choosing very low probability samples. However, the foundry fuzz tests are not fully random, so some edge cases need to be filtered out. Some tests do not filter enough.

* Fixed flakiness in `testFuzzECPairingX2WithValidInputsThatDoNotSumToZero`: if a+b = 0 as field elements, the test will/should fail. This is now filtered out.
* Fixed flakiness in `testFuzzRevertsVConsistency`. If r = 0 and x[i] = 0, the test will/should fail. This is now filtered out.

# Are these changes tested?
NA

### DIFF
--- a/solidity/test/hyperkzg/HyperKZGHelpers.t.pre.sol
+++ b/solidity/test/hyperkzg/HyperKZGHelpers.t.pre.sol
@@ -171,7 +171,13 @@ contract HyperKZGHelpersTest is Test {
     /// forge-config: default.fuzz.max-test-rejects = 100000
     function testFuzzRevertsVConsistency(uint256[3][] calldata v, uint256 r, uint256[] memory x, uint256 y) public {
         vm.assume(x.length > 0);
-        vm.assume(v.length == x.length);
+        uint256 xLength = x.length;
+        vm.assume(v.length == xLength);
+        bool xrAllZero = r == 0;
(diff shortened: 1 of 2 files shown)


solidity/test/hyperkzg/HyperKZGHelpers.t.pre.sol
```diff
@@ -171,7 +171,13 @@ contract HyperKZGHelpersTest is Test {
     /// forge-config: default.fuzz.max-test-rejects = 100000
     function testFuzzRevertsVConsistency(uint256[3][] calldata v, uint256 r, uint256[] memory x, uint256 y) public {
         vm.assume(x.length > 0);
-        vm.assume(v.length == x.length);
+        uint256 xLength = x.length;
+        vm.assume(v.length == xLength);
+        bool xrAllZero = r == 0;
+        for (uint256 i = 0; i < xLength; ++i) {
+            xrAllZero = xrAllZero && (x[i] == 0);
+        }
```
iajoiner, line 179: Just to be clear we filter it out if either r and all of x[i] are 0.

iajoiner (approved):

PR #612, opened 2025-03-07
feat: equality between varbianry columns

Dustin-Ray:
# Rationale for this change

Enables queries of the form:

```sql
"SELECT * FROM table WHERE a=b AND c=d AND e=f"
```
for the varbinary type.

Simple update to PoSQL to update the binary check operation to enable var binary support

# Are these changes tested?

- [x]  test equality queries

### DIFF
--- a/crates/proof-of-sql/src/base/database/group_by_util_test.rs
+++ b/crates/proof-of-sql/src/base/database/group_by_util_test.rs
@@ -8,6 +8,46 @@ use crate::{
 use bumpalo::Bump;
 
 #[test]
+#[should_panic(expected = "MAX can not be applied to varbinary")]
+fn we_cannot_apply_max_to_varbinary() {
+    let col: Column<'_, TestScalar> = Column::VarBinary((&[], &[]));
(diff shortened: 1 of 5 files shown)


crates/proof-of-sql/src/base/database/group_by_util_test.rs
```diff
@@ -7,6 +7,46 @@ use crate::{
...
 use bumpalo::Bump;
 
+#[test]
+#[should_panic(expected = "MAX can not be applied to varbinary")]
+fn we_cannot_apply_max_to_varbinary() {
+    let col: Column<'_, TestScalar> = Column::VarBinary((&[], &[]));
+    let indexes = &[];
+    let counts = &[];
+    let alloc = bumpalo::Bump::new();
+    let _ = max_aggregate_column_by_index_counts(&alloc, &col, counts, indexes);
```
iajoiner, line 17: Probably better to use assert!(matches!()) for errors

iajoiner (approved):

PR #613, opened 2025-03-07
feat: support for VarBinary literal

JayWhite2357:
# Rationale for this change

We want support for Binary literals.

1. Add a `VarBinary` variant to the `LiteralValue` enum.  
2. Use `from_byte_slice_via_hash` to convert byte slices into a `Scalar` where necessary.  
3. Implement direct unit tests for any new functionality.  
4. Add a new test in `equals_expr_test.rs`:  
   - Query: `SELECT a, b FROM table WHERE b = [4,5,6,7]`  
   - Data:  
     a (bigint) | b (varbinary)  
     123        | [1,2,3]  
     4567       | [4,5,6,7]  
   - Ensure the test follows the style of existing tests in the file.  
5. Add a `VarBinary` variant to the `Literal` enum in `intermediate_ast`.  
   - The underlying type should be `Vec<u8>`.  
6. Update the `lalrpop` file to support this new variant.  
   - Use the regex `0x([0-9a-fA-F]{2})*` to match hex literals.  
   - Example: `0xabcd1234` should be a valid literal (without quotes).  
7. If needed in `sqlparser`, represent the literal using `sqlparser::ast::Value::HexStringLiteral("abcd1234")`.
8. Add the same query above to `integration_tests.rs` as a test case.
   - Instead of starting with a parsed plan, it should parse the query "SELECT a, b FROM table WHERE b = 0x04050607".

# Are these changes tested?
Yes


iajoiner (approved):
@JayWhite2357 Really thanks!

PR #621, opened 2025-03-10
chore: Update GitHub Actions workflow to use Swatinem/rust-cache

zelosleone:
# Rationale for this change

Should be better to use swatinem/rust-cache action than manually setting up cache. It caches only the succesfull builds that are merged/from main for every PR. So only main builds are cached, and only crates are cached. Since we also have cargo.lock file (which is great for this) it's already pretty great. We should see it's affects after the succesful main builds are cached though.

Github workflow file is updated to use swatinem/rust-cache action.

# Are these changes tested?
Yes.


tlovell-sxt (commented):
Thanks! I haven't seen this action before but it looks good.

Do you think you could update the commit message/PR title to use `ci:` as the type and be more specific about what workflows are being updated?

```
ci: cache rust dependencies in lint-and-test workflows with Swatinem/rust-cache
```

zelosleone:
> Thanks! I haven't seen this action before but it looks good.
> 
> Do you think you could update the commit message/PR title to use `ci:` as the type and be more specific about what workflows are being updated?
> 
> ```
> ci: cache rust dependencies in lint-and-test workflows with Swatinem/rust-cache
> ```

No problem! I just did change the title of pr and commit, i also found about it and tested while trying to improve the runtime/compile time of tests.

zelosleone:
> @zelosleone this has a marginal improvement on the CI. I'm guessing that if you re-enable #610, we could get some significant improvements since the build time could be significantly smaller.

Hmm, although it improved it, I will see if we can improve the cache with specific commands for compiling during testing first. I feel like it could be faster?

PR #623, opened 2025-03-11
fix: add `exit` when YAML file not found

666devs:

# Rationale for this change

When the `exit` command is missing, the script keeps running even if the critical file isn't there. By adding `exit 0`, you make sure that the script stops when the file is missing—just as the message indicates

 Example:
 Add `NestedLoopJoinExec`.
 Closes #345.

 Since we added `HashJoinExec` in #323 it has been possible to do provable inner joins. However performance is not satisfactory in some cases. Hence we need to fix the problem by implement `NestedLoopJoinExec` and speed up the code
 for `HashJoinExec`.
-->

# Are these changes tested?

### DIFF
--- a/scripts/run_posql_examples.sh
+++ b/scripts/run_posql_examples.sh
@@ -13,6 +13,7 @@ if [ ! -f "$YAML_FILE" ]; then
   echo "Error: $YAML_FILE not found."
   echo "Please run this script from the repository root or update YAML_FILE accordingly."
   echo "Exiting with status 0, no commands executed."
+  exit 0
 
 fi
 


666devs:
> `exit 0` causes the script to immediately exit when something fails or goes wrong. This hides any error messages or warnings from the user and is generally a bad experience. There should never be a situation where the YAML file is not found. These scripts simply search for any yaml files and run the commands within.
> 
> If you are encountering a YAML not found error of any sort, that is a different issue that needs to be resolved, but this is not the solution to that at this time.
> 
> We will automatically close this PR in 3 days if no modifications are made. Thanks.

Got u, thanks!

PR #626, opened 2025-03-12
refactor(sol): improve sumcheck verification API

JayWhite2357:
# Rationale for this change

Part of larger effort of EVM verifier port.

* The sumcheck proof length can be considered to be part of the overall sumcheck proof message. So, it makes sense for this to be the only component hashed. `num_vars` is a public input, and is now required to be added to the transcript by callers of the function. Furthermore, is now considered to be a message from the prover that the verifier simply validates, in order to simplify serialization.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/sumcheck/proof_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/sumcheck/proof_test.rs
@@ -235,3 +235,47 @@ fn we_can_verify_many_random_test_cases() {
         );
     }
 }
+
+#[test]
+fn we_can_generate_and_verify_a_simple_sumcheck_proof() {
+    use crate::{base::proof::Keccak256Transcript, proof_primitive::hyperkzg::BNScalar};
+
+    let num_vars = 3;
+    let degree = 2;
+    let state = ProverState::new(
+        vec![(1.into(), vec![0, 1]), (-BNScalar::ONE, vec![2])],
+        vec![
+            (101..=108).map(Into::into).collect(),
+            (201..=208).map(Into::into).collect(),
+            (101..=108)
+                .zip(201..=208)
+                .map(|(a, b)| a * b)
+                .map(Into::into)
+                .collect(),
+        ],
+        num_vars,
+        degree,
+    );
+    let mut transcript = Keccak256Transcript::new();
+    transcript.extend_as_be([
+        0x0123_4567_89AB_CDEF_0123_4567_89AB_CDEF_u128,
+        0x0123_4567_89AB_CDEF_0123_4567_89AB_CDEF_u128,
+    ]);
+    let mut evaluation_point = vec![MontScalar::default(); num_vars];
+    let proof = SumcheckProof::<BNScalar>::create(&mut transcript, &mut evaluation_point, state);
+    let mut transcript = Keccak256Transcript::new();
+    transcript.extend_as_be([
+        0x0123_4567_89AB_CDEF_0123_4567_89AB_CDEF_u128,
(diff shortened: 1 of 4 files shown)


crates/proof-of-sql/src/proof_primitive/sumcheck/proof_test.rs
```diff
@@ -235,3 +235,48 @@ fn we_can_verify_many_random_test_cases() {
...
+    ]);
+
+    dbg!(&proof.coefficients);
+
+    let subclaim = proof
+        .verify_without_evaluation(&mut transcript, num_vars, &BNScalar::ZERO)
+        .unwrap();
+    assert_eq!(subclaim.evaluation_point, evaluation_point,);
+
+    dbg!(subclaim.expected_evaluation);
```
iajoiner, line 280: Please remove the dbg lines

solidity/src/proof/Sumcheck.pre.sol
```diff
@@ -33,13 +41,41 @@ library Sumcheck {
             function append_calldata(transcript_ptr, offset, size) {
                 revert(0, 0)
             }
-            function verify_sumcheck_proof(transcript_ptr, proof_ptr, num_vars, degree) ->
+
+            // actual_sum = coefficient_i + sum_{i=0}^{degree} coefficient_i,
```
MercysJest, line 45: Minor Typo: Shouldn't this be `actual_sum = coefficient_0 + sum_{i=0}^{degree} coefficient_i` ?

solidity/src/proof/Sumcheck.pre.sol
```diff
@@ -33,13 +41,41 @@ library Sumcheck {
...
+            //                  where coefficient_i is the coefficient of x^i
+            // round_evaluation = sum_{i=0}^{degree} coefficient_i * challenge^i
+            // NOTE: the coefficients are in "reverse" order, with the leading coefficient first
+            //       as a result, round_evaluation is computed with Horner's method
+            function process_round(proof_ptr, degree, challenge) -> proof_ptr_out, round_evaluation, actual_sum {
+                let coefficient := calldataload(proof_ptr)
+                proof_ptr := add(proof_ptr, WORD_SIZE)
+                round_evaluation := coefficient
```
MercysJest, line 53: Minor invariant assumption: When the degree is zero (`sumcheck_length=num_vars`), `round_evaluation` is never modded. This could be a minor completeness issue if the prover does not reduce coefficients mod p before calling. 
- `if sub(expected_evaluation, actual_sum) { err(ERR_ROUND_EVALUATION_MISMATCH) }` could fail. 
- the output `expected_evaluation` would not be modded, so assumes the verifier later on does a mod. 

To be robust, could just add a `mod` here. This case only comes up if degree = 0, so could just throw an error if degree is trivial too. Not sure if y'all have a use case for trivial degrees.

MercysJest:
Overall, sumcheck verifier looks good to me. Only have the minor typo and invariant comment.

solidity/src/proof/Sumcheck.pre.sol
MercysJest: fixed in f12f4df2f194b431fc96e93c3302976cc75fe848

solidity/src/proof/Sumcheck.pre.sol
MercysJest: fixed in f12f4df2f194b431fc96e93c3302976cc75fe848

MercysJest (dismissed):

iajoiner (approved):

PR #665, opened 2025-03-25
feat: add byte distribution

stuarttimwhite:

# Rationale for this change

Range check will benefit from a `ByteDistribution`, which will reduce the number of columns it needs to commit.

New `ByteDistribution`.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/base/byte/mod.rs
+++ b/crates/proof-of-sql/src/base/byte/mod.rs
@@ -1,0 +1,1 @@
+mod byte_distribution;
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/base/byte/mod.rs
```diff
@@ -0,0 +1 @@
+mod byte_distribution;
```
iajoiner, line 1: Hmm maybe we don't need the `byte` directory?

stuarttimwhite: There was going to be another file `byte_matrix_utils` next. I could put the bit and byte code together if that sounds better.

iajoiner: I see so let's keep it then.

iajoiner (approved):

iajoiner:
@stuarttimwhite Really thanks for this! This is pretty cool!

PR #670, opened 2025-03-26
chore(sol): update transcript

JayWhite2357:
# Rationale for this change

Some updates to the solidity transcript implementation are needed for upcoming PRs.

See individual commits:

* [refactor(sol): make draw_challenges return length-prefixed array](https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/670/commits/2cf60597f1fb6c819676813175bb17a03c75b080)

    Previously, `draw_challenges` would allocate memory of length `count`,
    fill that memory with `count` challenges,
    and return a pointer to the first challenge.
    This commit changes this so that it allocates an extra word,
    starts the memory with a value of `count`,
    and returns a pointer to that first value.
    
    This aligns with the solidity memory layout for arrays.
    While it is slightly less efficient, it is helpful for keeping the stack small.
    Although, future optimizations may wish to revert this.

* [feat(sol): add append_array](https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/670/commits/eddd9c0e66a04390442c94b875fa80877bc721b8)

    Adds an in-memory array to the transcript. This is similar to `append_calldata`, but is memory instead.

# Are these changes tested?
Yes

### DIFF
--- a/solidity/src/base/Transcript.sol
+++ b/solidity/src/base/Transcript.sol
@@ -90,4 +90,27 @@ library Transcript {
         }
         __resultTranscript = __transcript;
     }
+
+    /// @notice Append an array to the transcript, and update the state of the transcript.
+    /// @dev This is achieved by hashing the current transcript state with the array data.
+    /// The operation is done in place by temporarily swapping the array length with the transcript state.
+    /// @param __transcript The current state of the transcript
+    /// @param __data The array to append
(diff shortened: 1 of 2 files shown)


iajoiner (approved):

MercysJest (approved):
Looks good to me!

PR #678, opened 2025-03-27
chore: make hyperkzg proof EVM compatible

JayWhite2357:
# Rationale for this change

We need the hyperkzg proof to be easily deserialized in the EVM.

* Added a `load_small_setup_for_testing` method to load the Perpetual Powers of Tau setup.
* Changes the serailization format of the proof in both the Rust and the EVM code.
* Added tests that generate a proof in Rust and verify it in the EVM.

# Are these changes tested?
Yes. Note, code coverage doesn't pick up on the EVM tests since they are ignored by default.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -248,3 +248,20 @@ jobs:
         with:
           target: solidity
           slither-config: solidity/slither.config.json
+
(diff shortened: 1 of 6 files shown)


MercysJest:
Solidity changes look reasonable to me.

iajoiner (approved):

PR #685, opened 2025-03-29
feat: add `try_scale_cast_types`

stuarttimwhite:

# Rationale for this change

In order to prepare for a `ScaleCastExpr`, there is a need for a function that determines if one type can be scale cast to another type.

A new function `try_scale_cast_types`

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/base/database/column_type_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_type_operation.rs
@@ -207,6 +207,38 @@ pub fn try_cast_types(from: ColumnType, to: ColumnType) -> ColumnOperationResult
     }
 }
 
+/// Verifies that `from` can be cast to `to`.
+/// Casting can only be supported if the resulting data type is a superset of the input data type.
+/// For example Deciaml(6,1) can be cast to Decimal(7,1), but not vice versa.
+#[cfg_attr(not(test), expect(dead_code))]
+#[expect(clippy::missing_panics_doc)]
+pub fn try_scale_cast_types(from: ColumnType, to: ColumnType) -> ColumnOperationResult<()> {
+    match (from, to) {
+        (
+            ColumnType::TinyInt
+            | ColumnType::Uint8
+            | ColumnType::SmallInt
+            | ColumnType::Int
+            | ColumnType::Int128
+            | ColumnType::BigInt
+            | ColumnType::Decimal75(_, _),
+            ColumnType::Decimal75(precision, scale),
+        ) => {
+            let from_precision = i16::from(from.precision_value().unwrap());
+            let from_scale = i16::from(from.scale().unwrap());
+            let to_precision = i16::from(precision.value());
+            let to_scale = i16::from(scale);
+            to_scale >= from_scale && (to_precision - to_scale) >= (from_precision - from_scale)
+        }
+        _ => false,
+    }
+    .then_some(())
+    .ok_or(ColumnOperationError::ScaleCastingError {
+        left_type: from,
+        right_type: to,
+    })
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql/src/base/database/column_type_operation.rs
```diff
@@ -207,6 +207,37 @@ pub fn try_cast_types(from: ColumnType, to: ColumnType) -> ColumnOperationResult
...
+            | ColumnType::BigInt
+            | ColumnType::Decimal75(_, _),
+            ColumnType::Decimal75(precision, scale),
+        ) => {
+            let from_precision = i16::from(from.precision_value().unwrap());
+            let from_scale = i16::from(from.scale().unwrap());
+            let to_precision = i16::from(precision.value());
+            let to_scale = i16::from(scale);
+            to_scale >= from_scale
+                && (to_precision - to_scale.max(0)) >= (from_precision - from_scale.max(0))
```
iajoiner, line 230: Hmm why max(0)? Decimal(1, 0) -> Decimal(1, -1) is lossy.

stuarttimwhite: The way this is written, Decimal(1, 0) -> Decimal(1, -1) would be an invalid cast because the to type has a smaller scale. The `(to_precision - to_scale.max(0)) >= (from_precision - from_scale.max(0))` is supposed to check if the to type is big enough, and a negative scale doesn't take away from (or add to) the number of digits to the left of the decimal place like a positive value would.

crates/proof-of-sql/src/base/database/column_type_operation.rs
```diff
@@ -857,4 +887,98 @@ mod test {
...
+            try_scale_cast_types(from, ColumnType::Decimal75(two_prec, 1)).unwrap_err();
+            try_scale_cast_types(from, ColumnType::Decimal75(from_precision, 0)).unwrap();
+            try_scale_cast_types(from, ColumnType::Decimal75(from_precision, -1)).unwrap_err();
+            try_scale_cast_types(from, ColumnType::Decimal75(from_precision, 1)).unwrap_err();
+            try_scale_cast_types(from, ColumnType::Decimal75(forty_prec, 0)).unwrap();
+            try_scale_cast_types(from, ColumnType::Decimal75(forty_prec, -1)).unwrap_err();
+            try_scale_cast_types(from, ColumnType::Decimal75(forty_prec, 1)).unwrap();
+        }
+
+        let from_prec = Precision::new(20).unwrap();
```
iajoiner, line 915: We may as well name it `twenty_prec`

crates/proof-of-sql/src/base/database/column_type_operation.rs
```diff
@@ -857,4 +887,98 @@ mod test {
...
+
+        // from_with_negative_scale
+        let neg_scale = ColumnType::Decimal75(from_prec, -3);
```
iajoiner, line 918: We may as well refer to it as `decimal_twenty_minusthree` or something. Same for the rest.

crates/proof-of-sql/src/base/database/column_type_operation.rs
```diff
@@ -207,6 +207,36 @@ pub fn try_cast_types(from: ColumnType, to: ColumnType) -> ColumnOperationResult
     }
 }
 
+/// Verifies that `from` can be cast to `to`. This only supports converting types to a larger type (Decimal(10, 5) to Decimal (11, 5) for example).
```
iajoiner, line 210: Please explain what "larger type" means so that any external user or future contributor understands it

crates/proof-of-sql/src/base/database/column_type_operation.rs
stuarttimwhite: I'm trying to keep the variable names short so that `cargo f` doesn't break them into multiple lines. It's just a lot more readable when it's one line.

PR #687, opened 2025-03-31
feat: try get scaling factor with precision and scale

stuarttimwhite:

# Rationale for this change

The primary calculation that is needed to cast an integer or decimal type to another decimal type is the factor by which the original value needs to be multiplied.

New function `try_get_scaling_factor_with_precision_and_scale`, which returns a scalar, because the value can be as big 10^247.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
@@ -1,7 +1,11 @@
 use crate::base::{
-    database::{Column, ColumnType, ColumnarValue, LiteralValue},
+    database::{
+        try_scale_cast_types, Column, ColumnOperationResult, ColumnType, ColumnarValue,
+        LiteralValue,
+    },
     scalar::{Scalar, ScalarExt},
 };
+use bnum::types::U256;
 use bumpalo::Bump;
 use core::{cmp::Ordering, ops::Neg};
 use itertools::izip;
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
```diff
@@ -196,8 +200,8 @@ pub(crate) fn scale_and_add_subtract_eval<S: Scalar>(
 /// Panics when there is a casting issue. These errors should only happen if `is_right_bigger_int_type` is the wrong value.
 fn divide_integer_columns<
     'a,
-    L: NumCast + Copy + PrimInt + Neg<Output = L>,
-    R: NumCast + Copy + PrimInt + Neg<Output = R>,
+    L: num_traits::NumCast + Copy + PrimInt + Neg<Output = L>,
```
iajoiner, line 203: Is there any reason to import `num_traits::NumCast` whenever it is used instead of at the top?

crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
```diff
@@ -554,19 +558,42 @@ pub fn cast_column<'a, S: Scalar>(
...
+#[expect(clippy::missing_panics_doc)]
+#[cfg_attr(not(test), expect(dead_code))]
+fn try_get_scaling_factor_with_precision_and_scale<S: Scalar>(
+    from_type: ColumnType,
+    to_type: ColumnType,
+) -> ColumnOperationResult<(S, u8, i8)> {
+    try_scale_cast_types(from_type, to_type)?;
+    let to_precision = to_type.precision_value().unwrap();
+    let to_scale = to_type.scale().unwrap();
+    let power = to_precision - from_type.precision_value().unwrap();
```
iajoiner, line 573: If we convert Decimal(2, 1) to Decimal(5, 3) we should scale by 10^(3 - 1) = 100, not 10^(5 - 2) = 1,000

crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
stuarttimwhite: Yeah, there are two separate `from` functions.

crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
stuarttimwhite: Ah right.

crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
```diff
@@ -810,4 +837,135 @@ mod tests {
             ColumnType::VarBinary,
         );
     }
+
+    #[allow(clippy::too_many_lines)]
```
iajoiner, line 841: Maybe you want to break up the test?

crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
```diff
@@ -554,19 +558,42 @@ pub fn cast_column<'a, S: Scalar>(
     }
 }
 
+/// Tries to get the scale factor between the from and to types.
+/// The precision and scale are returned along with the scale so that the unwrapping
+/// can occur in the function that confirms that the types are castable
+#[expect(clippy::missing_panics_doc)]
```
iajoiner, line 564: I'd add a panic doc that says that panics are not possible if we are on numerical types

crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
```diff
@@ -238,8 +242,8 @@ fn divide_integer_columns<
 /// Panics when there is a casting issue. These errors should only happen if `is_right_bigger_int_type` is the wrong value.
 fn modulo_integer_columns<
     'a,
-    L: NumCast + Copy + PrimInt + Neg<Output = L>,
-    R: NumCast + Copy + PrimInt + Neg<Output = R>,
+    L: num_traits::NumCast + Copy + PrimInt + Neg<Output = L>,
+    R: num_traits::NumCast + Copy + PrimInt + Neg<Output = R>,
```
iajoiner, line 246: If we just replace `num_traits::NumCast` with `NumCast` and add `use num_traits::NumCast;` at the top what will happen?

crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
stuarttimwhite: If I remember correctly, in the past Jay suggested to just put the attribute there if the panics aren't possible.

crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
stuarttimwhite: That's what it was before. I was getting conflicting `from` implementations.

iajoiner: OK

iajoiner (approved):
@stuarttimwhite Really thanks!

iajoiner (approved):

iajoiner (approved):

Issue #545, opened 2025-02-13
chore: transition from `allow` to `expect` for turning off clippy lints


JayWhite2357:
# Background and Motivation

Currently, we have been using `allow` to allow clippy lints on a case-by-case basis. Using `expect` is the better solution to this.

See https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html#expectlint.

See the conversation here for an example of why this is useful: https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/539#discussion_r1951980743.

# Changes Required

- [x] Add `allow_attributes = "warn"` to the `Cargo.toml` file and make any necessary changes.

RaghavArora14:
/attempt #545
## Proposed Steps:

1. **Configure Cargo.toml:**  I will modify the `Cargo.toml` file, specifically the `[workspace.lints.clippy]` section, to include `allow_attributes = ["warn"]`. This will enable warnings for any remaining `#[allow(...)]` attributes, aiding in the identification of areas needing updates.

2. **Systematic Codebase Update: ** This phase constitutes the bulk of the work. I will employ a combination of techniques:
    * **IDE Search and Replace:** I will leverage my IDE's powerful search and replace capabilities, utilizing regular expressions to efficiently locate all occurrences of `#[allow(clippy::...)]`. The regular expression `#\[allow\(clippy::[^)]*\)` will be used to identify these instances.
    * **Manual Review and Replacement:**  Each identified instance will be manually reviewed to ensure the correct replacement with `#[expect(clippy::...)]`. This meticulous approach minimizes the risk of introducing errors or unintended behavior.  The focus will be on preserving the original lint name within the parentheses.
    * **Incremental Commits:**  To maintain a clean and manageable commit history, changes will be committed incrementally, allowing for easier debugging and rollback if necessary.

3. **Comprehensive Testing :**  After completing the code modifications, a thorough testing phase will be conducted. This involves executing the existing test suite to confirm that all functionality remains unaffected by the changes.  Any identified issues will be addressed before proceeding.

4. **Pull Request Submission :** Once the testing phase is successfully completed, the updated code will be pushed to a dedicated branch in my forked repository. A pull request will then be submitted, referencing issue #545 and including `/claim #545` in the description to claim the bounty.


## Time Estimate:

**Total Estimated Time: 24-36 hours**

[...]

PR #699, opened 2025-04-02
fix: transition from allow to expect for turning off clippy lints

divanshu-go:
/claim #545

* Changed Clippy annotations from `#[allow]` to `#[expect]` in multiple files to improve code clarity and maintainability. 

* Added `allow_attributes = "warn"` to the Clippy configuration in `Cargo.toml` to handle attribute warnings.

### DIFF
--- a/crates/proof-of-sql-parser/src/lib.rs
+++ b/crates/proof-of-sql-parser/src/lib.rs
@@ -35,7 +35,7 @@ pub use resource_id::ResourceId;
 pub mod sqlparser;
 
 // lalrpop-generated code is not clippy-compliant
-lalrpop_mod!(#[allow(clippy::all, missing_docs, clippy::missing_docs_in_private_items, clippy::pedantic, clippy::missing_panics_doc)] pub sql);
+lalrpop_mod!(#[allow(clippy::all, missing_docs, clippy::missing_docs_in_private_items, clippy::pedantic, clippy::missing_panics_doc, clippy::allow_attributes)] pub sql);
 
 /// Implement [`Deserialize`](serde::Deserialize) through [`FromStr`](core::str::FromStr) to avoid invalid identifiers.
 #[macro_export]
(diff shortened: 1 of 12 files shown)


iajoiner:
@divanshu-go Really thanks for your contribution! Please run clippy on your own via the following
```bash
cargo clippy --all-targets --all-features
```
and make sure replacing `allow` with `expect` actually works. In particular the `allow` in `lalrpop` is mandatory.

Please let me know if you have any questions! Please feel free to post in our Discord channel or contact me on social media.
Discord: chloemargaret
Instagram: chloezhooman

crates/proof-of-sql-parser/src/lib.rs
```diff
@@ -35,7 +35,7 @@ pub use resource_id::ResourceId;
 pub mod sqlparser;
 
 // lalrpop-generated code is not clippy-compliant
-lalrpop_mod!(#[allow(clippy::all, missing_docs, clippy::missing_docs_in_private_items, clippy::pedantic, clippy::missing_panics_doc)] pub sql);
+lalrpop_mod!(#[expect(clippy::all, missing_docs, clippy::missing_docs_in_private_items, clippy::pedantic, clippy::missing_panics_doc)] pub sql);
```
iajoiner, line 38: Please don't change this for this will not work

divanshu-go:
@iajoiner what is the expected result of this whole task
cause cargo's clippy is always going to complain about this or that or whatever.

iajoiner:
@divanshu-go No more complaints while the amount of allows are kept at a minimum, that is, anything that can be expect should be.

iajoiner:
@divanshu-go Please run ` cargo cl -- -D warnings`. Please run CI on your side and make sure it works before getting back to us.

iajoiner:
For the expect lints that fail please take a look and see whether you can remove the lints without things breaking. If so please just remove them. Otherwise you might need to conditionally add `expect` when we are in tests or not in tests or maybe even retain `allow` since sometimes the problem is that a lint is triggered in some but not all cases.

divanshu-go:
@iajoiner Instead of doing trival work as you mentioned in your comment https://github.com/spaceandtimelabs/sxt-proof-of-sql/pull/699#issuecomment-2784624876

I think using `cargo cl -- -D warnings -A unfulfilled_lint_expectations`
is a more viable option.

also see the below screenshot

image

iajoiner (approved):

PR #701, opened 2025-04-02
feat: scaling cast expr

stuarttimwhite:

# Rationale for this change

This PR finishes the creation of the `DecimalScalingCastExpr`. This will allow proof expressions to cast ints and decimals to decimals.

- Rename utility functions that are meant to cast to decimals to include the word decimal
- New `DecimalScalingCastExpr`
- New variant in `DynProofExpr`

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/decimal_scaling_cast_expr.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/decimal_scaling_cast_expr.rs
@@ -26,7 +26,6 @@ pub struct DecimalScalingCastExpr {
 
 impl DecimalScalingCastExpr {
     /// Creates a new `CastExpr`
-    #[expect(dead_code)]
     pub fn try_new(
         from_expr: Box<DynProofExpr>,
         to_type: ColumnType,
--- a/crates/proof-of-sql/src/sql/proof_exprs/decimal_scaling_cast_expr.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/decimal_scaling_cast_expr.rs
@@ -1,0 +1,95 @@
+use super::{
+    numerical_util::{
+        cast_column_to_decimal_with_scaling, try_get_scaling_factor_with_precision_and_scale,
+    },
+    DynProofExpr, ProofExpr,
+};
+use crate::{
+    base::{
+        database::{Column, ColumnOperationResult, ColumnRef, ColumnType, LiteralValue, Table},
+        map::{IndexMap, IndexSet},
+        proof::{PlaceholderResult, ProofError},
+        scalar::Scalar,
+    },
+    sql::proof::{FinalRoundBuilder, VerificationBuilder},
+};
+use alloc::boxed::Box;
+use bumpalo::Bump;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+pub struct DecimalScalingCastExpr {
+    from_expr: Box<DynProofExpr>,
+    to_type: ColumnType,
+    scaling_factor: [u64; 4],
+}
+
+impl DecimalScalingCastExpr {
+    /// Creates a new `CastExpr`
+    #[expect(dead_code)]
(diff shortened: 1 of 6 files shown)


crates/proof-of-sql/src/base/database/column_type_operation.rs
```diff
@@ -983,84 +984,118 @@ mod test {
...
-        try_scale_cast_types(zero_scale, ColumnType::Decimal75(twenty_one_prec, 0)).unwrap();
-        try_scale_cast_types(zero_scale, ColumnType::Decimal75(twenty_one_prec, 1)).unwrap();
-        try_scale_cast_types(zero_scale, ColumnType::Decimal75(twenty_one_prec, 2)).unwrap_err();
+        try_decimal_scale_cast_types(zero_scale, ColumnType::Decimal75(twenty_prec, -1))
+            .unwrap_err();
+        try_decimal_scale_cast_types(zero_scale, ColumnType::Decimal75(twenty_prec, 0)).unwrap();
+        try_decimal_scale_cast_types(zero_scale, ColumnType::Decimal75(twenty_prec, 1))
+            .unwrap_err();
+
+        try_decimal_scale_cast_types(zero_scale, ColumnType::Decimal75(nineteen_prec, -1))
```
iajoiner, line 1046: Actually you should be able to consolidate these tests using loops and a function to determine whether it should error out.

crates/proof-of-sql/src/sql/proof_exprs/decimal_scaling_cast_expr.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub fn new(from_expr: Box<DynProofExpr>, to_type: ColumnType) -> Self {
+        Self { from_expr, to_type }
+    }
+}
+
+impl ProofExpr for DecimalScalingCastExpr {
+    fn data_type(&self) -> ColumnType {
+        try_decimal_scale_cast_types(self.from_expr.data_type(), self.to_type)
+            .expect("Failed to cast column type");
+        self.to_type
```
iajoiner, line 39: Or maybe we should simply use `self.to_type`? The type check should happen elsewhere.

crates/proof-of-sql/src/sql/proof_exprs/decimal_scaling_cast_expr.rs
```diff
@@ -0,0 +1,96 @@
...
+        params: &[LiteralValue],
+    ) -> Result<S, ProofError> {
+        let unscaled_eval = self
+            .from_expr
+            .verifier_evaluate(builder, accessor, chi_eval, params)?;
+        try_get_scaling_factor_with_precision_and_scale::<S>(
+            self.from_expr.data_type(),
+            self.to_type,
+        )
+        .map(|(scaling_factor, _, _)| scaling_factor * unscaled_eval)
```
iajoiner, line 87: Since verifiers can receive malicious proof plans I guess we have to have type checks here

iajoiner (commented):
@stuarttimwhite Looks mostly good. Please address the comments.

crates/proof-of-sql/src/sql/proof_exprs/decimal_scaling_cast_expr.rs
stuarttimwhite: I don't think so. Scaling factor and unscaled_eval are both trusted at this point

crates/proof-of-sql/src/base/database/column_type_operation.rs
stuarttimwhite: I would rather explicitly state the expectation foreach case. If I made a function to dynamically determine the expected outcome, I suspect I'd just be recreating `try_decimal_scale_cast_types` in a test function, and I don't think testing a function with itself is an effective way to test.

stuarttimwhite: Also, these tests are not new, so I think if we wanted to make that change, it should be in a different PR.

iajoiner (approved):

PR #705, opened 2025-04-02
feat(sol): complete EVM verifier

JayWhite2357:
# Rationale for this change

This is the final PR for the EVM verifier. Follow up PRs will need to add further features to the verifier to support all proof types.

### DIFF
--- a/crates/proof-of-sql/src/sql/evm_proof_plan/evm_tests.rs
+++ b/crates/proof-of-sql/src/sql/evm_proof_plan/evm_tests.rs
@@ -1,0 +1,219 @@
+use crate::{
+    base::database::{
+        owned_table_utility::{bigint, owned_table},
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/sql/evm_proof_plan/evm_tests.rs
```diff
@@ -0,0 +1,219 @@
...
+        parse::QueryExpr,
+        proof::{ProofPlan, VerifiableQueryResult},
+    },
+};
+use ark_ec::{AffineRepr, CurveGroup};
+use ark_ff::PrimeField;
+use itertools::Itertools;
+
+fn evm_verifier_with_extra_args(
+    query: &QueryExpr,
```
iajoiner, line 21: Maybe we can switch to `DynProofPlan`?

PR #712, opened 2025-04-04
feat(sol): add `skip_plan_names`

JayWhite2357:
# Rationale for this change

The verifier needs to be able to skip over the table/column name prefix of the proof plan since it is unused in the proof proper (although part of public input).

### DIFF
--- a/solidity/src/verifier/PlanUtil.sol
+++ b/solidity/src/verifier/PlanUtil.sol
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: UNLICENSED
+// This is licensed under the Cryptographic Open Software License 1.0
+pragma solidity ^0.8.28;
+
+import "../base/Constants.sol";
+
+/// @title Plan Utility Library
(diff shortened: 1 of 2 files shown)


solidity/src/verifier/PlanUtil.sol
```diff
@@ -0,0 +1,65 @@
...
+/// @notice A library for handling utility functions related to plans.
+library PlanUtil {
+    /// @notice The Proof Plan is prefixed with metadata about the plan, primarily the names of the tables and columns.
+    /// @notice This method skips over the names in a plan and returns the updated pointer.
+    /// @dev The format of the plan is as follows:
+    /// @dev * number of tables (uint64)
+    /// @dev * table names
+    /// @dev     * length of table name (uint64)
+    /// @dev     * table name (variable length)
+    /// @dev * number of columns (uint64)
```
iajoiner, line 17: Maybe we should make it clear that we mean input columns

MercysJest:
Looks good to me!

PR #713, opened 2025-04-04
feat(sol): add `proof_plan_evaluate`

JayWhite2357:
# Rationale for this change

We need a method to delegate work to the different Proof Plan variants.

Add the `proof_plan_evaluate` method which simply reads the variant and delegates to the appropriate plan evaluator.

# Are these changes tested?
Yes


iajoiner (approved):

MercysJest (commented):
Looks good to me!

PR #718, opened 2025-04-06
perf(sol): reduce `log2_up` gas

JayWhite2357:
# Rationale for this change

Gas is expensive. We should use less of it.

`log2_up` is optimized to be $O(1)$ cost rather than $O(\log(value))$ cost. This has concrete gas improvements when value is at least 16. The typical situation will involve tables larger than 16 rows.

# Are these changes tested?
Yes, extra tests were added as well.


iajoiner (approved):
Note that `log2_up` is essentially `max(length of binary representation, 1)`. This works but maybe we have something even better.

PR #733, opened 2025-04-09
refactor!: ensure reasonable types for numeric ops

iajoiner:

# Rationale for this change
Currently `ProofExpr` of numeric types require scale casting as well as explicit passing of scales since arithmetic ops on `ProofExpr` always return `Scalar` type columns. In order to port the code to Solidity we need to simplify the logic.  

1. Make sure all arithmetic `ProofExpr` have decimal types and return decimal columns to ensure consistency && prevent overflows from being a thing
2. Cap precision to 75 and wrap around as opposed to banning ops that cause precision to be too large
# Are these changes tested?

Yes.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/proof_expr.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/proof_expr.rs
@@ -2,6 +2,7 @@ use crate::{
     base::{
         database::{Column, ColumnRef, ColumnType, LiteralValue, Table},
         map::{IndexMap, IndexSet},
+        math::decimal::Precision,
         proof::{PlaceholderResult, ProofError},
         scalar::Scalar,
     },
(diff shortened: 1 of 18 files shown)


crates/proof-of-sql/src/sql/proof_exprs/proof_expr.rs
```diff
@@ -52,3 +53,27 @@ pub trait ProofExpr: Debug + Send + Sync {
...
+    }
+
+    /// Get the scale of the expression
+    ///
+    /// # Panics
+    /// This panics if the scale is invalid
+    fn scale(&self) -> i8 {
+        self.data_type().scale().expect("Scale should be valid")
+    }
+}
```
stuarttimwhite, line 79: I think these two functions will always be used together, so this probably should just be one function. I also think this doesn't need to be in an extension trait. It could just be a function on ColumnType. I'm not super opinionated on this, so I'm fine with whatever you settle on.

iajoiner: @stuarttimwhite Since these two methods don't actually have much in common I'm not sure whether replacing two methods with one returning a tuple is really a better idea. `ColumnType` has `precision_value` and `scale` already and I propose to use them here which precisely established why we need them.

PR #748, opened 2025-04-14
feat(sol): EVM verifier

JayWhite2357:
# Rationale for this change

We want an EVM verifier.

* This adds a minor helper method (primarily to achive 100% unit test coverage).
* Adds the complete, main, verifier body.
* Adds Rust+EVM tests.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/evm_proof_plan/evm_tests.rs
+++ b/crates/proof-of-sql/src/sql/evm_proof_plan/evm_tests.rs
@@ -1,0 +1,200 @@
+use crate::{
+    base::database::{
+        owned_table_utility::{bigint, owned_table},
+        CommitmentAccessor, OwnedTableTestAccessor,
+    },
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql/src/sql/evm_proof_plan/evm_tests.rs
```diff
@@ -0,0 +1,219 @@
...
+        parse::QueryExpr,
+        proof::{ProofPlan, VerifiableQueryResult},
+    },
+};
+use ark_ec::{AffineRepr, CurveGroup};
+use ark_ff::PrimeField;
+use itertools::Itertools;
+
+fn evm_verifier_with_extra_args(
+    query: &QueryExpr,
```
iajoiner, line 21: Hmm it's better to use `DynProofPlan` directly since we are about to get rid of `QueryExpr` anyway

MercysJest (commented):
Looks good to me!

PR #756, opened 2025-04-16
fix: use workspace for all dependencies in proof-of-sql-planner

tlovell-sxt:
# Rationale for this change
Generally it's better to use workspace dependencies where possible to avoid accidentally introducing duplicate dependencies as new crates are developed. However, this is especially pertinent as we add automation to release proof-of-sql-planner to crates.io. This automation failed recently because the proof-of-sql dependency was defined with a path instead of from the workspace.

- proof-of-sql defined as default-features=false in workspace Cargo.toml
- all dependencies of proof-of-sql-planner are now defined from the workspace

# Are these changes tested?
These changes do not affect existing functionality, which should be verified by existing tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,8 +35,10 @@ clap = { version = "4.5.4" }
 criterion = { version = "0.5.1" }
 chrono = { version = "=0.4.39", default-features = false }
 curve25519-dalek = { version = "4", features = ["rand_core"] }
+datafusion = { version = "38.0.0", default-features = false }
 derive_more = { version = "0.99" }
 enum_dispatch = { version = "0.3.13" }
+getrandom = { version = "0.2.15", default-features = false }
 ff = { version = "0.13.0"}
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -53,7 +55,7 @@ num-bigint = { version = "0.4.4", default-features = false }
 opentelemetry = { version = "0.23.0" }
 opentelemetry-jaeger = { version = "0.20.0" }
 postcard = { version = "1.0" }
-proof-of-sql = { path = "crates/proof-of-sql" } # We automatically update this line during release. So do not modify it!
+proof-of-sql = { path = "crates/proof-of-sql", default-features = false } # We automatically update this line during release. So do not modify it!
```
tlovell-sxt, line 58: yes, I modified this line, but default-features is invalid in package dependencies w/ workspace=true, it has to be defined in the workspace. I verified that the sed that modifies this line should still be OK

iajoiner (approved):

PR #757, opened 2025-04-16
refactor: use schema accessor rather than index map

stuarttimwhite:

# Rationale for this change

The planner is currently using `DFSchema` in more places than is necessary, which makes use by external applications more difficult.

Replacing `IndexMap<TableReference, DFSchema>` with `impl SchemaAccessor`. This necessitates other changes as well.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql-planner/src/plan.rs
+++ b/crates/proof-of-sql-planner/src/plan.rs
@@ -1,6 +1,7 @@
 use super::{
-    aggregate_function_to_proof_expr, column_to_column_ref, df_schema_to_column_fields,
-    expr_to_proof_expr, table_reference_to_table_ref, AggregateFunc, PlannerError, PlannerResult,
+    aggregate_function_to_proof_expr, column_to_column_ref, expr_to_proof_expr,
+    schema_to_column_fields, table_reference_to_table_ref, AggregateFunc, PlannerError,
+    PlannerResult,
 };
 use alloc::vec::Vec;
 use datafusion::{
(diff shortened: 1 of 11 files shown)


crates/proof-of-sql-planner/src/plan.rs
```diff
@@ -117,20 +106,37 @@ fn table_scan_to_filter(
     ))
 }
 
+fn try_get_schema_as_vec_from_df_schema(
```
iajoiner, line 109: This probably belongs to `util.rs`

crates/proof-of-sql-planner/src/proof_plan_with_postprocessing.rs
```diff
@@ -69,25 +64,28 @@ pub fn logical_plan_to_proof_plan_with_postprocessing(
...
     use proof_of_sql::{
-        base::database::{ColumnField, ColumnRef, ColumnType, TableRef},
+        base::database::{ColumnField, ColumnRef, ColumnType, TableRef, TestSchemaAccessor},
         sql::{
             proof_exprs::{AliasedDynProofExpr, ColumnExpr, DynProofExpr, TableExpr},
             proof_plans::DynProofPlan,
         },
     };
+    use sqlparser::ast::Ident;
+    use std::hash::BuildHasherDefault;
```
iajoiner, line 88: Is this used?

crates/proof-of-sql-planner/src/util.rs
```diff
@@ -574,14 +545,4 @@ mod tests {
             ]
         );
     }
-
-    #[test]
```
iajoiner, line 578: Hmm it seems that this test is still relevant

crates/proof-of-sql-planner/src/plan.rs
stuarttimwhite: It's only used in `plan`, so I think it makes sense to keep as a private function.

crates/proof-of-sql-planner/src/util.rs
stuarttimwhite: `schema_to_column_fields` can't error anymore, and it doesn't take a `DFSchema` anymore, so I don't know how it would even be possible to keep this test.

iajoiner (approved):

PR #788, opened 2025-04-23
build: add the `blitzar` feature flag to the `nova-snark` crate

jacobtrombetta:
# Rationale for this change
GPU support was added to the Microsoft Nova project in a [recent PR](https://github.com/microsoft/Nova/pull/374). By building the crate with the `blitzar` feature flag, we can unlock performance gains during the `EvaluationEngine::prove` step.

Benchmarks on a Multi-A100 system using the `hyper-kzg` commitment scheme and a table size of `1,000,000` show a `1.75x` performance improvement related to `EvaluationEngine::prove`.

Without `blitzar`, `EvaluationEngine::prove` took `859.07 ms`

image

With `blitzar`, `EvaluationEngine::prove` took `489.53 ms`

image

- `blitzar` is added as a feature flag to the `nova-snark` crate.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/Cargo.toml
+++ b/crates/proof-of-sql/Cargo.toml
@@ -44,7 +44,7 @@ indexmap = { workspace = true, features = ["serde"] }
 indicatif = { workspace = true, optional = true }
 itertools = { workspace = true }
 merlin = { workspace = true, optional = true }
-nova-snark = { workspace = true, optional = true, features = [] }
+nova-snark = { workspace = true, optional = true, features = ["blitzar"] }
 num-traits = { workspace = true }
 num-bigint = { workspace = true, default-features = false }
 postcard = { workspace = true, features = ["alloc"] }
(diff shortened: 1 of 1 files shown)


crates/proof-of-sql/Cargo.toml
```diff
@@ -44,7 +44,7 @@ indexmap = { workspace = true, features = ["serde"] }
 indicatif = { workspace = true, optional = true }
 itertools = { workspace = true }
 merlin = { workspace = true, optional = true }
-nova-snark = { workspace = true, optional = true }
+nova-snark = { workspace = true, optional = true, features = [] }
```
iajoiner, line 47: Hmm you haven't turned off default features so this line didn't change anything, right?

iajoiner (changes requested):
Sorry please see the comment above

crates/proof-of-sql/Cargo.toml
iajoiner: @jacobtrombetta Well the reason why I flagged this is that this change doesn't actually do anything. If you don't want default features for at least some configuration of the `proof-of-sql` crate (such as when all features are false) you should put `default-features = false`. Then you can just enable `nova-snark` in the `blitzar` feature.

jacobtrombetta: I'm sorry, I think I'm a bit confused. 

`nova-snark` also has a `blitzar` feature flag - [see here](https://github.com/microsoft/Nova/blob/ad4d77ac89d6bbe9ef943056806e65ceb4ba3b3e/Cargo.toml#L30). I want to make sure the `blitzar` feature flag is defined in `nova-snark` if both `hyperkzg_proof` and `blitzar` are defined in `proof-of-sql`. If only `blitzar` is defined, I don't want to include `nova-snark`. I also don't want to disable any default features in `nova-snark`.

Perhaps it is best to make this `nova-snark = { workspace = true, optional = true, features = ["blitzar"]}`? I updated the code to do this, but will happily revert to whatever is best.

iajoiner (approved):

PR #792, opened 2025-04-24
bench: the benchmarks should use the `proof-of-sql-planner` crate when creating proof plans

jacobtrombetta:
# Rationale for this change
The next step in improving benchmarks is to use the `proof-of-sql-planner` crate when creating the proof plan. This PR moves all the old benchmarks to new proof planner. Additionally, two new queries are added to the benchmarks along with a number of quality of code improvements. 

- The `proof-of-sql-benches` crate now only uses `proof-of-sql-planner` for proof plans.
- Random column generation for `Decimal75` and `TimeStampTZ` column types are added.
- The query module is refactored.
- Two new queries are added to the query module.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql-benches/src/main.rs
+++ b/crates/proof-of-sql-benches/src/main.rs
@@ -33,6 +33,7 @@ use blitzar::{compute::init_backend, proof::InnerProductProof};
 use bumpalo::Bump;
 use clap::{ArgAction, Parser, ValueEnum};
 use curve25519_dalek::RistrettoPoint;
+use datafusion::config::ConfigOptions;
 use halo2curves::bn256::G2Affine;
 use nova_snark::{
     provider::{
@@ -44,7 +45,7 @@ use nova_snark::{
 use proof_of_sql::{
     base::{
         commitment::{Commitment, CommitmentEvaluationProof},
-        database::LiteralValue,
+        database::TableTestAccessor,
     },
     proof_primitive::{
         dory::{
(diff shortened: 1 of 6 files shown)


crates/proof-of-sql-benches/src/main.rs
```diff
@@ -393,9 +397,9 @@ fn bench_dynamic_dory(cli: &Cli, queries: &[QueryEntry]) {
 
     bench_by_schema::<DynamicDoryCommitment, DynamicDoryEvaluationProof>(
         "Dynamic Dory",
-        &Bump::new(),
         cli,
         queries,
+        &&prover_setup,
```
iajoiner, line 402: Are you sure having two `prover_setup`-like parameters in one function call is correct?

jacobtrombetta: I added it due to template confusion on my part. It looks much better now, thank you!

crates/proof-of-sql-benches/src/utils/queries.rs
```diff
@@ -1,20 +1,39 @@
+//! Queries used for benchmarking
+//!
+//! To add a new query,
+//! 1. Implement the `BaseEntry` trait
+//! 2. Add the new query to the `all_queries` function.
+//! 3. Add the new query to the `proof-of-sql-benches/main.rs` `Query` enum.
+//! 4. Add the new query to the `proof-of-sql-benches/main.rs` `Query` `to_string()` impl.
```
iajoiner, line 7: :)

iajoiner (approved):
@jacobtrombetta Really thanks!

PR #798, opened 2025-04-28
docs: update benchmarks in readme

jacobtrombetta:
# Rationale for this change
We ran benchmarks on four new queries: Filter, Complex Filter, Group By, and Join using the HyperKZG commitment scheme on the Single and Multi A100 VMs. This PR updates the benchmark images in the README to display the results.

- The README is updated to show the new queries that were benchmarked.
- New result graphs replace the previous results.

# Are these changes tested?
Yes


crates/proof-of-sql/README.md
```diff
@@ -137,32 +137,17 @@ a (BIGINT) | b (BIGINT) | c (VARCHAR)
...
-<p align="center">
-
-`b` | `sum_a` | `c`
----|---|---
-1 | -45585 | 301
-2 | -137574 | 300
-3 | -107073 | 282
-
-</p>
+Then, we run the following 10 queries against these data, prove, and verify the results:
```
iajoiner, line 140: Hmm there are only 4 queries here?

jacobtrombetta: Whoops, you are right, only 4. Thank you!

iajoiner (approved):

PR #823, opened 2025-05-07
fix: fix decimal casting issue

iajoiner:

# Rationale for this change

# Are these changes tested?

### DIFF
--- a/crates/proof-of-sql-planner/tests/e2e_tests.rs
+++ b/crates/proof-of-sql-planner/tests/e2e_tests.rs
@@ -162,15 +162,18 @@ fn test_tableless_queries() {
 fn test_simple_filter_queries() {
     let alloc = Bump::new();
     let sql = "select id, name from cats where age > 2;
-    select * from cats;
-    select name == $1 as name_eq from cats;
-    select 2 * age as double_age from cats";
+     select * from cats;
+     select name == $1 as name_eq from cats;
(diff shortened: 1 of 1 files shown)


crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -118,12 +121,24 @@ pub fn expr_to_proof_expr(
...
+                    // unless the source type is a timestamp. Otherwise, use a cast.
+                    if from_expr_scale != to_type_scale
+                        && !matches!(from_expr.data_type(), ColumnType::TimestampTZ(_, _))
+                    {
+                        Ok(DynProofExpr::try_new_scaling_cast(from_expr, to_type)?)
+                    } else {
+                        // Use a cast if the scale of the source and target types are the same
+                        // or if the source type is a timestamp.
+                        Ok(DynProofExpr::try_new_cast(from_expr, to_type)?)
+                    }
```
stuarttimwhite, line 141: Is there a reason we can't just try cast, adn then try scale cast if that fails?

iajoiner: @stuarttimwhite We are not currently doing that, right?

stuarttimwhite: I think we do support casting Timestamp to Timestamp. The function `try_scale_cast_types` allows it.

PR #849, opened 2025-05-19
perf: add if_rayon macro to the Scalar to CommittableColumn type conversion

jacobtrombetta:
# Rationale for this change
Conversion from the `Scalar` column type to a `CommitableCoumn` type in the `committable_columns` module offers an opportunity for parallelism. This PR adds the `if_rayon` macro to the `from` `impl` for the `Scalar` column type.

Benchmarks for the `Filter` query indicates a `2.14x` performance increase in `EqualsExpr::final_round_evaluate`. A `1.07x` performance increase overall.

Before, `EqualsExpr::final_round_evaluate` was `104.15ms`.
image

After,  `EqualsExpr::final_round_evaluate` is `48.57ms`.
image

To reproduce these benchmarks: `cargo run --release --bin proof-of-sql-benches -- -s hyper-kzg -q filter`.

- The `from` `impl` for the `Scalar` column type now uses the `if_rayon` macro

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -1,5 +1,6 @@
 use crate::base::{
     database::{Column, ColumnType, OwnedColumn},
+    if_rayon,
     math::decimal::Precision,
     posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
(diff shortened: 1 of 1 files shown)


iajoiner (approved):
If this does measurably improve performance, yes

PR #852, opened 2025-05-19
Perf: avoid creating a temporary vec in slice ops mul add

jacobtrombetta:
# Rationale for this change
The `mul_add_assign` module has a `mul_add_assign` function that is only used by the `MultilinearExtension` `mul_add` function. When calling `mul_add_assign`, a temporary vector is created with the `slice_ops::slice_cast(self) call. We don't need to create this temporary vector. Avoiding this creation will improve performance and memory allocation.

Before, `24.35ms`
image

After, `15.72ms`, `1.55x` improvement
image

- `Sync` is added to the `MultilinearExtension` trait to allow for parallelism
- `mul_add_assign` is updated to not require a temporary vector
- The `MultilinearExtension` trait no longer creates a temporary vector when calling `mul_add_assign`
- Tests are updated to call `mul_add_assign` correctly

# Are these changes tested?
Yes


iajoiner (approved):

crates/proof-of-sql/src/base/slice_ops/mul_add_assign_test.rs
```diff
@@ -3,20 +3,47 @@ use crate::base::scalar::test_scalar::TestScalar;
 
 #[test]
 fn test_mul_add_assign() {
-    let mut a = vec![1, 2, 3, 4];
+    let mut a = vec![
+        TestScalar::from(1i32),
+        TestScalar::from(2i32),
+        TestScalar::from(3i32),
+        TestScalar::from(4i32),
+    ];
```
stuarttimwhite, line 11: NIT: This could be `[1,2,3,4].map(TestScalar::from).to_vec()`

stuarttimwhite (dismissed):

iajoiner (approved):

PR #874, opened 2025-06-03
bench: add tracing to support test-net and main-net performance debuging

jacobtrombetta:
# Rationale for this change
Add more tracing to help identify the prover performance issue currently in MAIN and TEST nets.

NOTE: this branch is built off of v0.99.0.

- Traces are added to the section of code dominating 79% of the execution time in the logs.

# Are these changes tested?
Yes, locally.


tlovell-sxt:
These spans would probably be good to have in proof-of-sql regardless. Now that we've figured out the main reason why dependents were slow, maybe it'd be worth just updating this to add this tracing to latest instead of 0.99?

jacobtrombetta:
> These spans would probably be good to have in proof-of-sql regardless. Now that we've figured out the main reason why dependents were slow, maybe it'd be worth just updating this to add this tracing to latest instead of 0.99?

@tlovell-sxt good idea. Latest does have a span around the `mul_add` section of code. Other spans benchmark under `10 μs`, so I will plan on leaving those out to not flood the logs.

PR #890, opened 2025-06-09
Fix/slice exec try into proof plan

stuarttimwhite:

# Rationale for this change

# Are these changes tested?


crates/proof-of-sql-planner/tests/evm_tests.rs
```diff
@@ -468,3 +469,29 @@ fn we_can_verify_a_complex_filter_using_the_evm() {
...
+
+    let accessor = OwnedTableTestAccessor::<HyperKZGCommitmentEvaluationProof>::new_from_table(
+        TableRef::new("namespace", "table"),
+        owned_table([bigint("a", [5, 3, 2, 5, 3, 2, 102, 104, 107, 108])]),
+        0,
+        &ps[..],
+    );
+    let statements = Parser::parse_sql(
+        &GenericDialect {},
+        "SELECT COUNT(a) as a_count FROM namespace.table GROUP BY a",
```
iajoiner, line 487: We can add a slice to the query

crates/proof-of-sql/src/sql/evm_proof_plan/plans.rs
```diff
@@ -542,6 +543,7 @@ mod tests {
             &evm_slice_exec,
             &indexset![table_ref.clone()],
             &indexset![column_ref_a.clone(), column_ref_b.clone()],
+            &IndexSet::default(),
```
iajoiner, line 546: Hmm we should have output column names, right?

crates/proof-of-sql-planner/tests/evm_tests.rs
stuarttimwhite: I was unable to get that to work just using a LIMIT.

iajoiner: Hmm why?

stuarttimwhite: Because GROUP BY isn't composable, and I needed to use something that uses aliases as the inner plan.

iajoiner (approved):

crates/proof-of-sql-planner/tests/evm_tests.rs
```diff
@@ -681,3 +682,29 @@ fn we_can_verify_simple_slice_exec_using_the_evm() {
         .verify(&EVMProofPlan::new(plan.clone()), &accessor, &&vk, &[])
         .unwrap();
 }
+
+#[ignore = "This test requires the forge binary to be present"]
+#[test]
+#[expect(clippy::missing_panics_doc)]
+fn we_can_verify_a_slice_exec_using_the_evm() {
```
iajoiner, line 689: Hmm if there is no slice could you please rename it?

stuarttimwhite: This is just a test. I don't want to spend too much time on it. It does what it needs to, so I'd rather not spend any more time on it.

PR #894, opened 2025-06-11
refactor: split out `fold_expr_evals` and `fold_final_round_mles`

iajoiner:

# Rationale for this change

# Are these changes tested?


solidity/src/proof_gadgets/FoldUtil.pre.sol
```diff
@@ -0,0 +1,253 @@
...
+            function inequality_expr_evaluate(expr_ptr, builder_ptr, chi_eval) -> expr_ptr_out, result_eval {
+                revert(0, 0)
+            }
+            // slither-disable-start cyclomatic-complexity
+            // IMPORT-YUL ../proof_exprs/ProofExpr.pre.sol
+            function proof_expr_evaluate(expr_ptr, builder_ptr, chi_eval) -> expr_ptr_out, eval {
+                revert(0, 0)
+            }
+            // slither-disable-end cyclomatic-complexity
+            function fold_expr_evals(plan_ptr, builder_ptr, input_chi_eval, beta, column_count) -> plan_ptr_out, fold {
```
stuarttimwhite, line 162: I'm a little torn on these functions. We do have the folding logic in MathUtil, if you pass in the evaluations into the function. We also have other places (projection for example) where we need to evaluate a collection of proof evaluations without folding. This function is basically a combination of those two. This has the advantage of using one for loop instead of two, which might be worth it. I'll let you decide what you want to do. I just wanted to make sure you were aware of those two existing (or soon to exist) functions.

PR #905, opened 2025-06-17
bench: add queries to cover currently supported syntax

jacobtrombetta:
# Rationale for this change
Benchmarks should cover all of the supported [SQL syntax](https://github.com/spaceandtimefdn/sxt-proof-of-sql/blob/6f0938f5ca72b7d2475eb39fa31c47c7eca618ce/docs/SQLSyntaxSpecification.md#posql-sql-syntax). This PR adds benchmarks that cover the remainder of the syntax. These benchmarks will be used to identify performance optimizations that can be made.

- Support for random Uint8 columns are added to the benchmark crate
-  Benchmark queries are added to cover all of the supported SQL syntax listed here https://github.com/spaceandtimefdn/sxt-proof-of-sql/blob/6f0938f5ca72b7d2475eb39fa31c47c7eca618ce/docs/SQLSyntaxSpecification.md#posql-sql-syntax
- The SQL syntax was cleaned up for the all of the benchmark queries

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql-benches/src/utils/queries.rs
+++ b/crates/proof-of-sql-benches/src/utils/queries.rs
@@ -446,9 +446,9 @@ impl BaseEntry for Join {
     }
 
     fn sql(&self) -> &'static str {
-        "SELECT bench_table.a, bench_table_2.a \
-         FROM bench_table \
-         JOIN bench_table_2 on bench_table.a=bench_table_2.a;"
+        r"SELECT bench_table.a, bench_table_2.a
+          FROM bench_table
+          JOIN bench_table_2 on bench_table.a=bench_table_2.a;"
     }
 
     fn tables(&self) -> Vec<TableDefinition> {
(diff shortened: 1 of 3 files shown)


crates/proof-of-sql-benches/src/utils/queries.rs
```diff
@@ -400,24 +400,17 @@ impl BaseEntry for Coin {
...
+         SUM(
+         (CAST (to_address = $1 as bigint) - CAST (from_address = $1 as bigint))
+         * value * CAST(timestamp AS bigint)
+         ) AS weighted_value,
+         SUM(
+         (CAST (to_address = $1 as bigint) - CAST (from_address = $1 as bigint))
+         * value
+         ) AS total_balance,
+         COUNT(1) AS num_transactions
+         FROM bench_table;"
```
stuarttimwhite, line 413: I think I tend to prefer seeing these kinds of changes in a separate PR (unless the lint is complaining), just because it makes review a little more difficult when there's formatting changes along with new code. Not a big deal, just my personal preference. I can still review the rest.

PR #924, opened 2025-06-25
refactor: move shift logic from monotonic code to shift code

stuarttimwhite:

# Rationale for this change

There is logic in the monotonic gadget that should be in the shift gadget.

Moving creation and retrieval of shifted columns and chis into the shift gadget.

# Are these changes tested?
Yes

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_gadgets/shift.rs
+++ b/crates/proof-of-sql/src/sql/proof_gadgets/shift.rs
@@ -16,17 +16,36 @@ pub(crate) fn first_round_evaluate_shift<S: Scalar>(
     builder: &mut FirstRoundBuilder<'_, S>,
     num_rows: usize,
 ) {
-    // Note that we don't produce chi eval lengths here
-    // since it needs to be done in uniqueness check which uses shifts.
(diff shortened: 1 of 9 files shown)


crates/proof-of-sql/src/sql/proof_gadgets/shift.rs
```diff
@@ -16,17 +16,36 @@ pub(crate) fn first_round_evaluate_shift<S: Scalar>(
...
 }
 
+/// Perform final round evaluation of downward shift.
+pub(crate) fn final_round_evaluate_shift<'a, S: Scalar>(
+    builder: &mut FinalRoundBuilder<'a, S>,
+    alloc: &'a Bump,
+    alpha: S,
+    beta: S,
+    column: &'a [S],
+) -> &'a [S] {
```
iajoiner, line 31: Do you think we can get rid of `final_round_evaluate_shift_base`? If it is only called once we don't need it.

stuarttimwhite: Maybe later. For now, this is the easiest way to keep the unit tests unchanged. I want to focus on getting the monotonic and shift code working together the way it should be working together first.

PR #939, opened 2025-06-27
refactor: rearrange rust group by

JayWhite2357:
# Rationale for this change

In order to align the Rust with the Solidity verifier, we need to rearrange the Rust verifier and prover.

Refactoring of the Group By

# Are these changes tested?
By existing tests.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_plans/group_by_exec.rs
+++ b/crates/proof-of-sql/src/sql/proof_plans/group_by_exec.rs
@@ -180,6 +180,13 @@ impl ProofPlan for GroupByExec {
                 output_chi_eval,
             )?;
         }
+        let g_out_star_eval = builder.try_consume_final_round_mle_evaluation()?;
+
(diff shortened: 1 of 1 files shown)


crates/proof-of-sql/src/sql/proof_plans/group_by_exec.rs
```diff
@@ -126,16 +126,26 @@ impl ProofPlan for GroupByExec {
...
+            SumcheckSubpolynomialType::Identity,
+            g_in_star_eval + g_in_star_eval * g_in_fold_eval - input_chi_eval,
+            2,
+        )?;
+        // End compute g_in_star
+
+        let where_eval =
+            self.where_clause
+                .verifier_evaluate(builder, &accessor, input_chi_eval, params)?;
+
```
stuarttimwhite: NIT: would be nice if this was two commits. One for moving where, and one for moving the g_in values.

stuarttimwhite (approved):

PR #963, opened 2025-07-18
feat(sol): add scaling cast expr to solidity

iajoiner:
Add scaling cast expr to solidity

### DIFF
--- a/crates/proof-of-sql-planner/tests/evm_tests.rs
+++ b/crates/proof-of-sql-planner/tests/evm_tests.rs
@@ -374,6 +374,44 @@ fn we_can_verify_a_filter_with_arithmetic_using_the_evm() {
 #[ignore = "This test requires the forge binary to be present"]
 #[test]
 #[expect(clippy::missing_panics_doc)]
+fn we_can_verify_a_filter_with_arithmetic_with_scaling_using_the_evm() {
+    let (ps, vk) = load_small_setup_for_testing();
+
+    let accessor = OwnedTableTestAccessor::<HyperKZGCommitmentEvaluationProof>::new_from_table(
+        TableRef::new("namespace", "table"),
+        owned_table([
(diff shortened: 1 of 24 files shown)


crates/proof-of-sql/src/sql/evm_proof_plan/exprs.rs
```diff
@@ -648,6 +654,58 @@ impl EVMCastExpr {
...
+        column_refs: &IndexSet<ColumnRef>,
+    ) -> EVMProofPlanResult<ScalingCastExpr> {
+        let expr = ScalingCastExpr::try_new(
+            Box::new(self.from_expr.try_into_proof_expr(column_refs)?),
+            self.to_type,
+        )?;
+        if self.scaling_factor != expr.scaling_factor() {
+            return Err(EVMProofPlanError::IncorrectScalingFactor);
+        }
+        Ok(expr)
```
stuarttimwhite, line 705: I like something like `(self.scaling_factor == expr.scaling_factor()).then_some(expr).ok_or(EVMProofPlanError::IncorrectScalingFactor)`

crates/proof-of-sql-planner/tests/evm_tests.rs
```diff
@@ -371,6 +371,44 @@ fn we_can_verify_a_filter_with_arithmetic_using_the_evm() {
...
+        owned_table([
+            bigint("a", [5, 3, 2, 5, 3, 2]),
+            decimal75("b", 20, 2, [0, 100, 200, 300, 400, 500]),
+        ]),
+        0,
+        &ps[..],
+    );
+    let statements = Parser::parse_sql(
+        &GenericDialect {},
+        "SELECT a, b FROM namespace.table WHERE a + b = a - b",
```
stuarttimwhite, line 391: This query doesn't test scaling effectively. The filter is equivalent to `b = 0`, which will have the same result even if you scale by the wrong value.

PR #966, opened 2025-07-21
feat: add union to Solidity

iajoiner:
- Add `EVMUnionExec` to `evm_proof_plan`
- Add `UnionExec` to Solidity
- Add respective tests including cross-language tests
- Switch to `large-8-core-32gb-22-04` for foundry since slither is killed on the previous runner

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -194,7 +194,7 @@ jobs:
 
   foundrycheck: # Modified from the foundry book: https://book.getfoundry.sh/config/continuous-integration
     name: Foundry project
-    runs-on: ubuntu-latest
+    runs-on: large-8-core-32gb-22-04
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4


solidity/src/proof_plans/UnionExec.pre.sol
```diff
@@ -0,0 +1,478 @@
...
+                        // Consume chi evaluation for output
+                        output_chi_eval := builder_consume_chi_evaluation(builder_ptr)
+
+                        let d_star := builder_consume_final_round_mle(builder_ptr)
+                        // d_star + d_fold * d_star - output_chi = 0
+                        builder_produce_identity_constraint(
+                            builder_ptr,
+                            submod_bn254(addmod_bn254(d_star, mulmod_bn254(d_fold, d_star)), output_chi_eval),
+                            2
+                        )
```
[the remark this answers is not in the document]
iajoiner: @JayWhite2357 We will need to add one. No it doesn't. This will be a separate PR.

solidity/src/proof_plans/UnionExec.pre.sol
```diff
@@ -0,0 +1,477 @@
...
+                plan_ptr := add(plan_ptr, UINT64_SIZE)
+                if lt(input_count, 2) {
+                    // Not enough input plans for union
+                    err(ERR_UNION_NOT_ENOUGH_INPUT_PLANS)
+                }
+
+                // Initialize zero-sum constraint
+                zerosum_constraint := 0
+                output_length := 0
+                num_columns := 0
```
stuarttimwhite, line 406: This line is unnecessary. `input_count` is guaranteed to be at least 2.

solidity/src/proof_plans/UnionExec.pre.sol
```diff
@@ -0,0 +1,477 @@
...
+                // Process each input plan
+                for {} input_count { input_count := sub(input_count, 1) } {
+                    // Recursively evaluate the input plan
+                    let input_evaluations_ptr
+                    let input_length
+                    let input_chi_eval
+                    plan_ptr, input_evaluations_ptr, input_length, input_chi_eval :=
+                        proof_plan_evaluate(plan_ptr, builder_ptr)
+                    output_length := add(output_length, input_length)
+                    num_columns := mload(input_evaluations_ptr)
```
stuarttimwhite, line 417: I don't love that this value is being loaded from memory more than once. It only needs to be done the first time.

solidity/src/proof_plans/UnionExec.pre.sol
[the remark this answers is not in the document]
stuarttimwhite: Ideally there will be another function in `FoldLogExpr` for folding directly from first round mles.

solidity/src/proof_plans/UnionExec.pre.sol
iajoiner: @stuarttimwhite I don't think such guarantees exist if someone plugs in their custom-built `ProofPlan`s

stuarttimwhite (approved):

PR #984, opened 2025-08-04
feat: disable default features of `nova-snark` in order to disable `halo2curves/asm` which breaks wasm

iajoiner:
See title


stuarttimwhite (approved):

jacobtrombetta:
@iajoiner @JayWhite2357 this does degrade performance, I'm seeing approximately a 1.07x degredation in Nova's performance, which means it's 20-40ms slower. In the example below the overall Filter query performance, not just Nova, dropped 1.05x.

Below is an example

Nova before - 487.87ms
image

Nova after - 519.86ms - 1.07x performance degradation of Nova
image

PR #1009, opened 2025-08-12
docs: add protocols for most ProofExprs

iajoiner:
ProofExpr docs for all `ProofExpr`s but inequality, cast and scale cast (do the latter two need docs?)


docs/protocols/ProofExpr/add.tex
```diff
@@ -0,0 +1,43 @@
...
+\section{Definition}
+\noindent Let $a$ and $b$ be two columns with length $n$. $c$ is the sum $a + b$ if and only if $C(c) = C(a) + C(b)$.
+\section{Constraints}
+\begin{itemize}
+    \item Inputs: Input columns $a$, $b$.
+    \item Assumptions: The length of $a$ and $b$ is confirmed to be $n$.
+    \item Outputs: The sum $a + b$.
+    \item Hints and gadgets: None.
+    \item Challenges: None.
+    \item Constraints: None.
```
stuarttimwhite, line 37: I think technically C=A+B is a constraint here. It just isn't a degree 2 constraint, so we don't need an identity constraint. We never actually have to check it in our code because we just return A+B.

docs/protocols/ProofExpr/eq.tex
```diff
@@ -0,0 +1,59 @@
...
+\end{cases}$ if and only if the constraints hold.
+\section{Constraints}
+\begin{itemize}
+    \item Inputs: Input columns $a$, $b$.
+    \item Assumptions: The length of $a$ and $b$ is confirmed to be $n$.
+    \item Outputs: The column $s$.
+    \item Hints and gadgets: The pseudo-inverse $d$.
+    \item Challenges: None.
+    \item Constraints:
+    Let $d$ be the pseudo-inverse of $a-b$, that is,
```
stuarttimwhite, line 42: Don't define the hints in the constraints. The hints should only be defined within the Completeness proof.

docs/protocols/ProofExpr/eq.tex
```diff
@@ -0,0 +1,59 @@
...
+    $d_i := 
+\begin{cases}
+0, & \text{if } a_i = b_i,\\
+1/(a_i-b_i), & \text{if } a_i \neq b_i.
+\end{cases}$
+    The following constraints must hold:
+     \begin{eqnarray}
+    (a_i-b_i)s_i&=&0\\
+    (a_i-b_i)d_i&=&\chi_i-s_i
```
stuarttimwhite, line 51: The length of chi should be specified here. Maybe $\chi_n(i)$?

docs/protocols/ProofExpr/eq.tex
```diff
@@ -0,0 +1,59 @@
...
+\end{eqnarray}
+\end{itemize}
+\section{Completeness}
+Both constraints hold trivially due to the definition of $s$ and $d$.
+\section{Soundness}
+If (1) holds then either $a_i=b_i$ or $s_i=0$. If $a_i=b_i$ then (2) leads to $s_i=1$.
+The inverse is trivial.
```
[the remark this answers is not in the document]
stuarttimwhite: I agree with the comment here, but the suggested change is unnecessarily long. I would just say that if $s\ne 0$, then $(a_i-b_i)d_i=1$. So $a_i-b_i\ne 0$. So $a_i\ne b_i$.

docs/protocols/ProofExpr/not.tex
```diff
@@ -0,0 +1,43 @@
...
+
+\title{Not Expr}
+\author{MakeInfinite Inc}
+\date{Aug 2025}
+
+\begin{document}
+\maketitle
+
+\section{Definition}
+\noindent Let $a$ be a column with length $n$. $c$ is NOT $a$ if and only if $C(c) = C(\chi_n) - C(a)$.
```
stuarttimwhite, line 29: This constraint should be in the constraints.

docs/protocols/ProofExpr/not.tex
```diff
@@ -0,0 +1,43 @@
...
+\section{Constraints}
+\begin{itemize}
+    \item Inputs: Input column $a$.
+    \item Assumptions: The length of $a$ is confirmed to be $n$.
+    \item Outputs: The difference $\chi_n - a$.
+    \item Hints and gadgets: None.
+    \item Challenges: None.
+    \item Constraints: None.
+\end{itemize}
+\section{Completeness}
```
stuarttimwhite, line 39: I think you can just say "Trivial" for both completeness and soundness

docs/protocols/ProofExpr/or.tex
```diff
@@ -0,0 +1,43 @@
...
+
+\title{Or Expr}
+\author{MakeInfinite Inc}
+\date{Aug 2025}
+
+\begin{document}
+\maketitle
+
+\section{Definition}
+\noindent Let $a$ and $b$ be two columns with length $n$. $c$ is $\chi_n - (\chi_n-a) \odot (\chi_n-b) = a + b - a \odot b $ if and only if the constraints hold.
```
stuarttimwhite, line 29: This definition of $\lor$ here is way too complicated.

PR #1018, opened 2025-08-20
refactor!: remove `sql/postprocessing` from `proof_of_sql` crate

iajoiner:
# Rationale for this change
This is the second step of the process of removing the legacy parser

- Remove `sql/postprocessing`
- Add a few `expect`s for functions we may still want to use in the future.

# Are these changes tested?
Existing tests should pass.


crates/proof-of-sql/src/base/math/permutation.rs
```diff
@@ -29,6 +29,7 @@ pub struct Permutation {
 
 impl Permutation {
     /// Create a new permutation from a comparison function with the given length
+    #[expect(dead_code)]
```
stuarttimwhite, line 32: Should we be deleting this function now?

PR #1030, opened 2025-09-30
feat: add `GeneralizedFilterExec`

iajoiner:
A modern replacement for #797.


crates/proof-of-sql/src/sql/evm_proof_plan/plans.rs
```diff
@@ -73,6 +73,7 @@ impl EVMDynProofPlan {
                 )
                 .map(Self::SortMergeJoin)
             }
+            DynProofPlan::GeneralizedFilter(_) => Err(EVMProofPlanError::NotSupported),
```
stuarttimwhite: Can you add a test to hit this? I believe there used to be a test for this, before we had hit solidity/rust parity.

Issue #228, opened 2024-10-07
Make opaque `Scalar` conversions explicit rather than using `From`/`Into` trait.


JayWhite2357:
# Background and Motivation
Currently, the `Scalar` trait requires various conversions using the `From` and `Into` traits. Some of these are confusing, while others are clear.
In general, the rule should be that for types with a natural embedding into the associated `Scalar` field, `Scalar` should implement `From`. In all other cases, the conversion should either be auto-implemented, or a trait method, where appropriate.

**NOTE**: The following issue also necessitates changes to Scalar, so be aware of potential merge conflicts:
https://github.com/spaceandtimelabs/sxt-proof-of-sql/issues/234

# Changes Required

Each of these should be a separate PR.
- [ ] Remove `Into<[u64; 4]>`, `From<[u64; 4]>`, and `RefInto<[u64; 4]>` bounds.
    - Replace these with the trait methods `fn from_limbs(val: [u64; 4]) -> Self` and `fn to_limbs(&self) -> [u64; 4]`
    - These three conversions are the base conversions that are used for many other conversions. A `Scalar` is a field element (number mod some prime) with slightly less than 256 bits. Usually, `Scalar`s are internally stored in Montgomery form, so these conversions are non-free. However, they are needed in some situations.
- [ ] Remove `From<&'a String>`, `From<String>`, and `From<&'a str>` bounds. These three implementations ultimately rely on the conversions to and from limbs. As a result, these should have blanket/default implementations.
    - Replace these with `fn from_str_via_hash(val: &str) -> Self` and add a default implementation.
    - While it would be nice to `impl<S: Scalar> From<String> for S`, this is not allowed by the Rust type system. Solutions other than the default implementation of a trait method are welcome.
    - This conversion requires that the string is hashed in order to convert it to 256 bits which are ultimately converted into the `Scalar` itself. This is different from a parsing method.
- [ ] Remove the `VarInt` bound and replace it with a blanket `impl<S: Scalar> VarInt for S` implementation. The existing (and requested) implementation relies on the limb conversions only.

maujim:
@JayWhite2357 can you please take a look at the liked PR?

Dustin-Ray:
> [@JayWhite2357](https://github.com/JayWhite2357) can you please take a look at the liked PR?

[...]

PR #1175, opened 2026-05-09
refactor: blanket implement VarInt for Scalar

grantf04:
## Summary

- remove the direct `VarInt` supertrait bound from `Scalar`
- generalize scalar varint helpers and zig-zag conversion over any `S: Scalar`
- add a blanket `impl<S: Scalar> VarInt for S`, leaving the concrete `MontScalar` behavior covered through the generic implementation

This addresses the third checkbox in #228:

> Remove the `VarInt` bound and replace it with a blanket `impl<S: Scalar> VarInt for S` implementation.

/claim #228

## Verification

- `cargo fmt --check`
- `cargo check -p proof-of-sql --lib --no-default-features`
- `cargo test -p proof-of-sql --no-default-features --features std,test base::encode --lib`

Note: I also tried the default-feature check on Apple Silicon, but it fails before reaching this patch because the repo enables `halo2curves/asm`, whose build script reports `Currently feature asm can only be enabled on x86_64 arch.`

### DIFF
--- a/crates/proof-of-sql/src/base/encode/scalar_varint.rs
+++ b/crates/proof-of-sql/src/base/encode/scalar_varint.rs
@@ -1,8 +1,7 @@
 use crate::base::{
     encode::{ZigZag, U256},
-    scalar::MontScalar,
+    scalar::Scalar,
 };
-use ark_ff::MontConfig;
 use core::cmp::{max, Ordering};
 
 /// This function writes the input scalar x as a varint encoding to buf slice
@@ -14,7 +13,7 @@ use core::cmp::{max, Ordering};
 ///
 /// crash:
 /// - in case N is bigger than `buf.len()`
-pub fn write_scalar_varint<T: MontConfig<4>>(buf: &mut [u8], x: &MontScalar<T>) -> usize {
+pub fn write_scalar_varint<S: Scalar>(buf: &mut [u8], x: &S) -> usize {
     write_u256_varint(buf, x.zigzag())
 }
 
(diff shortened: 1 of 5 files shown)


grantf04:
Addressed the Copilot review comments in the latest push:

- Added an explicit S::MAX_BITS <= 254 assertion before scalar zigzag encoding, matching the existing assumption that bit 255 is never set before multiplying by 2.
- Updated the stale dalek-specific docs in zigzag.rs and scalar_varint.rs to refer to the generic Scalar / S path.

Validation:
- rustfmt --check crates/proof-of-sql/src/base/encode/zigzag.rs crates/proof-of-sql/src/base/encode/scalar_varint.rs --edition 2021 passes.
- git diff --check passes.
- cargo test -p proof-of-sql base::encode:: --lib does not complete on this macOS arm64 machine because workspace dependencies fail before reaching the patch: halo2curves enables an x86_64-only asm feature and blitzar-sys reports Linux-only support.

BossChaos:
## Code Review — Bounty #560

**PR:** [refactor: blanket implement VarInt for Scalar](https://github.com/spaceandtimefdn/sxt-proof-of-sql/pull/1175) by @grantf04

- ✅ Refactoring

**Wallet:** `0xdaE5d307339074A24F579dB48e7c639359D94904`

*Code review under Bounty #560 — SxT Test Coverage*

MyTH-zyxeon:
Review-assist for `/claim #228`.

This PR should be evaluated under issue #228, not #560. The body carries `/claim #228`, and the implementation targets the third #228 checkbox: remove the direct `VarInt` supertrait bound from `Scalar` and provide a blanket `impl<S: Scalar> VarInt for S`.

Useful review points:
- The main shape is coherent for the #228 checkbox: scalar varint helpers are generalized from `MontScalar<T>` to `S: Scalar`, and `VarInt` moves to a blanket implementation.
- Copilot originally raised four concrete review points: the generic zigzag bit-255 assumption and stale dalek-specific docs in `zigzag.rs` / `scalar_varint.rs`.
- The current diff appears to address those points by asserting `S::MAX_BITS <= 254` before zigzag encoding and updating the stale docs to describe generic `Scalar` / `S`.
- The author also reported `cargo fmt --check`, `cargo check -p proof-of-sql --lib --no-default-features`, and `cargo test -p proof-of-sql --no-default-features --features std,test base::encode --lib` passing. The default-feature check was blocked locally by Apple Silicon / `halo2curves/asm`, so maintainer-side Linux CI remains the useful validation surface.

Risk / routing notes:
- One third-party comment routes this as #560, but that appears mismatched with both the PR body and the diff; this is #228 scalar-conversion/refactor work.
- The key maintainer question is whether `S::MAX_BITS <= 254` is the right invariant to encode for every supported `Scalar`, or whether the generic zigzag implementation needs a more explicit trait-level bound/contract.

Suggested maintainer next step: compare this blanket-`VarInt` slice with the other #228 PRs and decide whether it can land as the independent third-checkbox implementation.

PR #1294, opened 2026-05-11
test: cover sumcheck prover round guards

NoeFabris:
/claim #560

## Summary
- add focused tests for invalid `prove_round` protocol states
- cover first-round verifier challenge, missing later-round verifier challenge, and inactive zero-variable prover guards

## Coverage
- Parent commit (`1bf50a63`) targeted run: `prover_round.rs` was 0/61 lines, 0/10 functions, 0/105 regions covered.
- This branch targeted run: `prover_round.rs` is 31/79 lines, 5/14 functions, 48/137 regions covered.
- The after line count includes the added in-file tests; the relevant production delta is that the previously uncovered guard paths in `prove_round` are now exercised.

## Verification
- `cargo test -p proof-of-sql --lib proof_primitive::sumcheck::prover_round::tests -- --nocapture`
- `cargo llvm-cov test -p proof-of-sql --lib --summary-only --json --output-path /tmp/sxt_sumcheck_after.json -- proof_primitive::sumcheck::prover_round::tests`
- `cargo fmt --check`
- `git diff --check`


NoeFabris:
/claim #560

Adding an explicit claim comment because the claim is in the PR body but has not appeared in the Algora issue table yet.

PR #1303, opened 2026-05-11
test: cover group by aggregation helpers

antaloaalonso:
/claim #560

## Summary
- Add focused coverage for `aggregate_columns` rejecting mismatched input lengths.
- Cover remaining `SUM` integer branches and the non-numeric `SUM` panic path.
- Cover `MAX`/`MIN` aggregation for boolean, integer, decimal, and timestamp columns.

## Validation
- `cargo fmt --check`
- `cargo test -p proof-of-sql --lib --no-default-features --features="arrow" base::database::group_by_util::tests`
- `cargo test -p proof-of-sql --lib --no-default-features --features="arrow"` (`1057 passed`)
- `cargo llvm-cov --package proof-of-sql --lib --no-default-features --features="arrow" --lcov --output-path /tmp/sxt-group-by-after.lcov`
  - Local ARM-safe coverage for `group_by_util.rs` dropped from 25 uncovered source lines to 2 uncovered source lines.
- `cargo check -p proof-of-sql --all-targets --no-default-features --features="arrow"`

Note: the official repository coverage workflow runs all features on Ubuntu. I validated the broad local suite with the ARM-safe `arrow` feature set because macOS cannot build the Linux-only `blitzar-sys` target, and the all-features path also enables x86_64-only asm locally.

### DIFF
--- a/crates/proof-of-sql/src/base/database/group_by_util.rs
+++ b/crates/proof-of-sql/src/base/database/group_by_util.rs
@@ -371,3 +371,156 @@ where
             .min_by(super::super::scalar::ScalarExt::signed_cmp)
     }))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::base::{
+        math::decimal::Precision,
+        posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
+        scalar::test_scalar::TestScalar,
+    };
+
+    #[test]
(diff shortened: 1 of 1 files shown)


antaloaalonso:
Quick review/reward check-in: the PR checks are still green and the linked Algora claim is pending. Happy to make any focused changes needed to get this merged/rewarded.

PR #1314, opened 2026-05-11
test: cache Dory test setup

yuganoffl:
## Summary
- add cached Dory public parameters and verifier setup for tests
- reuse cached setup in the Dory commitment evaluation proof tests called out in #557
- keep production setup generation unchanged

## Notes
This reduces repeated `PublicParameters::test_rand` / `VerifierSetup::from` work in the slow IPA test modules while preserving the same proof checks. The negative verifier-size test intentionally keeps independent setup generation.

## Validation
- `git diff --check`
- I could not run `cargo test` locally because this environment does not have `cargo`, `rustup`, `nix`, Docker, or Podman installed.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dory_commitment_evaluation_proof_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dory_commitment_evaluation_proof_test.rs
@@ -1,6 +1,7 @@
 use super::{
-    test_rng, DoryEvaluationProof, DoryProverPublicSetup, DoryScalar, DoryVerifierPublicSetup,
-    ProverSetup, PublicParameters, VerifierSetup,
+    cached_test_public_parameters, cached_test_verifier_setup, test_rng, DoryEvaluationProof,
+    DoryProverPublicSetup, DoryScalar, DoryVerifierPublicSetup, ProverSetup, PublicParameters,
+    VerifierSetup,
 };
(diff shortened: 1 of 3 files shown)


MyTH-zyxeon:

This looks like a focused `/claim #557` performance slice. I checked the current diff and it is limited to the Dory commitment evaluation test surface plus test-only Dory helpers:

- `dory_commitment_evaluation_proof_test.rs` and `dynamic_dory_commitment_evaluation_proof_test.rs` replace repeated `PublicParameters::test_rand` / `VerifierSetup::from` setup with shared cached test fixtures for the same IPA proof checks.
- `proof_primitive/dory/mod.rs` adds test-only `OnceLock` helpers for `max_nu` 4 and 6, using deterministic seeded setup generation.
- The negative verifier-size test is intentionally left on independent setup generation, so the error-path coverage is not accidentally masked by the cache.

Current gate note: the PR is open/mergeable but still blocked on review/workflow approval. The head SHA `d53dd0e6f24d` has visible Orca checks green, while `CI-Lint-And-Test` and `CI-Code-Coverage` are both `action_required` and the combined status API reports `statuses=[]`. I do not see a code-test failure signal from the available checks yet. Since the author could only run `git diff --check`, the useful maintainer follow-up is to approve the fork workflows and run the Dory-focused cargo tests once CI is available.

PR #1322, opened 2026-05-11
test: cover ristretto commitment helper paths

keshav55:
## Summary
- add direct coverage for Ristretto transcript byte serialization
- assert the non-blitzar commitment path fails explicitly on builds without the blitzar feature
- keep the slice isolated to `proof_primitive/inner_product/ristretto_point.rs`

/claim #560

## Verification
- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p proof-of-sql --no-default-features --features="std" ristretto -- --nocapture`
- `cargo llvm-cov --package proof-of-sql --no-default-features --features="std" --lib --summary-only -- ristretto` -> `ristretto_point.rs` 100.00% line coverage

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/inner_product/ristretto_point.rs
+++ b/crates/proof-of-sql/src/proof_primitive/inner_product/ristretto_point.rs
@@ -62,4 +62,22 @@ mod tests {
             commitment2.to_transcript_bytes()
         );
     }
+
+    #[test]
+    fn transcript_bytes_are_compressed_ristretto_bytes() {
+        assert_eq!(RISTRETTO_BASEPOINT_POINT.to_transcript_bytes().len(), 32);
+        assert_eq!(
+            RISTRETTO_BASEPOINT_POINT.to_transcript_bytes(),
+            RISTRETTO_BASEPOINT_POINT.compress().as_bytes()
+        );
+    }
+
+    #[cfg(not(feature = "blitzar"))]
+    #[test]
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist note for `/claim #560` triage:

- Scope checked at head `46e8c3132babbe48140cfdea57a2107ac39ec1a7`: one test-only file, `crates/proof-of-sql/src/proof_primitive/inner_product/ristretto_point.rs`, with `+18/-0` and no production-path changes.
- The diff covers `RistrettoPoint::to_transcript_bytes()` length/content against compressed Ristretto bytes, and adds a `#[cfg(not(feature = "blitzar"))]` expected-panic guard for `compute_commitments` on no-blitzar builds.
- There is overlap with #2380 on the transcript-bytes assertion, but this PR has the extra no-blitzar commitment-path check. If maintainers want only one landing PR for this slice, #1322 is the broader of the two; otherwise #2380 can be treated as the narrower transcript-only variant.
- Current PR state still looks approval/runner-gated rather than code-failing: `OPEN` / `MERGEABLE`, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`; Orca checks pass. Raw check-runs for this head show `Prepare`, `Agent`, and `Upload results` success, with only `Cleanup artifacts` failing on the GitHub Actions run.

I did not run local Rust validation in this pass because the local checkout is unavailable here; this is a diff/status/deconfliction review only.

PR #1325, opened 2026-05-11
ci: drop redundant proof-of-sql checks

Rav-xyl:

# Rationale for this change

Issue #557 was clarified to include broad CI runtime improvements. The `check` job currently runs several `cargo check -p proof-of-sql ...` commands immediately before equivalent `cargo check -p proof-of-sql --all-targets ...` commands with the same feature set.

The `--all-targets` checks are the stricter/superset checks for those feature combinations, so the earlier package-only invocations add another Cargo process without increasing the validation surface.

/claim #557

- Drop the package-only `proof-of-sql` check before the `--all-targets --all-features` check.
- Drop the package-only `proof-of-sql` checks before the `--all-targets` checks for the `test`, `blitzar`, `arrow`, `rayon`, `perf`, and `std` feature-only lanes.
- Keep the existing `--all-targets` checks intact, so CI still checks the same feature combinations across all package targets.

# Are these changes tested?

Partially, with static/workflow validation plus local Cargo checks for the Windows-compatible retained lanes.

Static/workflow validation:

- `git diff --check origin/main...HEAD`
- `python -c "import yaml, pathlib; yaml.safe_load(pathlib.Path('.github/workflows/lint-and-test.yml').read_text()); print('yaml ok')"`
- `bash -lc "tr -d '\r' < scripts/check_commits.sh | bash"`
- Verified every removed package-only command still has the corresponding retained `--all-targets` command for the same feature lane.

Local Cargo checks using `rustc 1.91.1` / `cargo 1.91.1`:

- `cargo +1.91.1 check -p proof-of-sql --no-default-features`
- `rustup +1.91.1 target add thumbv7em-none-eabi`
- `cargo +1.91.1 check -p proof-of-sql --target thumbv7em-none-eabi --no-default-features`
- `cargo +1.91.1 check -p proof-of-sql --all-targets --no-default-features --features="std"`
- `cargo +1.91.1 check -p proof-of-sql --all-targets --no-default-features --features="test"`
- `cargo +1.91.1 check -p proof-of-sql --all-targets --no-default-features --features="arrow"`
- `cargo +1.91.1 check -p proof-of-sql --all-targets --no-default-features --features="rayon"`

Windows limitation:

- `cargo +1.91.1 check -p proof-of-sql --all-targets --all-features`
- `cargo +1.91.1 check -p proof-of-sql --all-targets --no-default-features --features="perf"`

Those two reached `blitzar-sys v1.115.1` and stopped in its build script with `Unsupported OS. Only Linux is supported.` The retained Linux-only `blitzar`/`all-features`/`perf` lanes should be covered by project CI.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -53,33 +53,19 @@ jobs:
       - name: Run cargo check (proof-of-sql) (no features)
         run: cargo check -p proof-of-sql --no-default-features
       - name: Run cargo check (proof-of-sql) (all features)
-        run: |
-          cargo check -p proof-of-sql --all-features
-          cargo check -p proof-of-sql --all-targets --all-features
+        run: cargo check -p proof-of-sql --all-targets --all-features
       - name: Run cargo check (proof-of-sql) (just "test" feature)
-        run: |
-          cargo check -p proof-of-sql --no-default-features --features="test"
-          cargo check -p proof-of-sql --all-targets --no-default-features --features="test"
+        run: cargo check -p proof-of-sql --all-targets --no-default-features --features="test"
       - name: Run cargo check (proof-of-sql) (just "blitzar" feature)
-        run: |
-          cargo check -p proof-of-sql --no-default-features --features="blitzar"
-          cargo check -p proof-of-sql --all-targets --no-default-features --features="blitzar"
+        run: cargo check -p proof-of-sql --all-targets --no-default-features --features="blitzar"
       - name: Run cargo check (proof-of-sql) (just "arrow" feature)
-        run: |
-          cargo check -p proof-of-sql --no-default-features --features="arrow"
-          cargo check -p proof-of-sql --all-targets --no-default-features --features="arrow"
+        run: cargo check -p proof-of-sql --all-targets --no-default-features --features="arrow"
       - name: Run cargo check (proof-of-sql) (just "rayon" feature)
-        run: |
-          cargo check -p proof-of-sql --no-default-features --features="rayon"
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer-facing review assist only; I am not adding a new claim.

Current state I verified for this `#557` CI-runtime slice:

- PR is `OPEN` / `MERGEABLE` / `BLOCKED` / `REVIEW_REQUIRED` at head `ce9e535b27c0cb9e7a2460d6b76af74a4048ca6c`.
- Diff is one workflow file, `.github/workflows/lint-and-test.yml`, `+7/-21`.
- The change removes the package-only `cargo check -p proof-of-sql ...` invocation immediately before the matching stricter `cargo check -p proof-of-sql --all-targets ...` invocation for `all-features`, `test`, `blitzar`, `arrow`, `rayon`, `perf`, and `std` lanes.
- The no-feature package check and the `thumbv7em-none-eabi` no-std target check are still present, so the intended review question is narrow: whether each removed package-only lane is fully covered by the retained `--all-targets` lane with the same feature set.
- Orca checks are green. GitHub Actions are still approval-gated: `CI-Lint-And-Test` and `CI-Code-Coverage` both show `action_required`, the jobs API returns `jobs=[]`, and combined commit status returns `statuses=[]`.
- There were no existing comments or review submissions when I checked this, so the blocker appears to be maintainer review / workflow approval rather than a visible failing check.

This looks like a small, differentiated `#557` candidate because it reduces redundant Cargo processes while preserving the stricter all-target feature coverage. The main maintainer review item is confirming that no package-only-only target behavior is being intentionally kept for those feature lanes.

PR #1338, opened 2026-05-11
test: cover varying byte matrix edge cases

admin1-design:
/claim #560

Adds focused coverage for `compute_varying_byte_matrix` edge cases:
- empty scalar input returns no varying columns and the expected byte distribution
- constant scalar input returns no varying columns while preserving byte distribution

Validation:
- `rustfmt crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs`
- `cargo test -p proof-of-sql byte_matrix_utils --lib --no-default-features --features test` (4 passed)
- `git diff --check`

Note: the default-feature test path reaches `blitzar-sys`, whose build script reports `Unsupported OS. Only Linux is supported` on this Windows shell. The focused no-default-features library test passes locally and upstream CI should cover the default Linux path.

### DIFF
--- a/crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs
+++ b/crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs
@@ -54,6 +54,32 @@ mod tests {
     }
 
     #[test]
+    fn we_can_compute_varying_byte_matrix_for_empty_scalars() {
+        let alloc = Bump::new();
+        let scalars = Vec::<TestScalar>::new();
+        let expected_byte_distribution = ByteDistribution::new::<TestScalar, TestScalar>(&scalars);
+
+        let (varying_columns, byte_distribution) =
+            compute_varying_byte_matrix::<TestScalar>(&scalars, &alloc);
+
+        assert_eq!(byte_distribution, expected_byte_distribution);
+        assert!(varying_columns.is_empty());
+    }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow `/claim #560` test-only PR in `crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs` (`+26/-0`). It adds two boundary cases for `compute_varying_byte_matrix::<TestScalar>()`: empty scalar input and constant scalar input. Both assert that no varying byte columns are emitted while the returned `ByteDistribution` matches `ByteDistribution::new(...)` for the same input.

The main acceptance check I would confirm is whether `ByteDistribution::new(...)` is enough as the expected-value oracle here, or whether maintainers want one explicit distribution fixture so the tests pin exact byte-count behavior independently of the helper. For the constant case, `42u64` is a clean no-varying-byte fixture; an additional non-zero high-byte or signed/negative constant would be a useful follow-up only if the bounty reviewer wants stronger edge coverage.

The stale bot marked this PR inactive today, but the patch still looks reviewable and claim-scoped. I did not run Rust tests locally. GitHub currently shows the Orca security scans succeeded; I did not find public Actions runs for head `38204360789388f0d6214f6d2c78ca5976cb0457`, and the combined commit status API returns `state=pending` with `statuses=[]`.

JayWhite2357:

PR #1349, opened 2026-05-11
Add explicit Scalar limb methods

yangyu0330:
﻿## Summary

Part of #228. This PR handles the first refactor item by making scalar limb conversion explicit on the `Scalar` trait.

- removes the `Into<[u64; 4]>`, `From<[u64; 4]>`, and `RefInto<[u64; 4]>` requirements from `Scalar`
- adds `Scalar::from_limbs` and `Scalar::limbs` for little-endian non-Montgomery limb conversion
- updates generic call sites to use those explicit methods instead of conversion trait bounds
- removes the now-unused crate-private `ref_into` helper module

The existing concrete `From<[u64; 4]>` / `From<MontScalar<_>> for [u64; 4]` impls remain as compatibility shims, but generic code no longer depends on them through the `Scalar` bound.

## Testing

- `cargo fmt --all -- --config imports_granularity=Crate,group_imports=One`
- `cargo check -p proof-of-sql --lib --no-default-features --features arrow,test`
- `cargo test -p proof-of-sql --lib --no-default-features --features arrow,test scalar`
- `git diff --check`

Note: I could not run default/all-features checks on this Windows environment because the default `perf` feature pulls in `blitzar`, and `blitzar-sys` exits from its build script on non-Linux hosts.


MyTH-zyxeon:
Maintainer review-assist note for #228 triage:

I did a read-only pass over this older, comment-free PR because it is a focused first-checkbox implementation for #228.

- Scope: 16 files, `+93/-106`, centered on removing generic limb conversion bounds from `Scalar`, adding explicit `from_limbs` / `limbs`, deleting `base/ref_into.rs`, and updating Arrow i256, commitment, transcript, literal, scalar-ext, scaling-cast, test utility, and range-check call sites.
- Bounty fit: the PR body says `Part of #228` and targets the first refactor item only. It does not contain an explicit `/claim #228`, so this looks like review-routing help rather than an additional claim signal.
- Review state: Copilot left a summary review and no inline comments. I do not see human review comments yet.
- CI state: Orca checks are green. The only failing Actions run I can see for head `7ca3ee13a37d3ca00923d0a8915de84363894e90` is Copilot code review's `Cleanup artifacts` job, which fails while deleting artifacts with `HTTP 403 Resource not accessible by integration`; the combined commit status reports `statuses=[]`. That points to a cleanup/permission gate, not a Rust/test failure in this PR.

No claim action from me here; just leaving the compact triage note so maintainers can compare this first-checkbox slice against the other #228 entries.

PR #1353, opened 2026-05-11
test: cover shift proof gadget

swhan0329:
/claim #560

## Summary
- add non-blitzar unit coverage for the shift proof gadget first-round witness construction
- verify final-round shift constraints row-by-row with the mock verification builder
- assert shifted-column materialization and auxiliary chi/rho lengths

## Validation
- cargo fmt --check
- cargo test -p proof-of-sql shift --no-default-features --features="rayon test"
- cargo test -p proof-of-sql proof_gadgets --no-default-features --features="rayon test"
- cargo llvm-cov -p proof-of-sql --no-default-features --features="rayon test" --json --summary-only --output-path target/shift-coverage-summary.json --no-clean -- shift

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_gadgets/shift.rs
+++ b/crates/proof-of-sql/src/sql/proof_gadgets/shift.rs
@@ -201,6 +201,112 @@ pub(crate) fn verify_shift<S: Scalar>(
     Ok((shifted_column_eval, chi_n_plus_1_eval))
 }
 
+#[cfg(test)]
+mod unit_tests {
+    use super::{final_round_evaluate_shift, first_round_evaluate_shift, verify_shift};
+    use crate::{
+        base::{
+            polynomial::MultilinearExtension,
+            scalar::{test_scalar::TestScalar, Scalar},
+        },
+        sql::proof::{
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This looks like a clean, test-only coverage slice for the shift gadget: one file (`crates/proof-of-sql/src/sql/proof_gadgets/shift.rs`), `+106/-0`, no prior comments/reviews/threads, Orca checks green, and the PR is `MERGEABLE` while still gated on `reviewDecision=REVIEW_REQUIRED` / `mergeStateStatus=BLOCKED`.

Two small merge-readiness items I would check before approval:

1. The body currently contains `/claim #560` twice. One marker should be enough, so removing the duplicate would reduce Algora/claim parsing ambiguity.
2. `final_round_shift_constraints_verify_for_each_row` builds a shifted column of length `column.len() + 1`, but `run_verify_for_each_row` is called with `column.len()`. If that helper intentionally verifies only the source-row domain, this is fine; otherwise this may miss the appended shifted boundary row and should be widened or covered by a direct boundary assertion.

No local code change requested from my side; this is otherwise a compact candidate for maintainer review.

PR #1370, opened 2026-05-12
test: cover dynamic matrix index boundaries

Mykidzou:
Contributes to #560
/claim #560

## Summary
- Adds focused unit coverage for dynamic matrix row/column boundary behavior.
- Verifies the last valid column in each tested row maps to the expected index.
- Verifies out-of-width columns are rejected and the reserved `(1, 0)` gap remains unmapped.

## Validation
- `cargo +1.91.1 fmt --check` passed
- `cargo +1.91.1 test -p proof-of-sql --no-default-features --features test matrix_structure --lib` passed: 8 passed, 0 failed
- `git diff --check` passed

No security research, no protocol changes, test-only coverage.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dynamic_matrix_utils/matrix_structure.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dynamic_matrix_utils/matrix_structure.rs
@@ -273,4 +273,25 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn we_reject_columns_outside_the_full_row_width() {
+        for row in 0..32 {
+            let width = full_width_of_row(row);
+            let row_start = row_start_index(row);
+
+            assert_eq!(
+                index_from_row_and_column(row, width - 1),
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer review-assist for bounty #560:

This PR looks like a narrow, test-only coverage addition for dynamic matrix index boundaries.

- It validates the last valid column in rows 0..31 maps to `row_start + width - 1`.
- It validates the first out-of-width column maps to `None` for each tested row.
- It pins the reserved `(1, 0)` gap behavior and the `(1, 1)` round trip.
- Orca IaC/Secrets/Vulnerabilities checks are green, and the PR is `MERGEABLE`; the remaining visible gate is `REVIEW_REQUIRED` / `mergeStateStatus=BLOCKED`.

Deconfliction note: #2385 appears to cover broader matrix index round trips, while this PR specifically covers out-of-width rejection and the reserved row-one gap. That seems useful as an independent boundary slice for #560.

PR #1399, opened 2026-05-12
test: cover transcript scalar paths

lewisohiohu:
## Summary
- Add direct coverage for `extend_as_le_from_refs` matching little-endian byte extension
- Add scalar transcript coverage for `extend_scalars_as_be` matching explicit scalar limbs
- Assert scalar challenges are masked into the scalar challenge range and remain stateful across consecutive challenges

## Verification
- `cargo test -p proof-of-sql --no-default-features --features test base::proof::transcript::tests`
- `cargo fmt --check`
- `git diff --check`

Note: default-feature tests on this Windows machine attempt to build `blitzar-sys`, which currently errors with `Unsupported OS. Only Linux is supported.` The targeted verification above avoids default `blitzar`/`perf` while keeping `std` through the repo's `test` feature.

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/base/proof/transcript.rs
+++ b/crates/proof-of-sql/src/base/proof/transcript.rs
@@ -71,7 +71,10 @@ pub trait Transcript {
 #[cfg(test)]
 mod tests {
     use super::Transcript;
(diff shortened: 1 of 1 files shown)


JayWhite2357:

PR #1403, opened 2026-05-12
test: cover inner product ref cast

INDICUMA:
## Summary
- add direct coverage for inner_product_ref_cast with signed integer inputs cast into TestScalar values
- cover the uneven-length behavior so the helper keeps the same zip semantics as the other inner-product helpers

## Verification
- cargo fmt --check
- git diff --check
- cargo test -p proof-of-sql --no-default-features --features test inner_product
- cargo llvm-cov clean --workspace
- cargo llvm-cov --no-default-features --features test --no-report -p proof-of-sql -- base::slice_ops::inner_product_test
- cargo llvm-cov report --summary-only | rg 'base/slice_ops/inner_product.rs|TOTAL'\n\nCoverage: base/slice_ops/inner_product.rs now reports 100.00% line coverage for the focused inner_product_test run.\n\n/claim #560


JayWhite2357:

PR #1412, opened 2026-05-12
Preserve nullable column metadata in planner schemas

Treasure520520:
/claim #183

## Summary

This PR adds the first nullable-column metadata plumbing needed by the nullable column support work:

- adds a nullable flag to `ColumnField`, defaulting existing fields to non-nullable
- adds `ColumnField::new_nullable` and `ColumnField::is_nullable`
- preserves `ColumnField` nullability when converting planner metadata into Arrow/DataFusion `Field`s
- updates planner tests so one field is explicitly nullable and must stay nullable in the Arrow schema

## Scope

This is intentionally a small, reviewable step rather than a full nullable proof implementation. It does not yet add nullable column storage, validity masks, nullable expression evaluation, or proof support. The goal is to stop planner/schema metadata from erasing nullability so the larger nullable-column implementation has a stable base to build on.

## Validation

Passed:

```bash
cargo fmt --all --check
cargo check -p proof-of-sql-planner
cargo check -p proof-of-sql --no-default-features --features 'rayon test'
```

I also attempted the focused planner unit tests, but local test linking is currently blocked by this machine's unaccepted Xcode license/macOS SDK state:

```text
unable to find dynamic system library 'objc'
unable to find dynamic system library 'iconv'
```

The compile/type checks above pass with the same code changes.

### DIFF
--- a/crates/proof-of-sql-planner/src/context.rs
+++ b/crates/proof-of-sql-planner/src/context.rs
@@ -95,7 +95,7 @@ impl PoSqlTableSource {
                     Field::new(
                         column_field.name().value.as_str(),
                         (&column_field.data_type()).into(),
-                        false,
+                        column_field.is_nullable(),
                     )
                 })
                 .collect::<Vec<_>>(),
@@ -144,14 +144,14 @@ mod tests {
         // Non-empty
         let column_fields = vec![
             ColumnField::new("a".into(), ColumnType::SmallInt),
-            ColumnField::new("b".into(), ColumnType::VarChar),
+            ColumnField::new_nullable("b".into(), ColumnType::VarChar),
         ];
         let table_source = PoSqlTableSource::new(column_fields);
         assert_eq!(
             table_source.schema().all_fields(),
             vec![
                 &Field::new("a", DataType::Int16, false),
-                &Field::new("b", DataType::Utf8, false),
+                &Field::new("b", DataType::Utf8, true),
             ]
         );
         assert_eq!(
--- a/crates/proof-of-sql-planner/src/util.rs
+++ b/crates/proof-of-sql-planner/src/util.rs
@@ -137,9 +137,12 @@ pub fn column_fields_to_schema(column_fields: Vec<ColumnField>) -> Schema {
         column_fields
             .into_iter()
             .map(|column_field| {
-                //TODO: Make columns nullable
                 let data_type = (&column_field.data_type()).into();
-                Field::new(column_field.name().value.as_str(), data_type, false)
+                Field::new(
+                    column_field.name().value.as_str(),
+                    data_type,
+                    column_field.is_nullable(),
+                )
             })
             .collect::<Vec<_>>(),
     )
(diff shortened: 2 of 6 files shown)


Treasure520520:
Added a small follow-up commit with direct ColumnField coverage: default non-nullability, explicit nullable construction, and serde backward compatibility for schemas serialized before the nullable field existed. Additional validation: `cargo fmt --all --check` and `cargo check -p proof-of-sql --tests --no-default-features --features 'rayon test'` pass locally. Runtime test execution remains blocked only by this machine's unaccepted Xcode/macOS SDK license at link time.

Treasure520520:
Added another focused follow-up commit (a513206) that preserves ColumnField nullability in the base Arrow Field conversion (`impl From<&ColumnField> for Field`) and covers both nullable and non-nullable conversions directly. This closes one more metadata-erasure path beyond the planner schema conversion already in the PR.

Validation:
- `cargo fmt --all --check` passes.
- `git diff --check` passes.
- `cargo check -p proof-of-sql --no-default-features --features "rayon test"` is currently blocked on this machine before crate type-checking by the unaccepted Xcode license during dependency build-script linking (exit status 69), same local environment blocker noted earlier.

Treasure520520:
Added one more follow-up commit (bd1489f) for the nullable-column groundwork: `OwnedColumn::try_from(&ArrayRef)` now returns the existing `NullNotSupportedYet` error immediately when an Arrow array has a non-empty null bitmap, instead of allowing type-specific paths to panic on nulls or silently read values while ignoring null validity. Added regression coverage for nullable Int64, Utf8, and Boolean arrays.

Validation:
- `cargo fmt --all --check` passes.
- `git diff --check` passes.
- `cargo check -p proof-of-sql --no-default-features --features "rayon test"` is still blocked locally before crate type-checking by the unaccepted Xcode license during dependency build-script linking (exit status 69).

MyTH-zyxeon:
Maintainer review-assist for `/claim #183` (no new claim from me): I rechecked this as a narrow nullable-column groundwork slice.

Current state observed just now:
- PR is `OPEN`, `MERGEABLE`, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`.
- Head is `bd1489f67d9d2f11b3631b5b8e1cd9a4870a2579`.
- Scope is 6 files, `+104/-9`: planner/schema nullability preservation, `ColumnField` nullable flag/default/serde compatibility, `Field::from(&ColumnField)` nullability preservation, and fail-closed `OwnedColumn::try_from(&ArrayRef)` handling for non-empty null bitmaps.
- Visible status rollup is green for the 3 Orca checks; combined commit statuses returned `statuses=[]`.
- `CI-Code-Coverage` and `CI-Lint-And-Test` are both `action_required` with `jobs=[]`, so I cannot inspect failing Actions logs.
- Existing conversation has author follow-up notes and no review submissions.

Why this seems useful for #183: it does not claim full nullable proof support, but it preserves nullable metadata instead of erasing it and explicitly rejects nullable Arrow arrays at the owned-column boundary until real nullable storage/proof support exists. That is a safer foundation than silently treating null slots as ordinary values.

Pre-merge checks I would want before approval/rerun: maintainer confirmation that this partial groundwork is acceptable for the broader nullable-column route, plus a fork-gated CI rerun because the author reports local checks but full Actions are still blocked on approval.

JayWhite2357:

PR #1413, opened 2026-05-12
test: cover byte matrix empty and constant columns

zxzok:
## Summary
- Add byte matrix coverage for empty scalar columns.
- Add byte matrix coverage for constant scalar columns with no varying byte columns.
- Keep the change test-only and scoped to `byte_matrix_utils`.

/claim #560

## Validation
- `rustfmt --check crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs`
- `cargo test -p proof-of-sql byte_matrix_utils --no-default-features --features test` -> 4 passed

## Note
- Full `cargo fmt --check` currently reports an unrelated import-order diff in `crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_helper.rs` on current main; this PR does not touch that file.

### DIFF
--- a/crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs
+++ b/crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs
@@ -39,6 +39,36 @@ mod tests {
     use bumpalo::Bump;
 
     #[test]
+    fn we_can_compute_varying_byte_matrix_for_empty_column() {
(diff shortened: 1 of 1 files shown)


zxzok:
Closing this duplicate slice because I consolidated the byte-matrix coverage into #1419, which already has the active #560 claim and updated validation notes.

PR #1441, opened 2026-05-12
Refactor scalar limb conversions into explicit methods

### LINKED ISSUE #228: Make opaque `Scalar` conversions explicit rather than using `From`/`Into` trait.

JimmiFriborg:
## Summary
- removes implicit `[u64; 4]`/`RefInto` bounds from `Scalar`
- adds explicit `Scalar::from_limbs` and `Scalar::to_limbs`
- updates generic scalar call sites to use the explicit methods
- removes the now-unused `RefInto` helper

## Validation
- `cargo fmt`
- `cargo check -p proof-of-sql --lib`

/claim #228

Closes #228.


JimmiFriborg:
Added `/claim #228` to make the Algora bounty claim explicit. This PR is ready for maintainer review for the limb-conversion scope.

MyTH-zyxeon:
Maintainer-facing review assist for `/claim #228`:

- I rechecked the current head `a9dc7c95c3dde4a5b3b9d718f3d97fcfad42073a`. The PR is still open/mergeable but blocked on review, carries `/claim #228`, has no maintainer review yet, and the only visible checks are the three Orca scans passing. `CI-Code-Coverage` and `CI-Lint-And-Test` are both `action_required` with no jobs exposed until approval.
- Scope matches the #228 acceptance direction: the generic `Scalar` trait no longer requires opaque `[u64; 4]` `From`/`Into` or `RefInto` bounds, adds explicit `from_limbs`/`to_limbs`, removes the `ref_into` helper module, and migrates the main production call sites through Column/CommittableColumn/LiteralValue/Transcript/ScalarExt/range-check/scaling paths.
- Review risks to check before merge: the `u256` file appears to include line-ending churn alongside the real `to_limbs` import/use, so maintainers may want to confirm the diff is intentional; the explicit `decimal_op_left_upcast::<S, S>` type parameter change should be validated against the relevant decimal comparison tests; and any external code relying on generic `Scalar: From<[u64;4]> + Into<[u64;4]>` bounds should be considered a deliberate API tightening rather than an accidental break.

I did not run cargo locally in this pass; this is status/diff triage to make maintainer review and claim routing easier.

JimmiFriborg:
Quick refresh on this bounty claim for #228.

I rechecked the current PR head locally:

- branch: `JimmiFriborg/sxt-proof-of-sql:sxt-228-explicit-scalar-conversions`
- head: `a9dc7c95c3dde4a5b3b9d718f3d97fcfad42073a`
- scope: removes the implicit `[u64; 4]` / `RefInto` conversion bounds from `Scalar`, adds explicit `from_limbs` / `to_limbs` helpers, updates scalar call sites, and removes the now-unused `ref_into` helper

Fresh local verification was run in a disposable Rust container because this host does not have a native Rust toolchain:

- container deps: `clang`, `lld`, `pkg-config`, `libssl-dev`
- `rustc --version` → `rustc 1.91.1 (ed61e7d7e 2025-11-07)`
- `cargo --version` → `cargo 1.91.1 (ea2d97820 2025-10-10)`
- `cargo fmt --all -- --check` → passed
- `cargo check -p proof-of-sql --lib` → passed (`Finished dev profile [unoptimized + debuginfo] target(s) in 1m 14s`)

The PR still includes `/claim #228` and looks ready for maintainer review if the current scope matches what you want for that bounty. If there is a specific extra test target you want around decimal comparison / scalar conversion call sites, I can add that before merge.

JayWhite2357:

PR #1445, opened 2026-05-12
test: cover fold utility edge cases

TheWaste11202:
/claim #560

# Rationale for this change

Adds a small, non-overlapping test-only coverage slice for the fold utility helpers.

- Cover `fold_columns` when it receives no columns and should leave an existing result slice unchanged.
- Cover `fold_vals` for the single-value case where the result should be that value regardless of beta.

# Are these changes tested?

Yes.

- `cargo test -p proof-of-sql sql::proof_plans::fold_util --no-default-features --features=test`
- `cargo fmt --check`
- `git diff --check`

Notes:
- `scripts/check_commits.sh` could not be run directly on this Windows environment because `bash.exe` requires a WSL distribution that is not installed. I manually checked the commit message against the script regex; `test: cover fold utility edge cases` matches.
- Before submitting, I checked current open PR file lists and did not find another open PR touching `crates/proof-of-sql/src/sql/proof_plans/fold_util.rs` or `fold_util_test.rs`.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_plans/fold_util_test.rs
+++ b/crates/proof-of-sql/src/sql/proof_plans/fold_util_test.rs
@@ -106,6 +106,24 @@ fn we_can_fold_empty_columns() {
 }
 
 #[test]
+fn fold_columns_leaves_result_unchanged_without_columns() {
+    let alloc = Bump::new();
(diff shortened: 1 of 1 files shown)


JayWhite2357:

PR #1450, opened 2026-05-12
test: cover Dory prover state validation

TheWaste11202:
/claim #560

# Rationale for this change

Adds a small, non-overlapping test-only coverage slice for Dory prover state validation.

- Cover `ProverState::new` rejecting a `v1` witness vector whose length does not match `2^nu`.
- Cover `ProverState::new` rejecting a `v2` witness vector whose length does not match `2^nu`.

# Are these changes tested?

Yes.

- `cargo test -p proof-of-sql proof_primitive::dory::state_test --no-default-features --features=test`
- `cargo fmt --check`
- `git diff --check`

Notes:
- `scripts/check_commits.sh` could not be run directly on this Windows environment because `bash.exe` requires a WSL distribution that is not installed. I manually checked the commit message against the script regex; `test: cover Dory prover state validation` matches.
- Before submitting, I checked current open PR file lists and did not find another open PR touching `crates/proof-of-sql/src/proof_primitive/dory/state.rs` or `state_test.rs`.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/state_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/state_test.rs
@@ -2,6 +2,28 @@ use super::{rand_G_vecs, test_rng, ProverState, PublicParameters};
 use ark_ec::pairing::Pairing;
 
 #[test]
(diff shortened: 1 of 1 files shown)


JayWhite2357:

PR #1473, opened 2026-05-12
test: cover literal value scalar conversions

TheWaste11202:
## Summary
- cover primitive literal conversions into `TestScalar`
- cover timestamp, varchar, varbinary, decimal, and limb-backed scalar conversions

## Verification
- `cargo test -p proof-of-sql base::database::literal_value::tests --no-default-features --features=test`
- `cargo fmt --check`
- `git diff --check`

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/base/database/literal_value.rs
+++ b/crates/proof-of-sql/src/base/database/literal_value.rs
@@ -100,6 +100,7 @@ mod tests {
         scalar::{test_scalar::TestScalar, ScalarExt},
         try_standard_binary_serialization,
     };
+    use proptest::prelude::*;
 
     /// This allows us to reuse code within solidity more safely
     #[test]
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a focused `/claim #560` test-only change in `crates/proof-of-sql/src/base/database/literal_value.rs` (`+98/-0`). It adds direct `LiteralValue::to_scalar::<TestScalar>()` coverage for Boolean, integer widths, TimestampTZ, VarChar hashing, VarBinary hashing, Decimal75, and limb-backed Scalar values, then backs that with proptests for integer, bool/timestamp, and byte literal conversion paths.

The main review point I would confirm is whether the hash-backed VarChar/VarBinary expectations should also pin cross-implementation stability, not only `TestScalar` equality. The current tests are useful for the local conversion contract, but an explicit serialization/hash fixture would make future changes to hashed literal behavior easier to spot.

I did not run the Rust suite locally in this pass. Visible Orca checks are green; `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required` on head `d38c60c464e89c58c62f4230c29e5bd11d2fee8a`, and the combined commit status API returned `state=pending` with no statuses.

JayWhite2357:

PR #1475, opened 2026-05-12
test: cover timestamp error display

TheWaste11202:
## Summary
- cover display text for timestamp parsing errors
- cover `PoSQLTimestampError` conversion into `String`

## Verification
- `cargo test -p proof-of-sql base::posql_time::error::tests --no-default-features --features=test`
- `cargo fmt --check`
- `git diff --check`

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/base/posql_time/error.rs
+++ b/crates/proof-of-sql/src/base/posql_time/error.rs
@@ -38,3 +38,67 @@ impl From<PoSQLTimestampError> for String {
         error.to_string()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn timestamp_errors_preserve_payloads() {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only `/claim #560` change in `crates/proof-of-sql/src/base/posql_time/error.rs` (`+64/-0`). It adds coverage for `PoSQLTimestampError` payload retention, display output for `InvalidTimezoneOffset`, `InvalidTimeUnit`, and `UnsupportedPrecision`, plus the `PoSQLTimestampError -> String` conversion path.

The main review point I would confirm is whether the exact Display strings should be treated as stable user-facing API. The tests pin `Invalid time unit` while the variant stores an `error: "hours"` payload, so maintainers may want to decide whether that payload should appear in the message or intentionally remain hidden.

JayWhite2357:

PR #1464, opened 2026-05-12
test: cover scalar conversion error

himanalot:
/claim #560

Adds focused test coverage for scalar conversion error display behavior:
- verifies `ScalarConversionError::Overflow` includes its underlying reason in the public display output

Validation:
- cargo test -p proof-of-sql base::scalar::error --no-default-features --features=test
- cargo fmt --check
- git diff --check
- source scripts/check_commits.sh

AI-assisted with OpenAI Codex; I reviewed the diff and verification before submitting.

PR #1478, opened 2026-05-12
test: cover scalar conversion error display

TheWaste11202:
## Summary
- cover display text for scalar conversion overflow errors

## Verification
- `cargo test -p proof-of-sql base::scalar::error::tests --no-default-features --features=test`
- `cargo fmt --check`
- `git diff --check`

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/base/scalar/error.rs
+++ b/crates/proof-of-sql/src/base/scalar/error.rs
@@ -11,3 +11,21 @@ pub enum ScalarConversionError {
         error: String,
     },
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn scalar_conversion_error_displays_overflow_message() {
+        let error = ScalarConversionError::Overflow {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only change in `crates/proof-of-sql/src/base/scalar/error.rs` (`+18/-0`). It constructs `ScalarConversionError::Overflow`, checks the stored inner error payload with `matches!`, and also pins the `Display` output for that variant. I do not see runtime or dependency risk in the diff itself.

The main review decision is stability/duplication: the exact overflow display string becomes part of the asserted behavior here, and this overlaps with the nearby scalar-conversion display coverage in #1464. Compared with that lane, this PR also checks the variant payload before asserting `to_string()`, so it may still be useful if maintainers want a compact regression test rather than only a display assertion.

I did not run the Rust suite locally in this pass. Visible Orca checks are green; the two GitHub Actions runs for this head currently show `action_required`, and the combined commit status API returned no statuses.

JayWhite2357:

PR #1479, opened 2026-05-12
test: cover RefInto blanket conversion

TheWaste11202:
## Summary
- add a focused test for the `RefInto` blanket implementation
- verify reference conversion works without consuming the source value

## Verification
- `cargo test -p proof-of-sql base::ref_into::tests --no-default-features --features=test`
- `cargo fmt --check`
- `git diff --check`

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/base/ref_into.rs
+++ b/crates/proof-of-sql/src/base/ref_into.rs
@@ -23,3 +23,26 @@ where
         self.into()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[derive(Debug, PartialEq, Eq)]
+    struct Wrapper(u64);
+
+    impl From<&Wrapper> for u64 {
+        fn from(value: &Wrapper) -> Self {
+            value.0
+        }
+    }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a very small `/claim #560` test-only change in `crates/proof-of-sql/src/base/ref_into.rs` (`+23/-0`). It adds a local `Wrapper` fixture, implements `From<&Wrapper> for u64`, and verifies the blanket `RefInto` implementation converts from a reference without consuming or mutating the original value.

The main review point is scope: this pins the trait boundary for a simple owned fixture, which is useful because `ref_into()` is intentionally a thin forwarding helper. If maintainers want stronger confidence, a follow-up could add a borrowed wrapper whose conversion returns a non-`Copy` target or a domain type used elsewhere in the crate, but this PR is clean as a minimal regression test for the blanket implementation.

I did not run the Rust suite locally in this pass. Visible Orca checks are green; `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required` on head `e3d4bfc3c4638459f5438d5539b2e11b48b419e3`, and the combined commit status API returned `state=pending` with no statuses.

JayWhite2357:

PR #1482, opened 2026-05-12
test: cover analyze error display

TheWaste11202:
## Summary
- add focused tests for `AnalyzeError` display messages
- cover conversion from `AnalyzeError` into `String`

## Verification
- `cargo test -p proof-of-sql sql::error::tests --no-default-features --features=test`
- `cargo fmt --check`
- `git diff --check`

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/sql/error.rs
+++ b/crates/proof-of-sql/src/sql/error.rs
@@ -71,3 +71,84 @@ impl From<IntermediateDecimalError> for AnalyzeError {
 
 /// Result type for analyze errors
 pub type AnalyzeResult<T> = Result<T, AnalyzeError>;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only change in `crates/proof-of-sql/src/sql/error.rs` (`+81/-0`). It pins several `AnalyzeError` display strings, the `AnalyzeError -> String` conversion path, and nested wrapper behavior for `IntermediateDecimalError::OutOfRange` plus `PlaceholderError::InvalidPlaceholderIndex`.

The main review point I would confirm is whether the exact user-facing wording for `InvalidDataType`, `DataTypeMismatch`, `DifferentColumnLength`, and `NotEnoughInputPlans` should be treated as stable API surface. The wrapper tests are useful for preserving error provenance, but the broad display-string assertions will make future copy changes test-visible, so this is a good place for maintainer intent to be explicit.

JayWhite2357:

PR #1489, opened 2026-05-12
test: cover zigzag decoding

TheWaste11202:
## Summary
- add ZigZag decoding tests for even encoded values
- add ZigZag decoding tests for odd encoded negative values

## Verification
- `cargo test -p proof-of-sql base::encode::zigzag_test --no-default-features --features=test`
- `cargo fmt --check`
- `git diff --check`

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/base/encode/zigzag_test.rs
+++ b/crates/proof-of-sql/src/base/encode/zigzag_test.rs
@@ -2,6 +2,7 @@ use crate::base::{
     encode::{ZigZag, U256},
     scalar::test_scalar::TestScalar,
 };
+use proptest::prelude::*;
 
 #[test]
 fn small_scalars_are_encoded_as_positive_zigzag_values() {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow `/claim #560` test-only change in `crates/proof-of-sql/src/base/encode/zigzag_test.rs` (`+51/-0`). It adds explicit decode coverage for even encoded values mapping to positive scalars, odd encoded values mapping to negative scalars, and proptests for positive `u64`, signed `i64`, and arbitrary `u64` encoded values round-tripping through `ZigZag`.

The main review point I would confirm is the final property `decoded.zigzag() == encoded` for arbitrary `u64` encoded values. It is a useful invariant for the low-word path, but it intentionally does not exercise non-zero high words in `U256::from_words`; if maintainers want broader codec confidence, a follow-up high-word/boundary case would cover the 256-bit surface more directly.

I did not run the Rust suite locally in this pass. Visible Orca checks are green. I did not see public GitHub Actions runs for head `779895ad99186f14e43734ce86b3638ba1aca725`; the combined commit status API returned `state=pending` with no statuses.

JayWhite2357:

PR #1503, opened 2026-05-13
Make scalar limb conversions explicit

### LINKED ISSUE #228: Make opaque `Scalar` conversions explicit rather than using `From`/`Into` trait.

szyunlin:
This removes the implicit `From<[u64; 4]>`, `Into<[u64; 4]>`, and `RefInto<[u64; 4]>` bounds from `Scalar` and replaces generic limb conversions with explicit `from_limbs` and `to_limbs` methods.

Validation:
- `cargo fmt --check`
- `cargo check -p proof-of-sql`

Closes #228


szyunlin:
Review

MyTH-zyxeon:
Maintainer-facing review assist for issue `#228`:

- Current visible state: `OPEN` / `MERGEABLE` / `mergeStateStatus=BLOCKED` / `reviewDecision=REVIEW_REQUIRED` at head `c8c54ef27e74869b6ca30d1b3e933ef930c11f4c`.
- Existing conversation only has the author's short `Review` comment; I do not see review threads or prior maintainer-facing deconfliction here.
- Visible status rollup is green for the three Orca checks. The combined status API returns only those Orca checks, and I do not see public GitHub Actions runs for this exact head SHA, so this still needs maintainer-side CI approval/rerun evidence before merge.
- Scope is code, not docs-only: this removes the implicit `[u64; 4]` `From` / `Into` / `RefInto` requirements from `Scalar`, adds explicit `from_limbs` / `to_limbs` methods, removes the `ref_into` module, and updates 16 files (`+83/-90`) across scalar, commitment, transcript, i256/U256, literal, scaling-cast, and range-check call sites.

Pre-merge checks I would focus on:

1. Confirm this is intended as the limb-conversion checkbox of `#228` only; string hashing and blanket `VarInt` are separate slices in other PRs.
2. Run `cargo fmt --all -- --check` plus `cargo check -p proof-of-sql` once fork CI is approved, because the PR body reports those locally but there is no public Actions evidence for this head.
3. Compare with newer limb-slice candidates such as `#2243` / `#1732` and pick one canonical lander, so the bounty routing does not split across equivalent `Scalar::from_limbs` / `Scalar::to_limbs` implementations.

No requested code change from me; this is review/deconfliction context for maintainer validation and bounty routing.

JayWhite2357:

PR #1510, opened 2026-05-13
test: cover powers of tau argument errors

sypham98-prog:
## Summary
- Return parse errors from the powers-of-tau binary creation helper instead of exiting inside the helper
- Keep the CLI behavior the same by printing the error and exiting from `main`
- Add coverage for the helper parse-error path

## Verification
- cargo fmt --check
- cargo test -p powers-of-tau-util --bin powers-of-tau-util
- cargo llvm-cov --no-report --all-features --workspace --exclude proof-of-sql-benches --ignore-filename-regex evm_tests
- cargo llvm-cov report --summary-only --ignore-filename-regex '[/\\]evm_tests\\.rs$'

Part of #560.

/claim #560

### DIFF
--- a/crates/powers-of-tau-util/src/main.rs
+++ b/crates/powers-of-tau-util/src/main.rs
@@ -82,20 +82,16 @@ fn write_commitment_key_to_binary(binary_path: &str, setup: &CommitmentKey<E>) {
     }
 }
 
-fn create_binary_file(args: &[String]) {
-    let (ptau_path, binary_path, n) = match parse_args(args) {
-        Ok(result) => result,
-        Err(err) => {
-            eprintln!("Error: {err}");
-            std::process::exit(1);
-        }
-    };
+fn create_binary_file(args: &[String]) -> Result<(), String> {
+    let (ptau_path, binary_path, n) = parse_args(args)?;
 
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review assist for #560: I checked this as a narrow coverage PR. The diff is contained to `crates/powers-of-tau-util/src/main.rs` (`+22/-10`) and makes `create_binary_file` return `Result<(), String>` so the parse-error branch can be tested directly while keeping the CLI behavior in `main` the same: print the error and exit non-zero. It also updates the existing round-trip test to unwrap the new result and adds coverage for the missing-argument usage error. Visible checks are green (Orca IaC, Secrets, and Vulnerabilities), with the PR still `BLOCKED` / `REVIEW_REQUIRED` and no existing comments, reviews, or review threads found before this note.

Claim hygiene / overlap note: this is adjacent to other powers-of-tau / test-runtime work such as #1933, but #1510 is differentiated by the argument-parse error path and the small helper refactor needed to test it without process exit. Maintainers should still de-dupe the #560 Codecov delta if another powers-of-tau PR lands first. No requested changes from me.

JayWhite2357:

PR #1514, opened 2026-05-13
test: cover add const edge cases

rafael81:
/claim #560

## Summary
- add a no-op empty-slice regression test for `add_const`
- cover the generic `S: Into<T>` path with an `i32` constant added into an `i64` slice

## Validation
- `cargo fmt --check`
- `cargo test -p proof-of-sql add_const --no-default-features --features=test`
- `git diff --check`

This PR is intentionally small and test-only. It was AI-assisted, then manually reviewed and validated before submission.


JayWhite2357:

PR #1555, opened 2026-05-13
test: cover ark serde macros

Jiangzhewen:
## Summary
- add focused tests for the checked and unchecked ark canonical serialization serde macros
- cover serde JSON roundtrips for both macro variants
- assert checked deserialization rejects invalid field-element bytes

/claim #560

## Validation
- cargo test -p proof-of-sql ark_serde --no-default-features --features=test
- cargo fmt --check
- git diff --check
- source scripts/check_commits.sh

### DIFF
--- a/crates/proof-of-sql/src/base/serialize.rs
+++ b/crates/proof-of-sql/src/base/serialize.rs
@@ -51,3 +51,45 @@ macro_rules! impl_serde_for_ark_serde_unchecked {
 
 pub(crate) use impl_serde_for_ark_serde_checked;
 pub(crate) use impl_serde_for_ark_serde_unchecked;
+
+#[cfg(test)]
+mod tests {
+    use ark_bn254::Fr;
+    use ark_ff::PrimeField;
+    use ark_serialize::{CanonicalDeserialize, CanonicalSerialize};
+
+    #[derive(CanonicalSerialize, CanonicalDeserialize, Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

I checked this as a compact test-only candidate. The change is limited to `crates/proof-of-sql/src/base/serialize.rs` and adds a `#[cfg(test)]` module that exercises both `impl_serde_for_ark_serde_checked!` and `impl_serde_for_ark_serde_unchecked!` through `serde_json` round trips.

- The checked path has a negative case: `we_reject_invalid_checked_ark_serde_bytes` builds an invalid `Fr` byte vector sized from `Fr::MODULUS_BIT_SIZE.div_ceil(8)` and asserts deserialization fails.
- The unchecked path is covered separately, so the two macro behaviors are not collapsed into one smoke test.
- Scope is narrow: one file, +42/-0, no production behavior changes outside the test module.

Deconfliction notes: before posting this, the PR had zero top-level comments and zero reviews. Related open search for `ark serde macros` also surfaced #2245, but the stricter search for `checked and unchecked ark canonical serialization` only matched this PR. Current PR state is `OPEN` / `BLOCKED` / `REVIEW_REQUIRED` at head `445f6fd69803098b2b40494eb047ffd5ac4cad92`; Orca checks are green. I did not run local cargo because the `/tmp/sxt-proof-of-sql` checkout is not present in this environment.

JayWhite2357:

PR #1557, opened 2026-05-13
test: cover degree four interpolation

Jiangzhewen:
## Summary
- add a focused degree-four coefficient reconstruction case for interpolate_evaluations_to_reverse_coefficients
- derive evaluations from explicit leading-coefficient-first polynomial coefficients and assert the original coefficients are recovered

/claim #560

## Validation
- cargo test -p proof-of-sql interpolate --no-default-features --features=test
- cargo fmt --check
- git diff --check
- source scripts/check_commits.sh

### DIFF
--- a/crates/proof-of-sql/src/base/polynomial/interpolate_test.rs
+++ b/crates/proof-of-sql/src/base/polynomial/interpolate_test.rs
@@ -161,3 +161,21 @@ fn we_can_interpolate_evaluations_to_reverse_coefficients_with_degree_3_degenera
         vec![S::from(0), S::from(0), S::from(2), S::from(1)]
     );
 }
+
+#[test]
+fn we_can_interpolate_evaluations_to_reverse_coefficients_with_degree_4() {
+    let coefficients = [S::from(3), S::from(-2), S::from(5), S::from(0), S::from(7)];
+    let evaluations = (0..coefficients.len())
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist for `#560`:

This is a small, focused coverage slice for `interpolate_evaluations_to_reverse_coefficients`: it builds degree-four evaluations from explicit leading-coefficient-first coefficients and checks the recovered coefficients exactly.

Suggested maintainer checks:

1. Keep this distinct from the nearby interpolation PRs. `#2159` appears to target sparse degree-five coverage and `#2386` broader coefficient round trips, while this PR is the only visible open entry explicitly named around degree-four interpolation.
2. The assertion compares the returned coefficients directly to the original fixed array, so it is a clear regression guard for ordering and reconstruction. That exactness is worth preserving if the patch is rebased.
3. The validation path in the body is targeted: `cargo test -p proof-of-sql interpolate --no-default-features --features=test`. That is the relevant check to rerun before merge.

Visible state at the checked head `b95ee334061bda9bffaba14a104448472dfae8fe`: open, mergeable, `BLOCKED` / `REVIEW_REQUIRED`, Orca checks green, with no prior top-level comments or reviews visible before this note.

JayWhite2357:

PR #1578, opened 2026-05-13
test: cover TableExec accessors

realkoreanbeef:
## Summary

- add non-blitzar unit coverage for `TableExec::table_ref()` and `TableExec::schema()`
- assert `TableExec` exposes column result fields, table references, and typed column references consistently
- cover the empty-schema case so source plans that project no columns still retain their table reference

/claim #560

## Verification

```text
cargo fmt --all
cargo test -p proof-of-sql table_exec_accessor --no-default-features --features test
cargo check -p proof-of-sql --no-default-features --features test
git diff --check
```


JayWhite2357:

PR #1586, opened 2026-05-14
perf: avoid duplicate verifier setup pairings

Jiangzhewen:
/claim #557

## Summary
- avoid recomputing the Dory verifier setup left-prefix pairing for every level
- build each `chi[k]` from the previous prefix plus the right-half pairing instead of recomputing the full prefix
- preserve the existing verifier setup values while reducing setup work and removing the now-unused `MultiUnzip` import

## Verification
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p proof-of-sql --no-default-features --features 'std test' proof_primitive::dory::setup_test`
- `cargo test -p proof-of-sql --no-run --no-default-features --features="std"`

Note: `cargo clippy -- -D warnings` does not run on this local macOS arm64 environment because the default feature set enables x86_64-only `halo2curves/asm`; CI should cover the Linux default-feature clippy job.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/setup.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/setup.rs
@@ -3,7 +3,6 @@ use crate::{base::impl_serde_for_ark_serde_unchecked, utils::log};
 use alloc::vec::Vec;
 use ark_ec::pairing::{Pairing, PairingOutput};
 use ark_serialize::{CanonicalDeserialize, CanonicalSerialize, Compress, Validate};
-use itertools::MultiUnzip;
 use num_traits::One;
 #[cfg(feature = "std")]
 use std::{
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist note for `/claim #557`: I took a metadata-level pass over this verifier-setup performance branch so maintainers can triage it faster.

Current PR state I see:
- Head: `35eb495ccd040146ffef81dfff3a10bfd52a6ca5`
- State: `OPEN`, `MERGEABLE`, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`
- Scope: 1 file, `+29/-29`
- Prior discussion: no top-level PR comments and no latest reviews at the time of this pass
- Checks: Orca IaC/Secrets/Vulnerabilities are green; `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required`; combined status API returned `statuses=[]`

What this branch appears to cover for #557:
- Replaces the previous tuple-map/`MultiUnzip` construction in `VerifierSetup::new` with explicit vectors and a single forward pass
- Reuses the prior prefix pairing as `Delta_1L_2L` / `chi[k - 1]` instead of recomputing the full left prefix each level
- Builds `chi[k]` from the previous prefix plus the right-half pairing while preserving the existing verifier setup shape

Pre-merge focus I would suggest: confirm the `PairingOutput` accumulation semantics exactly match the previous `Pairing::multi_pairing(&Gamma_1_nu[..1 << k], &Gamma_2_nu[..1 << k])` result for every `k`, especially around the `left_chi + right_chi` step, and make sure the targeted `proof_primitive::dory::setup_test` covers equality of all derived setup vectors. I did not run local cargo commands in this pass; this note is based on PR metadata, changed-file scope, visible patch, declared validation, and visible check state.

JayWhite2357:

PR #1592, opened 2026-05-14
test: cover literal expression evaluation

realkoreanbeef:
## Summary
- add focused coverage for `LiteralExpr::new`, `value`, and `data_type`
- cover first/final round prover evaluation producing constant columns with table row length
- cover verifier evaluation multiplying the literal scalar by `chi_eval`
- assert literal expressions report no column references

/claim #560

## Verification
- `cargo test -p proof-of-sql --no-default-features --features test sql::proof_exprs::literal_expr::tests --lib`
- `cargo fmt --check -p proof-of-sql`
- `git diff --check`

Note: default-feature test build was not used on this aarch64 runner because upstream `blitzar-sys`/`halo2curves asm` default perf dependencies are x86_64-only here.


JayWhite2357:

PR #1597, opened 2026-05-14
test: cover placeholder expression evaluation

realkoreanbeef:
## Summary
- add PlaceholderExpr coverage for converting a positional parameter into a repeated literal column during first-round evaluation
- add verifier coverage that confirms placeholder scalar values are multiplied by the supplied chi evaluation

/claim #560

## Verification
- `cargo fmt --check`
- `cargo test -p proof-of-sql --no-default-features --features test placeholder_expr --lib`
- `git diff --check`

Note: checks run with no default features on this aarch64 environment; default-feature tests are not used here because this repo's blitzar/halo2 asm path has been unavailable on aarch64 in prior runs.


JayWhite2357:

PR #1599, opened 2026-05-14
test: shrink Dory commitment test setup

realkoreanbeef:
## Summary
- reduce the Dory commitment unit-test public parameter size from `max_nu = 5` to a named `DORY_COMMITMENT_TEST_MAX_NU = 4`
- keep the same commitment inputs and expected pairing assertions
- this still covers the largest expected index used by these tests while avoiding over-provisioned setup generation in 20 tests

/claim #557

## Verification
- `cargo fmt`
- `cargo test -p proof-of-sql --no-default-features --features test proof_primitive::dory::dory_compute_commitments_test -- --nocapture` (20 passed; finished in 15.89s)
- `cargo fmt --check`
- `git diff --check`

Note: default-feature focused test was attempted first but is unavailable on this aarch64 runner because `halo2curves` asm and `blitzar-sys` require x86_64, matching the known local limitation from prior SxT work.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dory_compute_commitments_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dory_compute_commitments_test.rs
@@ -12,9 +12,12 @@ use ark_ec::pairing::Pairing;
 use ark_std::test_rng;
 use num_traits::Zero;
 
+const DORY_COMMITMENT_TEST_MAX_NU: usize = 4;
+
 #[test]
 fn we_can_compute_a_dory_commitment_with_varbinary_values() {
-    let public_parameters = PublicParameters::test_rand(5, &mut test_rng());
+    let public_parameters =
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

- Scope checked: one test file only, `crates/proof-of-sql/src/proof_primitive/dory/dory_compute_commitments_test.rs` (`+42/-20`).
- The change names the reduced setup bound as `DORY_COMMITMENT_TEST_MAX_NU = 4` instead of leaving the smaller parameter implicit, while keeping the same commitment inputs and expected pairing assertions.
- PR body verification is useful for review: focused no-default-features test reports `20 passed`, plus `cargo check`, `cargo fmt --check`, and `git diff --check`.
- GitHub Actions runs show `action_required`, so I am treating that as workflow approval/visibility rather than a code failure unless maintainers say otherwise.

This looks like a narrow test-runtime/setup reduction for the #557 lane, with an explicit local-run caveat for the known aarch64 default-feature limitation.

JayWhite2357:

PR #1604, opened 2026-05-14
test: trim Dory commitment proof length matrix

realkoreanbeef:
## Summary
- trim the Dory commitment evaluation proof random length matrix from every length on every setup to representative cases
- keep coverage for large non-power-of-two, power-of-two, smaller non-power-of-two, and small boundary lengths
- avoid rebuilding/rerunning redundant Dory setups across the full length list

/claim #557

## Verification
- `cargo fmt --check`
- `cargo test -p proof-of-sql --no-default-features --features test proof_primitive::dory::dory_commitment_evaluation_proof_test -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dory_commitment_evaluation_proof_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dory_commitment_evaluation_proof_test.rs
@@ -52,20 +52,28 @@ fn test_random_ipa_with_length_1() {
 
 #[test]
 fn test_random_ipa_with_various_lengths() {
-    let lengths = [128, 100, 64, 50, 32, 20, 16, 10, 8, 5, 4, 3, 2];
-    let setup_setup = [(4, 4), (4, 3), (6, 2)];
-    for setup_p in setup_setup {
-        let public_parameters = PublicParameters::test_rand(setup_p.0, &mut test_rng());
+    let test_cases = [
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist for `/claim #557`:

I checked this as a narrow test-performance slice. Current public state is `OPEN` / `MERGEABLE` but still `mergeStateStatus=BLOCKED` / `reviewDecision=REVIEW_REQUIRED`; Orca checks are green, while `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required`, so the next blocker looks like maintainer workflow approval rather than a visible test failure.

The diff is intentionally small: it replaces the 13-length x 3-setup Dory commitment evaluation proof matrix with 6 representative cases and leaves production code untouched. The representative set still covers large non-power-of-two, power-of-two, smaller non-power-of-two, small power-of-two, and tiny boundary lengths.

One pre-merge check I would call out: the old matrix included `128`, while the new set keeps `100` for the largest setup and `64` for a power-of-two case. If maintainers specifically wanted the largest power-of-two boundary retained, adding or swapping a single `(4, 4, 128)` case would preserve that signal; otherwise this looks like a reasonable runtime-reduction tradeoff for #557 once the fork workflows are approved and the claimed `cargo fmt`, targeted test, `cargo check`, and `git diff --check` commands are visible from CI.

JayWhite2357:

PR #1609, opened 2026-05-14
Materialize nullable Arrow columns for proof tests

mt211211:
/claim #183

## Summary
This is a focused PoC for #183. It adds a nullable Arrow RecordBatch conversion path that materializes each nullable input column as:

- the existing physical value column, with null slots canonicalized to the type default
- a generated `<column>__presence` boolean column

That gives proof plans an explicit presence bitmap they can combine with ordinary expressions while the broader nullable SQL design is still being worked through. The existing `TryFrom<RecordBatch>` behavior is left unchanged.

## Why this slice
The issue design discussion describes nullable column `c` as a value column plus a presence column. This PR lands that representation at the Arrow boundary and keeps it isolated from the larger end-to-end nullable SQL work, so maintainers can review or reuse it independently.

The added proof test starts from an Arrow batch with real null slots, materializes `score` and `score__presence`, then proves a filtered arithmetic query that only accepts present rows.

## Tests
- `cargo test -p proof-of-sql --no-default-features --features "arrow test" owned_and_arrow_conversions_test -- --nocapture`

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/owned_and_arrow_conversions.rs
+++ b/crates/proof-of-sql/src/base/arrow/owned_and_arrow_conversions.rs
@@ -34,6 +34,9 @@ use arrow::{
 use snafu::Snafu;
 use sqlparser::ast::Ident;
 
+/// Suffix used for the generated validity bitmap column when nullable Arrow arrays are materialized.
+pub const NULL_PRESENCE_COLUMN_SUFFIX: &str = "__presence";
+
 #[derive(Snafu, Debug)]
 #[non_exhaustive]
 /// Errors caused by conversions between Arrow and owned types.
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:
Maintainer-facing review note for `/claim #183` after a current refresh:

- State I see: `OPEN` / `MERGEABLE`, Orca checks green, but `mergeStateStatus=BLOCKED` and `reviewDecision=REVIEW_REQUIRED`; no prior visible top-level comments.
- This is a focused nullable-column slice: `owned_table_from_record_batch_with_nulls` leaves the existing no-null `TryFrom<RecordBatch>` path unchanged, materializes nullable Arrow arrays as value column + `<name>__presence`, and includes a proof test that gates arithmetic on the generated presence column.
- Acceptance points worth checking before merge: confirm `__presence` is the intended public naming convention and collision policy; confirm unsupported nullable Arrow types should stay `UnsupportedType`; run the full maintainer CI/fork workflow because the PR body only reports focused Arrow tests and public checks currently expose Orca only.
- Payout hygiene: this PR carries `/claim #183`. If this slice is accepted independently from the broader nullable foundation / #1955 work, it should stay on the #183 review path; if it is superseded, marking that explicitly would avoid duplicate review and claim ambiguity.

PR #1611, opened 2026-05-14
test: cover ProjectionExec accessors

realkoreanbeef:
## Summary
- add a focused unit test for `ProjectionExec::input` and `ProjectionExec::aliased_results`
- keep the test independent of the default blitzar feature path so it can run on non-x86_64 environments

/claim #560

## Verification
- `cargo fmt --check`
- `cargo test -p proof-of-sql --no-default-features --features=test we_can_access_projection_input_and_aliased_results`
- `cargo check -p proof-of-sql --no-default-features --features=test`
- `git diff --check`

Note: default-feature test runs are not available on this aarch64 runner because `blitzar-sys`/`halo2curves` require x86_64/asm.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_plans/projection_exec.rs
+++ b/crates/proof-of-sql/src/sql/proof_plans/projection_exec.rs
@@ -52,6 +52,40 @@ impl ProjectionExec {
     }
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        base::database::ColumnType,
+        sql::{
+            proof_exprs::{ColumnExpr, DynProofExpr},
+            proof_plans::TableExec,
+        },
+    };
+
+    #[test]
+    fn we_can_access_projection_input_and_aliased_results() {
+        let table_ref = TableRef::new("sxt", "sxt_tab");
+        let input = DynProofPlan::Table(TableExec::new(
+            table_ref.clone(),
+            vec![ColumnField::new("a".into(), ColumnType::BigInt)],
+        ));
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer-facing deconfliction note for #560 routing:

I checked this PR as a narrow #560 coverage slice. Current state is `OPEN`, `MERGEABLE`, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED` at head `02320d14bd2792ef3921fcf78e8738a2842089a4`; there were no existing issue comments, reviews, or review-thread comments when I checked. The visible Orca checks are green, and the combined commit status contexts are empty.

The patch is test-only in `crates/proof-of-sql/src/sql/proof_plans/projection_exec.rs` (`+34/-0`) and covers the two accessor paths `ProjectionExec::input()` and `ProjectionExec::aliased_results()` using a `TableExec` input plus a single aliased `ColumnExpr`. I did not rerun local cargo in this pass because the local `/tmp/sxt-proof-of-sql` checkout was unavailable; this note is remote triage/deconfliction only.

For award/routing hygiene, compare this with #2424 before accepting either lane. #1611 is the earlier, narrower accessor-only claim; #2424 is a later broader ProjectionExec accessor claim that also covers `get_column_result_fields()` alias/type metadata and already has a maintainer-facing review-assist note. If both remain open, I would treat them as overlapping submissions for the same ProjectionExec accessor coverage area unless maintainers explicitly want the narrower #1611 slice separately.

JayWhite2357:

PR #1625, opened 2026-05-15
test: cover HyperKZG setup deserialization errors

digzrow-coder:
/claim #560

Contributes to #560

## Summary
- Add focused coverage for HyperKZG public setup deserialization failure paths.
- Assert truncated compressed setup bytes are rejected instead of silently producing a partial setup.
- Assert reader I/O errors are propagated through the reader-based setup deserializer.

This is a small, test-only #560 coverage slice in `public_setup.rs`. I checked current open PRs for HyperKZG/public setup overlap; #1624 covers HyperKZG column commitments, while this covers setup deserialization error paths.

## Demo
https://github.com/digzrow-coder/sxt-proof-of-sql/releases/download/pr-1625-demo-20260515/pr-1625-hyperkzg-setup-errors-demo.mp4

## Bounty / Coverage Accounting
- Bounty terms: $0.10 per newly covered production line, subject to maintainer/Algora accounting.
- Changed file: `crates/proof-of-sql/src/proof_primitive/hyperkzg/public_setup.rs`
- Patch size: 31 added test lines, 0 production-line changes.
- Covered behavior: truncated compressed setup slice rejection and reader error propagation in the HyperKZG public setup deserialization path.
- Exact rewarded covered-line count is left to the maintainer coverage baseline, but this PR is intentionally scoped to previously untested error paths rather than broad/noisy test churn.

## Verification
- `cargo +stable-x86_64-pc-windows-gnu test -p proof-of-sql --no-default-features --features=test proof_primitive::hyperkzg::public_setup::std_tests`: 6 passed, 0 failed
- Previous local verification: `cargo check -p proof-of-sql --no-default-features --features=test`: passed
- Previous local verification: `rustfmt --check crates/proof-of-sql/src/proof_primitive/hyperkzg/public_setup.rs`: passed
- Previous local verification: `git diff --check`: passed

Note: full `cargo fmt --check` on this Windows/GNU setup reports an unrelated import-order diff in `crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_helper.rs`, so I verified formatting on the touched file directly to avoid unrelated churn. Invoking the default MSVC toolchain on this machine fails before tests because `link.exe` is not installed; the GNU command above is the validation target used for this PR.

AI-assisted with OpenAI Codex; I reviewed the diff, proof asset, and validation output before posting.

PR #1624, opened 2026-05-14
test: cover HyperKZG column commitments

jamilahmadzai:
/claim #560

## Summary
- Add focused coverage for HyperKZG CPU commitment dispatch across the current `CommittableColumn` variants.
- Assert the computed commitments against explicit generator/scalar products with a non-zero setup offset.
- Keep this test-only and separate from the existing arithmetic/transcript coverage slice in #1297.

## Verification
- `cargo test -p proof-of-sql --no-default-features --features=test proof_primitive::hyperkzg::commitment`
- `cargo llvm-cov -p proof-of-sql --no-default-features --features=test --lib --lcov --output-path /tmp/hyperkzg_commitment_columns_after.lcov -- proof_primitive::hyperkzg::commitment`
- `cargo check -p proof-of-sql --no-default-features --features=test`
- `cargo fmt --check`
- `git diff --check`
- `source scripts/check_commits.sh`

Local focused LCOV moved existing production coverage in `commitment.rs` from 35/109 to 75/109 lines, for 40 pre-existing production lines newly hit in this slice. The repository coverage workflow remains authoritative for final accounting.

## Payout
Payment method: Algora bounty-platform payout to GitHub user @jamilahmadzai for issue #560 coverage-pool accounting.

PR #1626, opened 2026-05-15
test: cover HyperKZG offset rejection

digzrow-coder:
/claim #560

Closes #560

## Summary
- Add focused coverage for the HyperKZG verifier guard that rejects non-zero generator offsets.
- Build a valid small HyperKZG proof, then verify it with `generators_offset = 1` and assert the explicit `NovaError::InvalidPCS` result.
- Keep this test-only and separate from #1625, which covers public setup deserialization errors, and #1624, which covers HyperKZG column commitment dispatch.

## Verification
- `cargo test -p proof-of-sql --no-default-features --features="test,hyperkzg_proof" proof_primitive::hyperkzg::commitment_evaluation_proof::tests::we_reject_nonzero_generator_offsets_when_verifying_hyperkzg_proofs`: 1 passed, 0 failed
- `cargo check -p proof-of-sql --no-default-features --features="test,hyperkzg_proof"`: passed
- `rustfmt --check crates/proof-of-sql/src/proof_primitive/hyperkzg/commitment_evaluation_proof.rs`: passed
- `git diff --check`: passed

Note: full `cargo fmt --check` on this Windows/GNU setup reports an unrelated import-order diff in `crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_helper.rs`, so I verified formatting on the touched file directly to avoid unrelated churn.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/hyperkzg/commitment_evaluation_proof.rs
+++ b/crates/proof-of-sql/src/proof_primitive/hyperkzg/commitment_evaluation_proof.rs
@@ -160,6 +160,11 @@ mod tests {
             test_commitment_evaluation_proof_with_length_1,
             test_random_commitment_evaluation_proof, test_simple_commitment_evaluation_proof,
         },
+        base::{
+            commitment::{Commitment, CommitmentEvaluationProof, CommittableColumn},
+            polynomial::compute_evaluation_vector,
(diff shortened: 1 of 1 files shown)


digzrow-coder:
Closing this second #560 slice to avoid duplicate review noise from the same contributor. Keeping #1625 open as the active tracked claim for the HyperKZG coverage work.

PR #1734, opened 2026-05-18
test: cover big decimal precision scaling

JackSpiece:
/claim #560

## Summary
- Added focused tests for `BigDecimalExt::try_into_bigint_with_precision_and_scale`
- Covered successful scale padding into the integer representation
- Covered the `LossyCast` path when scaled digits exceed the requested precision

## Verification
- cargo test -p proof-of-sql --no-default-features --features std base::math::big_decimal_ext
- cargo fmt --check

### DIFF
--- a/crates/proof-of-sql/src/base/math/big_decimal_ext.rs
+++ b/crates/proof-of-sql/src/base/math/big_decimal_ext.rs
@@ -89,4 +89,24 @@ mod tests {
             .unwrap_err();
         assert!(matches!(err, IntermediateDecimalError::ConversionFailure));
     }
+
+    #[test]
+    fn try_into_bigint_pads_scale_when_precision_allows() {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

- This PR is test-only in `crates/proof-of-sql/src/base/math/big_decimal_ext.rs`, adding focused coverage for `try_into_bigint_with_precision_and_scale`.
- The new assertions cover both the scale-padding path (`12.3` with precision 4 / scale 2 -> `1230`) and the precision rejection path for an over-wide scaled integer (`12345` with precision 4 / scale 0 -> `LossyCast`). That looks targeted at existing decimal conversion behavior rather than changing runtime semantics.
- Public status visibility currently shows the Orca checks green; `gh run list --commit 48852edcc57e...` returned no visible workflow runs and the combined status API returned `statuses=[]`, so I do not see a public code-level CI failure. The remaining gate appears to be review / workflow approval.

jackspiece:

PR #1736, opened 2026-05-18
test: cover owned column error messages

JackSpiece:
/claim #560

## Summary
- Added focused coverage for owned column error display messages
- Covered type cast, scalar conversion, unsupported operation, and column coercion errors
- Kept the change test-only

## Verification
- cargo test -p proof-of-sql --no-default-features --features std base::database::owned_column_error
- cargo fmt --check
- git diff --check

### DIFF
--- a/crates/proof-of-sql/src/base/database/owned_column_error.rs
+++ b/crates/proof-of-sql/src/base/database/owned_column_error.rs
@@ -40,3 +40,46 @@ pub(crate) enum ColumnCoercionError {
 
 /// Result type for operations related to `OwnedColumn`s.
 pub type OwnedColumnResult<T> = core::result::Result<T, OwnedColumnError>;
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

- This PR is test-only in `crates/proof-of-sql/src/base/database/owned_column_error.rs`, adding focused display-string coverage for `OwnedColumnError` and `ColumnCoercionError`.
- The assertions exercise the existing `Display` contract for type-cast, scalar-conversion, unsupported-operation, overflow, and invalid-coercion cases, so the change appears targeted at current diagnostics rather than runtime behavior.
- Public status visibility currently shows the Orca checks green; `gh run list --commit 867d1e1c73af...` returned no visible workflow runs and the combined status API returned `statuses=[]`, so I do not see a public code-level CI failure. The remaining gate appears to be review / workflow approval.

jackspiece:

PR #1738, opened 2026-05-18
test: cover dory public setup accessors

JackSpiece:
/claim #560

## Summary
- Added focused coverage for Dory prover and verifier public setup wrappers
- Covered constructor sigma storage and wrapped setup accessors
- Kept the change test-only

## Verification
- cargo test -p proof-of-sql --no-default-features --features std proof_primitive::dory::dory_public_setup::tests
- cargo fmt --check
- git diff --check

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dory_public_setup.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dory_public_setup.rs
@@ -57,3 +57,32 @@ impl<'a> DoryVerifierPublicSetup<'a> {
         self.verifier_setup
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{DoryProverPublicSetup, DoryVerifierPublicSetup};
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

- This PR is test-only in `crates/proof-of-sql/src/proof_primitive/dory/dory_public_setup.rs`, adding focused coverage for the public setup accessors.
- The tests build prover/verifier setups from `PublicParameters::test_rand(1, ...)`, then verify both the stored `sigma` value and that the accessor returns the same wrapped setup reference via `ptr::eq`. That makes the assertions targeted at the existing wrapper contract rather than changing Dory behavior.
- Public status visibility currently shows the Orca checks green; `gh run list --commit 23b9c3e3037a...` returned no visible workflow runs and the combined status API returned `statuses=[]`, so I do not see a public code-level CI failure. The remaining gate appears to be review / workflow approval.

jackspiece:

PR #1741, opened 2026-05-18
test: cover verifiable result tampering helpers

JackSpiece:
/claim #560

## Summary
- Added focused coverage for verifiable query result tampering helpers
- Covered zero-column results, empty-result default row insertion, and first-row mutations across owned column variants
- Kept the change test-only

## Verification
- cargo fmt --check
- DOCS_RS=1 cargo check --tests -p proof-of-sql --no-default-features --features std,blitzar
- git diff --check

## Note
The targeted helper module is gated behind the `blitzar` feature. On Android/Termux, `blitzar-sys` cannot link or run because its build script supports only Linux x86_64, so I used the docs build path to typecheck the feature-gated test build locally.


jackspiece:

PR #1742, opened 2026-05-18
test: cover limb serialization order

JackSpiece:
/claim #560

## Summary
- Added focused coverage for the limb serialization helpers
- Verified the reversed wire order for serialization and deserialization
- Added a round-trip check for boundary limb values

## Verification
- cargo test -p proof-of-sql --no-default-features --features std limbs_
- cargo fmt --check
- git diff --check

### DIFF
--- a/crates/proof-of-sql/src/base/standard_serializations/limbs.rs
+++ b/crates/proof-of-sql/src/base/standard_serializations/limbs.rs
@@ -13,3 +13,52 @@ pub(crate) fn deserialize_to_limbs<'de, D: Deserializer<'de>>(
     let limbs = <[u64; 4]>::deserialize(deserializer)?;
     Ok([limbs[3], limbs[2], limbs[1], limbs[0]])
 }
+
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

- This PR is test-only in `crates/proof-of-sql/src/base/standard_serializations/limbs.rs`, adding serde coverage around the existing limb-order reversal helpers.
- The direct serialization assertion checks `[1, 2, 3, 4] -> [4, 3, 2, 1]`, the deserialization assertion checks the inverse wire order, and the round-trip case includes `u64::MIN` / `u64::MAX` boundaries. That makes the coverage targeted at the current conversion contract rather than changing runtime behavior.
- Public status visibility currently shows the Orca checks green; `gh run list --commit 004deacf4d2a...` returned no visible workflow runs and the combined status API returned `statuses=[]`, so I do not see a public code-level CI failure. The remaining gate appears to be review / workflow approval.

jackspiece:

PR #1746, opened 2026-05-18
test: cover standard serialization helpers

partyplatter08-lab:
/claim #560

## Summary

- Adds direct coverage for the standard binary serialization contract, including fixed-width big-endian output and consumed-byte reporting during decode.
- Adds coverage for limb serialization adapters, verifying reverse-order serialization, reverse-order deserialization, and rejection of malformed limb arrays.

## Testing

- `cargo fmt --check`
- `cargo --config 'target.x86_64-unknown-linux-gnu.rustflags=[]' --config 'target.x86_64-unknown-linux-gnu.linker="cc"' test -p proof-of-sql standard_serializations --no-default-features --features test`
- `cargo --config 'target.x86_64-unknown-linux-gnu.rustflags=[]' --config 'target.x86_64-unknown-linux-gnu.linker="cc"' test -p proof-of-sql --lib --no-default-features --features test`

### DIFF
--- a/crates/proof-of-sql/src/base/standard_serializations/binary.rs
+++ b/crates/proof-of-sql/src/base/standard_serializations/binary.rs
@@ -51,4 +51,21 @@ mod tests {
         let reserialized = try_standard_binary_serialization(deserialized).unwrap();
         assert_eq!(serialized, reserialized);
     }
+
+    #[test]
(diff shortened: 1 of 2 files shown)


partyplatter08-lab:
Closing this to avoid adding more noise to an already very crowded coverage bounty thread. Thanks for the opportunity.

PR #1748, opened 2026-05-18
test: cover standard binary serialization

JackSpiece:
/claim #560

## Summary
- Added focused coverage for the standard binary serialization helpers
- Verified fixed-width integers serialize as big-endian bytes
- Covered consumed-length reporting when deserializing with trailing bytes

## Verification
- cargo test -p proof-of-sql --no-default-features --features std base::standard_serializations::binary::tests
- cargo fmt --check
- git diff --check

### DIFF
--- a/crates/proof-of-sql/src/base/standard_serializations/binary.rs
+++ b/crates/proof-of-sql/src/base/standard_serializations/binary.rs
@@ -51,4 +51,23 @@ mod tests {
         let reserialized = try_standard_binary_serialization(deserialized).unwrap();
         assert_eq!(serialized, reserialized);
     }
+
+    #[test]
+    fn serializes_fixed_width_integers_as_big_endian() {
+        let serialized = try_standard_binary_serialization(0x0102_0304_u32).unwrap();
+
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow `/claim #560` test-only PR in `crates/proof-of-sql/src/base/standard_serializations/binary.rs` (`+19/-0`). It adds two focused unit tests: one locks fixed-width integer serialization to big-endian bytes, and the other verifies deserialization reports the consumed length correctly when trailing bytes remain.

The main review point I would check is that the trailing-byte test is intentionally asserting parser boundary behavior rather than accepting/ignoring the suffix as a full-buffer validation path. The implementation calls `try_standard_binary_deserialization(&serialized)` and checks the returned `consumed == 2`, so the test is scoped to the helper's documented consumed-length contract and does not change production serialization code.

I did not run the Rust suite locally in this pass. GitHub currently shows the Orca security checks succeeded, with the PR still `OPEN` / `MERGEABLE` but gated by `REVIEW_REQUIRED` / `BLOCKED` rather than a visible code failure.

jackspiece:

PR #1400, opened 2026-05-12
test: cover owned table utility constructors

INDICUMA:
## Summary
- cover every owned table utility column constructor with direct assertions on generated identifiers and `OwnedColumn` values
- cover the `owned_table` convenience wrapper by checking resulting row/column counts and generated column types
- include timestamp, decimal, scalar, string, binary, boolean, unsigned, and signed integer helper paths

## Verification
- `cargo fmt --check`
- `cargo test -p proof-of-sql --no-default-features --features test owned_table_utility::tests`
- `cargo llvm-cov --no-default-features --features test --no-report -p proof-of-sql -- owned_table_utility::tests`
- `cargo llvm-cov report --summary-only | rg 'base/database/owned_table_utility.rs|TOTAL'`
  - `base/database/owned_table_utility.rs`: 100.00% line coverage under the targeted test run
- `git diff --check`

/claim #560

PR #1749, opened 2026-05-18
test: cover owned table utility constructors

JackSpiece:
/claim #560

## Summary
- Added direct coverage for owned table utility constructors across supported column variants
- Verified constructor output for numeric, scalar, string, binary, decimal, and timestamp columns
- Covered the mismatched column length panic path through `owned_table`

## Verification
- cargo test -p proof-of-sql --no-default-features --features std base::database::owned_table_utility::tests
- cargo fmt --check
- git diff --check

### DIFF
--- a/crates/proof-of-sql/src/base/database/owned_table_utility.rs
+++ b/crates/proof-of-sql/src/base/database/owned_table_utility.rs
@@ -320,3 +320,72 @@ pub fn timestamptz<S: Scalar>(
         OwnedColumn::TimestampTZ(time_unit, timezone, data.into_iter().collect()),
     )
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::base::{math::decimal::Precision, scalar::test_scalar::TestScalar};
+
+    #[test]
+    fn constructors_build_expected_owned_columns() {
+        let table = owned_table::<TestScalar>([
+            boolean("bools", [true, false]),
+            uint8("u8s", [1_u8, 2]),
+            tinyint("i8s", [-1_i8, 2]),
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

I checked this as a focused test-only candidate. The change stays in `crates/proof-of-sql/src/base/database/owned_table_utility.rs` and adds a `#[cfg(test)]` module around the owned-table helper constructors.

- `constructors_build_expected_owned_columns` builds one table covering boolean, uint8, signed integer widths, int128, scalar, varchar, varbinary, decimal75, and timestamptz helpers, then asserts both shape (`12` columns / `2` rows) and each resulting `OwnedColumn` payload.
- The decimal path checks `Precision::new(12).unwrap()` plus scale `2`, so it is not just a constructor smoke test.
- `owned_table_panics_for_mismatched_column_lengths` covers the `ColumnLengthMismatch` path through the public helper surface.
- Scope is narrow: one file, `+69/-0`, no production behavior changes outside the test module.

Deconfliction notes: before posting this, the PR had zero top-level comments and zero reviews. Related search does show nearby owned-table PRs (#1400 same title and #2032 helper coverage), so this is crowded, but this PR's review surface is still distinguishable by the broad constructor-variant assertions plus mismatch panic check. Current PR state is `OPEN` / `BLOCKED` / `REVIEW_REQUIRED` at head `2ba130454d9bd5d9d47ef53b384d3ef7f08fffb3`; Orca checks are green. I did not run local cargo because the `/tmp/sxt-proof-of-sql` checkout is not present in this environment.

jackspiece:

PR #1760, opened 2026-05-19
test: cover record batch conversion without blitzar

gshaowei6:
/claim #560

## Scope

Exercises the Arrow record batch conversion paths that are available without the `blitzar` feature.

- Runs the existing record batch commitment path under the `arrow` feature gate instead of tying it to `blitzar`.
- Covers `batch_to_columns` for Boolean, UInt8, and Utf8 arrays, including field-name preservation.
- Covers nullable Arrow values returning `RecordBatchToColumnsError`.
- Covers offset-aware table commitment creation from a record batch.
- Checks append-time type mismatch handling and verifies the existing commitment range is left unchanged.
- Covers duplicate field-name handling for both initial table commitment creation and record batch append.

The mixed-length panic branches remain unreachable through Arrow's safe `RecordBatch` constructor because Arrow validates equal column lengths before the commitment layer sees the batch.

## Validation

Windows:

```text
cargo fmt --all -- --check
passed

cargo test -p proof-of-sql --lib --no-default-features --features "arrow test" record_batch_conversion
7 passed

cargo test -p proof-of-sql --lib --no-default-features --features "arrow test"
1060 passed
```

WSL Ubuntu-26.04:

```text
cargo fmt --all -- --check
passed

cargo test -p proof-of-sql --lib --no-default-features --features "arrow test" record_batch_conversion --quiet
7 passed

cargo test -p proof-of-sql --lib --no-default-features --features "arrow test" --quiet
1060 passed
```

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/record_batch_conversion.rs
+++ b/crates/proof-of-sql/src/base/arrow/record_batch_conversion.rs
@@ -99,18 +99,160 @@ impl<C: Commitment> TableCommitment<C> {
     }
 }
 
-#[cfg(all(test, feature = "blitzar"))]
+#[cfg(all(test, feature = "arrow"))]
 mod tests {
     use super::*;
     use crate::base::{
+        arrow::arrow_array_to_column_conversion::ArrowArrayToColumnConversionError,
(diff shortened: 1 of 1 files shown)


JayWhite2357:

PR #1766, opened 2026-05-19
test: cover timestamp scale cast branches

gshaowei6:
/claim #560

## Scope

Covers the timestamp-specific branches in `scale_cast_binary_op`.

- Builds second- and millisecond-precision `TimestampTZ` column expressions.
- Verifies the left-hand timestamp branch wraps the lower-precision operand in a scaling cast.
- Verifies the right-hand timestamp branch does the same when the lower-precision operand is on the right.

The assertions compare against `DynProofExpr::try_new_scaling_cast` directly, so the test exercises the timestamp path rather than the decimal scaling helper.

## Validation

Windows:

```text
cargo fmt --all -- --check
passed

cargo test -p proof-of-sql --lib --no-default-features --features "test" scale_cast_binary_op
5 passed

cargo test -p proof-of-sql --lib --no-default-features --features "test"
962 passed
```

WSL Ubuntu-26.04:

```text
cargo fmt --all -- --check
passed

cargo test -p proof-of-sql --lib --no-default-features --features "test" scale_cast_binary_op --quiet
5 passed

cargo test -p proof-of-sql --lib --no-default-features --features "test" --quiet
962 passed
```

### DIFF
--- a/crates/proof-of-sql/src/sql/scale.rs
+++ b/crates/proof-of-sql/src/sql/scale.rs
@@ -73,7 +73,10 @@ pub fn scale_cast_binary_op(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::base::database::{ColumnRef, TableRef};
+    use crate::base::{
+        database::{ColumnRef, TableRef},
+        posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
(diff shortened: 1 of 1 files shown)


JayWhite2357:

PR #1772, opened 2026-05-19
Reject nullable RecordBatch schemas during Arrow conversions

er1c-cartman:
## Summary
- Reject nullable Arrow fields before converting `RecordBatch` values into Proof-of-SQL owned columns.
- Add explicit conversion errors for nullable Arrow schema fields.
- Add regression coverage for nullable schema fields, including the case where the actual values contain no nulls.

## Why
Proof-of-SQL owned columns are non-nullable today. Before this change, a nullable Arrow field could be accepted when the concrete array had no null values, silently dropping the schema-level nullability bit during conversion.

Refs #183.

## Validation
- `cargo fmt --check`
- `git diff --check`
- `scripts/check_commits.sh`
- `cargo test -p proof-of-sql --no-default-features --features="arrow test" --lib` (1055 passed)

## Bounty claim
/claim #183

This claims a partial, productive slice of the nullable column bounty: it adds an explicit guard for nullable Arrow RecordBatch inputs so unsupported nullable data cannot be silently converted into non-nullable columns, with tests covering the new rejection path.

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/owned_and_arrow_conversions.rs
+++ b/crates/proof-of-sql/src/base/arrow/owned_and_arrow_conversions.rs
@@ -20,7 +20,7 @@ use crate::base::{
     posql_time::{PoSQLTimeUnit, PoSQLTimeZone, PoSQLTimestampError},
     scalar::Scalar,
 };
-use alloc::sync::Arc;
+use alloc::{string::String, sync::Arc};
 use arrow::{
     array::{
         ArrayRef, BooleanArray, Decimal128Array, Decimal256Array, Int16Array, Int32Array,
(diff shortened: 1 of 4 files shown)


MyTH-zyxeon:
Quick review from the #183 nullable-column angle:

This looks like a useful safety slice while full nullable support is still in flight: it catches schema-level nullability even when the concrete Arrow array has no null slots, which is exactly the case that can silently lose the `nullable=true` bit today.

One small acceptance note I would suggest adding to the PR body or tests: `try_append_record_batch` and `try_from_record_batch` are also protected because they both route through `batch_to_columns`. The direct `batch_to_columns` test proves the shared helper, but an API-level regression would make the safety boundary more obvious to reviewers.

Integration watch-out: once a full nullable path such as #1955 lands, this unconditional `field.is_nullable()` rejection should stay scoped to the non-nullable `OwnedTable`/`Column` conversion path, or be replaced by `NullableOwnedColumn` ingestion. Calling that out in the error/doc comment would help avoid this temporary guard being mistaken for the final nullable-column behavior.

JayWhite2357:

PR #1775, opened 2026-05-19
test: reduce Dory compute setup sizes

zhengjynicolas:
## Summary

/claim #557

This trims the generated Dory public-parameter sizes used by the compute commitment test modules without removing any cases or assertions.

- `dory_compute_commitments_test` now uses `max_nu = 4`, which still covers the largest asserted static Gamma index (`Gamma_2[9]`).
- `dynamic_dory_compute_commitments_test` now uses `max_nu = 3`, which still covers the largest asserted dynamic Gamma indices (`Gamma_1[7]`, `Gamma_2[6]`).

## Timing

Focused hot-run command before this change:

```sh
cargo test -p proof-of-sql --lib --no-default-features --features="test" dory_compute_commitments
```

- before: test binary finished in 12.99s; `time` real 13.30s
- after: test binary finished in 8.74s; `time` real 9.03s

## Verification

```sh
cargo test -p proof-of-sql --lib --no-default-features --features="test" dory_compute_commitments
cargo fmt --all -- --check
git diff --check
```

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dory_compute_commitments_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dory_compute_commitments_test.rs
@@ -12,9 +12,11 @@ use ark_ec::pairing::Pairing;
 use ark_std::test_rng;
 use num_traits::Zero;
 
+const TEST_MAX_NU: usize = 4;
+
 #[test]
 fn we_can_compute_a_dory_commitment_with_varbinary_values() {
-    let public_parameters = PublicParameters::test_rand(5, &mut test_rng());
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:

- Scope checked: test-only changes in two Dory compute commitment test files (`34` additions / `30` deletions).
- Coverage rationale is easy to review: the reduced generated setup sizes still cover the largest asserted Gamma indices called out in the PR body (`Gamma_2[9]`, `Gamma_1[7]`, `Gamma_2[6]`), and the PR keeps the assertions/cases intact.
- GitHub Actions runs show `action_required`, so I am treating that as workflow permission/visibility rather than a code failure unless maintainers say otherwise.

This looks like a narrow, reviewable performance/test-runtime reduction for the #557 lane.

PR #1776, opened 2026-05-19
test: cover timestamptz column union

Yassinbrine:
/claim #560

## Summary
- Add focused regression coverage for unioning `TimestampTZ` columns.
- Exercise the `TimestampTZ` branch in `column_union` with multiple input slices.

## Testing

### DIFF
--- a/crates/proof-of-sql/src/base/database/union_util.rs
+++ b/crates/proof-of-sql/src/base/database/union_util.rs
@@ -245,7 +245,11 @@ pub fn table_union<'a, S: Scalar>(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::base::{map::IndexMap, scalar::test_scalar::TestScalar};
+    use crate::base::{
+        map::IndexMap,
+        posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

I checked this PR against the current body, diff, visible checks, comments, reviews, and review-thread state. The change is a narrow `crates/proof-of-sql/src/base/database/union_util.rs` test-only slice: it adds focused coverage for `column_union` over `TimestampTZ` columns with matching `PoSQLTimeUnit::Second` and UTC timezone, concatenating two input slices into the expected `[1, 2, 3, 4, 5]` result.

I do not see comments, reviews, or unresolved review threads on this PR right now. Orca checks are green; `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required`, which looks like fork workflow approval rather than an observed code failure. The author notes that cargo was unavailable locally, so CI approval is the validation gate.

Non-overlap note: this is distinct from the recent #560 coverage lanes around timestamp scale casts, planner conversion/filter-pushdown, AnalyzeError conversion, bit-distribution errors, evaluation-vector bounds, table/record-batch error display coverage, and standard serialization coverage.

PR #1783, opened 2026-05-19
test: cover table ref parsing and serialization

Yassinbrine:
/claim #560

## Summary
- Add focused coverage for `TableRef` construction from names and string components.
- Cover parse/display/serde round-tripping.
- Cover invalid component-count errors and `Equivalent` matching behavior.

## Test plan
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/database/table_ref.rs
+++ b/crates/proof-of-sql/src/base/database/table_ref.rs
@@ -142,3 +142,57 @@ impl<'d> Deserialize<'d> for TableRef {
         TableRef::from_str(&string).map_err(serde::de::Error::custom)
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer-facing validation note for `/claim #560`:

- Scope is focused: one test-only file, `crates/proof-of-sql/src/base/database/table_ref.rs`, adding coverage for table refs built from names/components, JSON serde round-trip, invalid component counts, and borrowed-reference equivalence.
- Public checks visible from head `991118d4385c` show Orca scans passing; the GitHub Actions runs I can see are still `action_required` rather than code-failing.
- I did not find open review threads, prior comments, or reviews on this PR, so the remaining blocker appears to be maintainer review / workflow approval rather than a requested code change.

No duplicate claim is being opened from this note.

PR #1784, opened 2026-05-20
test: reduce slow proof test verification work

elektropionir2:
/claim #557

## Summary
- make `exercise_verification` return the already verified result table so targeted slow tests do not immediately repeat the same successful verification
- reuse that verified table across additional proof expression and proof plan tests that were verifying twice only to assert the table
- shrink the two-sort-merge-join fixture while preserving the same expected output, duplicate join rows, filtered rows, and nonmatching rows

## Coverage preserved
The `exercise_verification` helper still performs the successful verification and all existing tamper checks. The changed tests now reuse the table from that successful verification instead of immediately running an identical second verification just to read the table.

The reduced two-sort-merge-join fixture still covers the behavior the test was asserting:
- two nested sort-merge joins
- duplicate output rows from duplicate matching IDs
- rows filtered out by `human != "Gretta"`
- rows filtered out by `hospital != "Clear Creek"`
- left-side rows that pass the first filter but do not survive the joins
- right-side rows that do not join
- the same expected result table as before

Follow-up commit `2677bb6a` extends the same verified-table reuse to more duplicated-verification tests in inequality, projection, slice, union, table, legacy filter, and sort-merge join coverage.

Follow-up commit `a56bc26d` removes more duplicated successful verification passes from filter, AND, and OR tests, including the random-table loops; those tests still assert the same expected tables using the successful verification result returned by the helper.

Follow-up commit `3d4552c1` applies the same helper-table reuse across add/subtract, cast, column, equality, literal, multiply, NOT, and scaling-cast expression tests. This removes another 25 duplicate successful verification calls while preserving the existing expected-table assertions.

Follow-up commit `7d225283` applies the same proof-plan cleanup to aggregate, group-by, legacy filter, and projection tests, removing 12 more duplicate successful verification passes while keeping the same expected-table assertions.

## Local timing
Measured with `BLITZAR_BACKEND=cpu` in a Docker Rust 1.91.1 environment with clang/lld, using warm focused runs after the initial compile:

| Test | Before | After |
| --- | ---: | ---: |
| `we_can_compare_columns_with_extreme_values` | 18.3s | 12.9s |
| `we_can_prove_and_get_the_correct_result_from_a_complex_query_involving_two_sort_merge_joins` | 19.1s | 9.6s |

## Validation
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib we_can_compare_columns_with_extreme_values`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib we_can_prove_and_get_the_correct_result_from_a_complex_query_involving_two_sort_merge_joins`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib sort_merge_join_exec_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib we_can_prove_and_get_the_correct_result_from_a_sort_merge_join`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib slice_exec_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib inequality_expr_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib and_expr_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib or_expr_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib filter_exec_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib proof_exprs`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib aggregate_exec_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib group_by_exec_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib projection_exec_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib legacy_filter_exec_test`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib proof_plans`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/sql/proof/verifiable_query_result_test_utility.rs
+++ b/crates/proof-of-sql/src/sql/proof/verifiable_query_result_test_utility.rs
@@ -27,10 +27,12 @@ pub fn exercise_verification(
     expr: &(impl ProofPlan + Serialize),
     accessor: &impl TestAccessor<RistrettoPoint>,
     table_ref: &TableRef,
-) {
-    res.clone()
+) -> OwnedTable<Curve25519Scalar> {
+    let verified_table = res
+        .clone()
         .verify(expr, accessor, &(), &[])
-        .expect("Verification failed");
+        .expect("Verification failed")
+        .table;
 
     // try changing the result
     let mut res_p = res.clone();
@@ -80,6 +82,8 @@ pub fn exercise_verification(
         fake_accessor.update_offset(table_ref, offset_generators + 1);
         assert!(res.clone().verify(expr, &fake_accessor, &(), &[]).is_err());
     }
+
+    verified_table
 }
 
 fn tampered_table<S: Scalar>(table: &OwnedTable<S>) -> OwnedTable<S> {
--- a/crates/proof-of-sql/src/sql/proof_exprs/add_subtract_expr_test.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/add_subtract_expr_test.rs
@@ -58,11 +58,7 @@ fn we_can_prove_a_typical_add_subtract_query() {
         ),
     );
     let verifiable_res = VerifiableQueryResult::new(&ast, &accessor, &(), &[]).unwrap();
-    exercise_verification(&verifiable_res, &ast, &accessor, &t);
-    let res = verifiable_res
-        .verify(&ast, &accessor, &(), &[])
-        .unwrap()
-        .table;
+    let res = exercise_verification(&verifiable_res, &ast, &accessor, &t);
     let expected_res = owned_table([
         smallint("a", [3_i16, 4]),
         bigint("c", [2_i16, 0]),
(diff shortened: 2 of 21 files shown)


elektropionir2:
Added one broader local validation pass to the PR body: \\BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib proof_plans\\ passes in Docker Rust 1.91.1 with 67 proof-plan tests passing. This covers the proof-plan files touched by the latest follow-up commit in addition to the narrower focused commands already listed.

MyTH-zyxeon:

This PR looks like a focused `/claim #557` runtime-reduction candidate rather than a behavior change. The main helper change makes `exercise_verification` return the table from the already-successful verification, then the touched proof expression/plan tests reuse that result instead of immediately running an identical second `verify(...)` just to assert the table.

What I checked from the public PR state:
- `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`
- Orca IaC, Secrets, and Vulnerabilities checks are all `SUCCESS`
- GitHub Actions runs for `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required`, with commit status showing `pending` and `statuses=[]`, so this appears maintainer-approval gated rather than a visible test failure
- The PR body lists focused `BLITZAR_BACKEND=cpu cargo test ...` coverage plus `cargo fmt --all -- --check` / `git diff --check`

The one review point I would focus on is whether all reused tables still come from the same successful verification path, not from an unverified shortcut. From the diff shape, the helper still performs the successful verification and the existing tamper checks, then returns the verified table, so the change seems aligned with reducing duplicate verification work while preserving the assertions.

JayWhite2357:

PR #1787, opened 2026-05-20
test: reuse verified proof expression tables

surim0n:
## Summary
- make `exercise_verification` return the table from its successful verification pass
- reuse that verified table across matching proof-expression and proof-plan tests
- remove 67 immediate duplicate `verify` calls while keeping the same expected-result assertions and all existing tamper checks

## Coverage preserved
`exercise_verification` still performs the successful verification, then still checks tampered result data, tampered final-round evaluations, tampered commitments, and offset changes. The updated tests now assert expected output against the table from that successful verification instead of running the exact same proof verification a second time.

## Local validation
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --lib --no-default-features --features test -- --list`
- `git diff --check`

## Platform note
Most changed tests require the `blitzar` feature. `blitzar-sys` only supports Linux and fails to build on this macOS machine, so I could not run the targeted `blitzar` tests locally. The no-default test build/list path above validates the shared helper outside that Linux-only feature gate; CI should be the authoritative Linux validation for the changed tests.

/claim #557

### DIFF
--- a/crates/proof-of-sql/src/sql/proof/verifiable_query_result_test_utility.rs
+++ b/crates/proof-of-sql/src/sql/proof/verifiable_query_result_test_utility.rs
@@ -27,10 +27,12 @@ pub fn exercise_verification(
     expr: &(impl ProofPlan + Serialize),
     accessor: &impl TestAccessor<RistrettoPoint>,
     table_ref: &TableRef,
-) {
-    res.clone()
+) -> OwnedTable<Curve25519Scalar> {
+    let verified_table = res
+        .clone()
         .verify(expr, accessor, &(), &[])
-        .expect("Verification failed");
+        .expect("Verification failed")
+        .table;
 
     // try changing the result
     let mut res_p = res.clone();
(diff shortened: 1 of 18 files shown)


MyTH-zyxeon:

- Scope checked: test-only proof-expression/proof-plan cleanup across 18 files (`+74/-338`).
- The central review point is `exercise_verification`: it still performs the successful verification and still exercises the tamper cases, but now returns the verified result table so matching tests can assert expected output without immediately running a duplicate `verify`.
- PR body calls out the preserved tamper coverage: result data, final-round evaluations, commitments, and offset changes still get checked.
- GitHub Actions runs show `action_required`, so I am treating that as workflow approval/visibility rather than a code failure unless maintainers say otherwise.

This is a broader-than-usual #557 test-runtime cleanup, but the duplicated-work removal is mechanical and the review hinge is whether the returned verified table is exactly the one produced by the retained successful verification path.

JayWhite2357:

PR #1791, opened 2026-05-20
test: cover owned column variant helpers

fabicholas:
/claim #560

## Rationale for this change
Adds focused unit coverage for OwnedColumn helper behavior across every supported variant. These helpers are small but broad match surfaces, so exercising every variant helps prevent regressions as column types evolve.

## What changes are included in this PR?
- Adds sample OwnedColumn values covering all variants.
- Tests len, is_empty, and column_type for every variant.
- Tests slice and try_permute across every variant.

## Are these changes tested?
cargo fmt --check passes. I attempted cargo test -p proof-of-sql owned_column_variants, but this Windows environment cannot build blitzar-sys because its build script is Linux-only (Unsupported OS. Only Linux is supported). I expect Linux CI to run the Rust tests.

### DIFF
--- a/crates/proof-of-sql/src/base/database/owned_column.rs
+++ b/crates/proof-of-sql/src/base/database/owned_column.rs
@@ -485,6 +485,90 @@ mod test {
         );
     }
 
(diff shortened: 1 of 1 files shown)


JayWhite2357:

PR #1804, opened 2026-05-20
fix: reject nullable Arrow arrays before owned conversion

junn-dev:
/claim #183

## Summary
- reject Arrow arrays with null values before converting them into OwnedColumn
- add coverage for boolean, integer, decimal, string, binary, and timestamp arrays with nulls
- preserve existing non-null conversion behavior

## Why
OwnedColumn does not support nullable columns yet. Some Arrow conversion paths used the backing alues() buffer directly, which can ignore Arrow's null bitmap and treat null slots as real values. This makes null handling inconsistent and can silently ingest invalid data while nullable-column support is still being built.

## Tests
- cargo test -p proof-of-sql --no-default-features --features arrow,test we_reject_nulls_for_all_arrow_types_before_reading_backing_values
- cargo test -p proof-of-sql --no-default-features --features arrow,test we_can_convert_between_owned_column_and_array_ref
- git diff --check

Note: default Windows test build pulls litzar-sys, which only supports Linux, so these were run with --no-default-features --features arrow,test.

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/owned_and_arrow_conversions.rs
+++ b/crates/proof-of-sql/src/base/arrow/owned_and_arrow_conversions.rs
@@ -23,7 +23,7 @@ use crate::base::{
 use alloc::sync::Arc;
 use arrow::{
     array::{
-        ArrayRef, BooleanArray, Decimal128Array, Decimal256Array, Int16Array, Int32Array,
+        Array, ArrayRef, BooleanArray, Decimal128Array, Decimal256Array, Int16Array, Int32Array,
         Int64Array, Int8Array, LargeBinaryArray, StringArray, TimestampMicrosecondArray,
         TimestampMillisecondArray, TimestampNanosecondArray, TimestampSecondArray, UInt8Array,
     },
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:
Maintainer-facing review assist for `/claim #183`:

I rechecked this as a narrow nullable-safety slice, not as the full nullable-column implementation. Current state is `OPEN` / `MERGEABLE`, but still `mergeStateStatus=BLOCKED` with `reviewDecision=REVIEW_REQUIRED`. Orca checks are green. The GitHub Actions runs for head `7a87575b1d720373295f44599b5d502294915995` show `CI-Code-Coverage` and `CI-Lint-And-Test` as `action_required`, and their job lists are empty, so this looks like the usual fork-workflow approval gate rather than a code failure.

Scope looks useful and bounded: 2 files, `+55/-4`, adding a single early `value.null_count() > 0` rejection before `OwnedColumn::try_from(&ArrayRef)` reads Arrow backing buffers, plus coverage across boolean, unsigned/signed integers, decimal128/decimal256, string, large binary, and second/millisecond/microsecond/nanosecond timestamp arrays.

- Approve/rerun the blocked fork workflows so the declared focused tests execute on GitHub.
- Confirm maintainers want this counted as a conservative #183 hardening slice while full nullable support is still open; it prevents silent null-bitmap loss but does not attempt SQL null semantics.
- Keep the regression focus on null rejection before `values()`/backing-buffer reads, because that is the actual safety boundary this PR improves.

No author code change looks necessary from the visible diff unless CI reports a compile/test issue after workflow approval.

JayWhite2357:

PR #1809, opened 2026-05-20
test: cover column field accessors and serde

Yassinbrine:
/claim #560

## Summary
- Add focused coverage for `ColumnField` construction and accessor methods.
- Cover clone/equality behavior.
- Cover serde round-tripping for column field metadata.

## Test plan
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/database/column_field.rs
+++ b/crates/proof-of-sql/src/base/database/column_field.rs
@@ -31,3 +31,33 @@ impl ColumnField {
         self.data_type
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer-facing validation note for `/claim #560`:

- Scope is very small: one test-only file, `crates/proof-of-sql/src/base/database/column_field.rs`, adding accessor, clone/equality, and serde round-trip coverage for `ColumnField`.
- Public checks visible from head `35aac14e297e` show Orca scans passing; the GitHub Actions runs I can see are still `action_required` rather than code-failing.
- I did not find open review threads, prior comments, or reviews on this PR, so the remaining blocker appears to be maintainer review / workflow approval rather than a requested code change.

No duplicate claim is being opened from this note.

PR #1815, opened 2026-05-20
test: cover reference conversion helper

Yassinbrine:
/claim #560

## Summary
- Add focused coverage for the `RefInto` blanket implementation.
- Verify that reference-based conversion does not consume the source value.
- Cover explicit target type inference for `ref_into()`.

## Test plan
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/ref_into.rs
+++ b/crates/proof-of-sql/src/base/ref_into.rs
@@ -23,3 +23,35 @@ where
         self.into()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[derive(Debug, Eq, PartialEq)]
+    struct RefSource {
+        value: usize,
+    }
+
+    impl From<&RefSource> for usize {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review assist for maintainers:

I reviewed this as a narrow #560 coverage PR. The patch is limited to `crates/proof-of-sql/src/base/ref_into.rs` and adds tests around the `RefInto` blanket implementation without changing production behavior.

What it covers:
- `ref_into()` dispatches through `From<&RefSource>` and returns the expected target value.
- The source value remains available after conversion, which is the key reference-conversion behavior this helper is meant to preserve.
- Explicit target type inference works when assigning `source.ref_into()` into a typed `usize` binding.

Visible status while reviewing: head `c1b7f4282c833f599ade2f83e131340028d57f1a`; Orca IaC, Secrets, and Vulnerabilities checks are green; no comments or reviews were present before this note. The PR body says cargo was not available locally, so the remaining maintainer decision is whether the CI/test environment accepts the focused unit coverage as written.

PR #1834, opened 2026-05-20
test: precompute EVM verifier arguments

zergzorg:
/claim #557

## Summary
- precompute the serialized EVM verifier arguments once per `evm_verifier_all` call
- reuse the same result/query/params/proof/table-length/commitment values across the four forge variants
- remove unconditional `dbg!` formatting from the hot verifier path

## Expected effect
- avoids repeating plan column/table scans, `EVMProofPlan::new(plan.clone())`, bincode serialization, hex encoding, and captured stderr formatting four times per EVM verifier case
- preserves all four forge invocations (`plain`, `--via-ir`, `--optimize`, and both) and does not remove native verification assertions
- effect is intentionally limited to Rust-side pre-forge setup; forge execution still dominates, so CI/maintainer timing should be the source of truth for bounty accounting

## Validation
- `git diff --check`
- `cargo test --package proof-of-sql-planner --test evm_tests --no-run`
- `cargo test --package proof-of-sql-planner --test evm_tests we_can_verify_a_slice_exec_using_the_evm -- --ignored --nocapture`

## Notes
- `cargo test --package proof-of-sql-planner --all-features --test evm_tests --no-run` is not runnable on this Apple Silicon environment because `halo2curves/asm` only supports x86_64.
- The real Foundry-based EVM smoke path currently fails locally on both this branch and a clean `origin/main`, so I did not use local forge timings as evidence; CI/maintainer timing is the source of truth.

### DIFF
--- a/crates/proof-of-sql-planner/tests/evm_tests.rs
+++ b/crates/proof-of-sql-planner/tests/evm_tests.rs
@@ -30,14 +30,22 @@ use proof_of_sql::{
 use proof_of_sql_planner::sql_to_proof_plans;
 use sqlparser::{ast::Ident, dialect::GenericDialect, parser::Parser};
 
+struct EvmVerifierArgs {
+    result_hex: String,
+    query_hex: String,
+    params: String,
+    proof_hex: String,
+    table_lengths: String,
+    commitments: String,
+}
+
 #[expect(clippy::missing_panics_doc)]
-fn evm_verifier_with_extra_args(
+fn evm_verifier_args(
     plan: &DynProofPlan,
     params: &str,
     verifiable_result: &VerifiableQueryResult<HyperKZGCommitmentEvaluationProof>,
     accessor: &impl CommitmentAccessor<HyperKZGCommitment>,
-    extra_args: &[&'static str],
-) -> bool {
+) -> EvmVerifierArgs {
     let commitments = plan
         .get_column_references()
         .into_iter()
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

- Current PR gate: `state=OPEN`, `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`.
- Scope is narrow and test-only: 1 changed file, 37 additions / 19 deletions in `crates/proof-of-sql-planner/tests/evm_tests.rs`.
- Visible security checks are green: Orca Infrastructure as Code, Secrets, and Vulnerabilities all report `SUCCESS` on `74fb1755106965eb17f4002b029ed781422b4c5e`.
- GitHub Actions runs `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required`, not failed; there are no failing job logs to debug from this head SHA.
- Patch review: the PR factors the repeated EVM verifier argument construction into `EvmVerifierArgs`, computes the bincode/hex/table-length/commitment payloads once per `evm_verifier_all` call, then reuses them across the four existing forge variants (`plain`, `--via-ir`, `--optimize`, and both). Native verification coverage and all four forge invocations stay in place, and the removed `dbg!` calls reduce unconditional formatting/noise on the verifier path.

Suggested maintainer action: approve/run the blocked Actions and review this as a small #557 test-performance variant if the precomputed verifier-argument helper is acceptable.

zergzorg:
Resolved the merge conflict after the current upstream main changes in #2609 by merging main into this branch without rewriting history. The resolution keeps the new explicit HyperKZG test setup path (`TEST_SETUP_FILE`) and the existing precomputed EVM verifier-argument helper from this PR, so the scope remains the same test-only performance cleanup.

Validation run from a clean temporary checkout:
- `cargo fmt --check`
- `cargo test -p proof-of-sql-planner --test evm_tests --no-run`

GitHub now reports the PR as `BLOCKED`/review-required rather than `DIRTY`; Orca checks are queued on the new head `d68dec3`.

PR #1838, opened 2026-05-20
test: cover arrow schema utility

ProtonsAndElectrons:
/claim #560

## Summary
- add unit coverage for empty Arrow schemas in `get_posql_compatible_schema`
- cover Float16, Float32, and Float64 conversion to `Decimal256(20, 10)`
- assert non-float fields preserve their data type and nullability

## Tests
- `cargo test -p proof-of-sql --lib --no-default-features --features=arrow base::database::arrow_schema_utility`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/database/arrow_schema_utility.rs
+++ b/crates/proof-of-sql/src/base/database/arrow_schema_utility.rs
@@ -35,3 +35,55 @@ pub fn get_posql_compatible_schema(schema: &SchemaRef) -> SchemaRef {
 
     Arc::new(Schema::new(new_fields))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn we_can_convert_empty_schema_to_posql_compatible_schema() {
+        let schema = Arc::new(Schema::empty());
+
+        let converted = get_posql_compatible_schema(&schema);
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only change in `crates/proof-of-sql/src/base/database/arrow_schema_utility.rs` (`+52/-0`). It covers `get_posql_compatible_schema` for an empty schema and for mixed Float16/Float32/Float64 plus non-float fields, asserting that all float widths normalize to `Decimal256(20, 10)` while names, order, and nullability are preserved for the checked fields.

The main review point I would confirm is whether the conversion contract intentionally treats Float16, Float32, and Float64 identically as `Decimal256(20, 10)`. This PR is useful if that is the intended compatibility boundary, but it will make that precision/scale choice explicit in tests. It also does not assert field metadata preservation, so maintainers may want to decide whether metadata should be part of this utility's stable behavior.

I did not run the Rust suite locally in this pass. The PR body records targeted cargo/fmt/diff checks by the contributor. Visible Orca checks are green; the two GitHub Actions runs for this head currently show `action_required`, and the combined commit status API returned `state=pending` with no statuses.

JayWhite2357:

PR #1841, opened 2026-05-20
test(planner): add uppercase visitor edge-case coverage

liangtovi-debug:
## Summary
- add focused unit coverage for `uppercase_identifier` to verify quote style is preserved
- add planner SQL coverage for multi-part table names and join-side compound identifiers
- keep scope test-only in `crates/proof-of-sql-planner/src/uppercase_column_visitor.rs`

## Validation
- `cargo test -p proof-of-sql-planner uppercase_column_visitor -- --nocapture`

/claim #560

### DIFF
--- a/crates/proof-of-sql-planner/src/uppercase_column_visitor.rs
+++ b/crates/proof-of-sql-planner/src/uppercase_column_visitor.rs
@@ -47,8 +47,12 @@ pub fn statement_with_uppercase_identifiers(mut statement: Statement) -> Stateme
 
 #[cfg(test)]
 mod tests {
-    use super::statement_with_uppercase_identifiers;
-    use sqlparser::{dialect::GenericDialect, parser::Parser};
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, planner test-only change in `crates/proof-of-sql-planner/src/uppercase_column_visitor.rs` (`+28/-2`). It adds direct coverage that `uppercase_identifier` preserves quote style while uppercasing the identifier value, plus SQL parser coverage for qualified table names and join-side compound identifiers.

The main review point I would check before merging is the alias behavior captured by the expected SQL: table/column references are uppercased (`T.COL_A`, `T.ID`, `S.ID`), but aliases remain written as `AS t` / `s`. That may be exactly the current visitor contract, but this test will lock that distinction in, so it is worth confirming that aliases should remain case-preserved while references are normalized.

JayWhite2357:

PR #1846, opened 2026-05-20
test: cover sumcheck random scalars

ProtonsAndElectrons:
/claim #560

# Rationale for this change

`SumcheckRandomScalars` splits verifier randomness into subpolynomial multipliers and the entrywise evaluation point, then turns that point into the entrywise multiplier vector used during sumcheck verification. This adds focused coverage for that helper without changing production behavior.

- Registers a dedicated `sumcheck_random_scalars_test` module.
- Covers the constructor split between `subpolynomial_multipliers` and `entrywise_point`.
- Covers `compute_entrywise_multipliers` for a truncated table length and for an empty entrywise point.

# Are these changes tested?

- `cargo test -p proof-of-sql --lib --no-default-features --features std sumcheck_random_scalars`
- `cargo fmt --check`
- `git diff --check`


JayWhite2357:

PR #1848, opened 2026-05-20
test: cover sumcheck prover round state

GYC-lab:
## Summary

- Add focused tests for `ProverState::new` and `ProverState::create`.
- Cover `prove_round` first/follow-up round behavior, verifier-message error paths, and private helper behavior.

/claim #560

## Verification

- `cargo fmt --check`
- `git diff --check`
- `cargo test --manifest-path crates/proof-of-sql/Cargo.toml proof_primitive::sumcheck::prover --no-default-features --features="test"` locally with a temporary `sqlparser` path override because GitHub dependency fetches were timing out in this environment.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/sumcheck/prover_round.rs
+++ b/crates/proof-of-sql/src/proof_primitive/sumcheck/prover_round.rs
@@ -223,19 +223,13 @@ mod tests {
 
         in_place_fix_variable(&mut multiplicand, TestScalar::from(3), 2);
 
-        assert_eq!(
-            multiplicand,
-            vec![
-                TestScalar::from(4),
-                TestScalar::from(16),
-                TestScalar::from(64),
-                TestScalar::from(256),
-                TestScalar::from(16),
-                TestScalar::from(32),
-                TestScalar::from(64),
-                TestScalar::from(128),
-            ]
-        );
+        let expected_prefix = [
+            TestScalar::from(4),
+            TestScalar::from(16),
+            TestScalar::from(64),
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:

This is a `/claim #560` test-only sumcheck prover coverage slice at current head `a9346e10` (`+222/-0`, two files). It adds `ProverState::new` / `ProverState::create` tests plus `prove_round`, `in_place_fix_variable`, and `vec_elementwise_add` helper coverage, with no production-code changes.

Current head appears to have addressed the prefix/scratch-tail part of Copilot's `in_place_fix_variable` concern by asserting only the active prefix. The two review points I would still decide before merge are:

- `prover_state_new_initializes_round_state` calls `ProverState::new` with `list_of_products` referencing multiplicand indices `[0, 1]` while `flattened_ml_extensions` contains only one MLE. That is fine if the test is intentionally just checking field storage, but it creates an impossible prover state if evaluated later; using `[0]` or adding a second MLE would make the fixture invariant-valid.
- `vec_elementwise_add_adds_until_the_shorter_input_ends` locks in zip/truncate behavior for unequal lengths. If sumcheck reductions are expected to combine equal `degree + 1` vectors, this test may make a future defensive length assertion harder to add.

Visible status from my side: Orca scan check runs are green; `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required` for `a9346e10`, and the combined commit status API is still `pending` with `statuses=[]`. I did not run the Rust suite locally in this pass.

JayWhite2357:

PR #1857, opened 2026-05-20
Add column repetition coverage tests

Amhitox:
/claim #560

This adds focused unit coverage for the column repetition helpers without changing production behavior.

What changed:
- added coverage for additional numeric repetition branches: tinyint, smallint, bigint, int128
- added decimal and timestamp repetition cases for column-wise repetition
- added scalar, decimal, and timestamp cases for elementwise repetition
- added zero-repeat assertions for empty output behavior

Tested with:
```bash
CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=gcc RUSTFLAGS='' cargo test -p proof-of-sql column_repetition_operation
cargo fmt --all
git diff --check
```

The linker override is only for my local environment because this checkout expects clang, which is not installed here.


JayWhite2357:

PR #1871, opened 2026-05-20
test: cover metadata map and decimal parsing slices

Da7-Tech:
/claim #560

## Summary
- add a focused test for `ColumnCommitmentMetadataMap::from_column_fields_with_max_bounds` covering ordered max bounds and unordered metadata
- add parser coverage proving only DECIMAL columns with precision above 38 and explicit scale are collected
- keep the patch test-only and scoped to two small, non-overlapping coverage slices

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features test column_commitment_metadata_map -- --nocapture` -> 6 passed
- `cargo test -p proof-of-sql --lib --no-default-features --features test utils::parse -- --nocapture` -> 2 passed
- `cargo fmt --all -- --check` -> passed
- `git diff --check` -> passed

Coverage accounting note: `cargo llvm-cov` is not installed in my local environment, so final covered-line accounting is left to the maintainer baseline. The default-feature test path on macOS arm64 also hits the existing `halo2curves` asm restriction, so validation used the project's no-default targeted test pattern.

Manual payment fallback if needed: PayPal `aa.631@hotmail.com`.

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/column_commitment_metadata_map.rs
+++ b/crates/proof-of-sql/src/base/commitment/column_commitment_metadata_map.rs
@@ -141,6 +141,7 @@ mod tests {
     use crate::base::{
         commitment::{column_bounds::Bounds, ColumnBounds},
         database::{owned_table_utility::*, ColumnType, OwnedTable},
+        math::decimal::Precision,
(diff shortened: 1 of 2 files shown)


JayWhite2357:

PR #1874, opened 2026-05-20
ci: cache nix flake checks

zergzorg:
/claim #557

# Rationale for this change

This is a small CI-runtime slice for #557. The `check-nix-flake` job installs Nix and then runs `nix flake check` plus `nix develop --command cargo check`; without a Nix store cache, each run has to rebuild/refetch the flake inputs and dev shell closure.

- Add `nix-community/cache-nix-action@v7` immediately after `cachix/install-nix-action@v31` in the `check-nix-flake` job.
- Key the cache by runner OS plus `flake.nix`, `flake.lock`, and `rust-toolchain.toml`, so dependency/toolchain changes invalidate it.
- Allow fallback to the latest `nix-${{ runner.os }}-` cache prefix.
- Cap the saved Linux Nix store cache at `1G` before upload to avoid oversized cache artifacts.

# Are these changes tested?

Yes, workflow-only validation:

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/lint-and-test.yml"); puts "workflow yaml ok"'`
- `git diff --check`

I also checked the action metadata for `nix-community/cache-nix-action@v7`; the inputs used here are supported by its `action.yml`.

# Limitations

I did not run the GitHub-hosted `check-nix-flake` job locally. The expected improvement is visible on repeated CI runs where the Nix store cache can be restored instead of rebuilding the same flake/dev-shell closure.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -404,6 +404,12 @@ jobs:
         uses: actions/checkout@v4
       - name: Install nix
         uses: cachix/install-nix-action@v31
+      - name: Restore Nix store cache
+        uses: nix-community/cache-nix-action@v7
+        with:
+          primary-key: nix-${{ runner.os }}-${{ hashFiles('flake.nix', 'flake.lock', 'rust-toolchain.toml') }}
+          restore-prefixes-first-match: nix-${{ runner.os }}-
+          gc-max-store-size-linux: 1G
       - name: Check flake
         run: nix flake check
       - name: Check project with flake
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -405,10 +405,9 @@ jobs:
       - name: Install nix
         uses: cachix/install-nix-action@v31
       - name: Restore Nix store cache
-        uses: nix-community/cache-nix-action@v7
+        uses: nix-community/cache-nix-action@7df957e333c1e5da7721f60227dbba6d06080569
         with:
           primary-key: nix-${{ runner.os }}-${{ hashFiles('flake.nix', 'flake.lock', 'rust-toolchain.toml') }}
-          restore-prefixes-first-match: nix-${{ runner.os }}-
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

- Current visible state: `OPEN` / `MERGEABLE`, with `mergeStateStatus=BLOCKED` and `reviewDecision=REVIEW_REQUIRED`.
- Scope: 1 workflow file, `.github/workflows/lint-and-test.yml`, `+6/-0`, adding `nix-community/cache-nix-action@v7` after `cachix/install-nix-action@v31` for the `check-nix-flake` job.
- Checks: Orca IaC, Secrets, and Vulnerabilities are green. GitHub Actions `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required`; both run job lists returned `jobs=[]`, so this looks like fork-workflow approval/review gating rather than a visible test failure.
- Combined commit statuses for head `ca3594b5ca210b4c98b4e970f014d2e60f257f78` returned `statuses=[]`.
- Review surface before this note: no top-level comments or reviews were visible.

Pre-merge checks I would focus on: confirm the project is comfortable introducing this third-party Nix cache action pinned by tag rather than SHA, verify fork-approved cold/warm `check-nix-flake` behavior, and make sure the `restore-prefixes-first-match: nix-${{ runner.os }}-` fallback cannot restore an incompatible store when `flake.nix`, `flake.lock`, or `rust-toolchain.toml` changes.

Suggested maintainer action: approve/rerun the fork workflows, then merge only if the Nix cache action policy and cold/warm workflow behavior are acceptable.

zergzorg:
Follow-up hardening pushed in `ba8c8e0d`:

- pinned `nix-community/cache-nix-action` to the current `v7` tag commit instead of a movable tag
- removed the broad OS-only restore fallback so Nix store restores use the exact flake/toolchain cache key

Validation:
- `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/lint-and-test.yml')"`\n- `git diff --check`

PR #1805, opened 2026-05-20
test: run table test accessor without blitzar

mitre88:
/claim #560

## Summary
- run the existing `table_test_accessor_test` module under plain `#[cfg(test)]` so it contributes to the no-default `test` coverage path
- keeps `owned_table_test_accessor_test` behind `feature = "blitzar"` because that slice is already claimed separately

## Coverage
- `crates/proof-of-sql/src/base/database/table_test_accessor.rs`: 72/88 covered lines in the no-default LCOV run

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features="test" table_test_accessor`
- `LLVM_COV=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/llvm-cov LLVM_PROFDATA=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/llvm-profdata cargo llvm-cov -p proof-of-sql --lib --no-default-features --features="test" --lcov --output-path target/table-test-accessor-no-blitzar.lcov`
- `cargo fmt --all -- --check`
- `git diff --check`

AI-assisted with OpenAI Codex; I reviewed and validated the diff before submitting.

PR #1899, opened 2026-05-20
test: cover table test accessor

nam157:
/claim #560

## Summary
- run the existing `TableTestAccessor` tests in the standard Arrow-profile test build instead of requiring `blitzar`
- cover accessor creation, borrowed-table insertion, column access, commitments, metadata, schema lookup, and offset updates
- keep production behavior unchanged; test cfg only

## Newly covered production lines
- crates/proof-of-sql/src/base/database/table_test_accessor.rs: 20-25, 40-42, 44-46, 52-60, 66-68, 78-87, 97-106, 113-115, 120-122, 125, 127-134, 139-148, 153-157
- 72 lines that were previously uncovered in the local baseline LCOV

## Validation
- cargo fmt --all -- --check
- cargo test -p proof-of-sql --no-default-features --features="arrow" table_test_accessor -- --nocapture
- cargo llvm-cov -p proof-of-sql --no-default-features --features="arrow" --lcov --output-path target/table-test-accessor.lcov -- table_test_accessor
- cargo test -p proof-of-sql --no-default-features --features="arrow"

Note: workspace-level arrow/blitzar test attempts on this Apple Silicon machine can fail because halo2curves enables asm only for x86_64.

### DIFF
--- a/crates/proof-of-sql/src/base/database/mod.rs
+++ b/crates/proof-of-sql/src/base/database/mod.rs
@@ -113,7 +113,7 @@ mod owned_table_test_accessor_test;
 
 mod table_test_accessor;
 pub use table_test_accessor::TableTestAccessor;
-#[cfg(all(test, feature = "blitzar"))]
+#[cfg(test)]
 mod table_test_accessor_test;
 
 /// Module providing utilities for filtering columns based on selection vectors.


MyTH-zyxeon:
Maintainer-facing deconfliction note for #560 routing:

I checked this PR as a #560 coverage submission before review. Current state is `OPEN`, `MERGEABLE`, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED` at head `93a7ea19286f2d316bf09f7c386d0c6470ccbee7`; there were no existing issue comments, reviews, or review-thread comments when I checked. The visible Orca checks are green; raw combined status has no contexts, and the GitHub Actions check suites are `action_required`.

The patch is the same one-line test-only gate change as #1805 in `crates/proof-of-sql/src/base/database/mod.rs` (`#[cfg(all(test, feature = "blitzar"))]` -> `#[cfg(test)]` for `table_test_accessor_test`). Since #1805 is already open with the same diff and an earlier `/claim #560`, this looks like a duplicate/competing submission for the same `TableTestAccessor` no-default coverage slice rather than additive coverage.

For award/routing hygiene, I would de-dupe this against #1805 and treat only one PR as the `table_test_accessor_test` coverage lane. I did not rerun local cargo in this pass because the local `/tmp/sxt-proof-of-sql` checkout was unavailable; this note is triage/deconfliction only.

PR #1923, opened 2026-05-21
test: cover decimal proof expression metadata

MolhamHamwi:
/claim #560

# Rationale for this change

The `DecimalProofExpr` default metadata helpers were not exercised in the std-only coverage baseline. This adds a narrow unit test that constructs decimal add, subtract, and multiply proof expressions and asserts that their `precision()` and `scale()` helpers agree with the expression result type metadata.

- Adds an in-file unit test for `DecimalProofExpr` metadata helpers.
- Covers the helper implementations through `AddExpr`, `SubtractExpr`, and `MultiplyExpr`.

# Are these changes tested?

Yes.

- `cargo test -p proof-of-sql --no-default-features --features std decimal_proof_expr_metadata -- --nocapture`
- `cargo llvm-cov -p proof-of-sql --no-default-features --features std --lib --lcov --output-path target/proof-expr-decimal-metadata.lcov -- decimal_proof_expr_metadata`
- `git diff --check`

Focused LCOV for `crates/proof-of-sql/src/sql/proof_exprs/proof_expr.rs` moved the production helper lines from 0 hits in the baseline to covered in this focused run.

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/proof_expr.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/proof_expr.rs
@@ -78,3 +78,46 @@ pub(crate) trait DecimalProofExpr: ProofExpr {
         self.data_type().scale().expect("Scale should be valid")
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::DecimalProofExpr;
+    use crate::{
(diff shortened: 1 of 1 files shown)


JayWhite2357:

PR #1927, opened 2026-05-21
test: reduce dynamic Dory compute setup sizes

justtryingtohelp46:
## Summary
- shrink dynamic Dory compute commitment test setup generation from `max_nu = 5` to the smallest setup size each test actually exercises
- add named setup-size constants for empty, two-column, and three-column dynamic commitment cases
- keep all existing assertions, inputs, offsets, and expected commitment terms unchanged

## Why this helps #557
These tests only need the setup rows referenced by their explicit expected commitments. Generating `PublicParameters::test_rand(5, ...)` creates unused Dory setup data, so this trims avoidable setup/prover construction work without weakening the tests.

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features test dynamic_dory_compute_commitments_test -- --nocapture`
- `git diff --check`

/claim #557

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_compute_commitments_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_compute_commitments_test.rs
@@ -11,9 +11,13 @@ use crate::{
 use ark_ec::pairing::Pairing;
 use num_traits::Zero;
 
+const EMPTY_COLUMN_SETUP_NU: usize = 1;
+const TWO_COLUMN_SETUP_NU: usize = 2;
+const THREE_COLUMN_SETUP_NU: usize = 3;
+
 #[test]
 fn we_can_handle_calling_with_an_empty_committable_column() {
-    let public_parameters = PublicParameters::test_rand(5, &mut test_rng());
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Thanks for keeping this PR scoped to #557. I did a read-through of the patch against the Dory setup-size acceptance goal.

This looks like a narrow test-performance slice rather than a behavior change:

- every edit is in `crates/proof-of-sql/src/proof_primitive/dory/dynamic_dory_compute_commitments_test.rs`; the production commitment path is untouched.
- the old blanket `PublicParameters::test_rand(5, ...)` setup is replaced with named case-size constants: empty-column cases use `1`, two-column/mixed cases use `2`, and three-column/offset cases use `3`.
- the existing signed, unsigned, offset, mixed-column, and empty-column assertions are preserved, so the PR keeps the same commitment semantics while avoiding unnecessary setup generation.

The only merge-side blocker I can see is maintainer workflow approval/rerun: `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required`, while the visible Orca checks are green. I would keep this under the existing `/claim #557` performance lane rather than opening any new claim.

JayWhite2357:

PR #1936, opened 2026-05-21
test: cover offset byte conversions

alexgduarte:
## Summary

- Add direct unit coverage for Dory `OffsetToBytes` conversions.
- Cover signed edge-value biasing for `i8`, `i16`, `i32`, `i64`, and `i128`.
- Cover raw `u8`, `u64`, `bool`, and `[u64; 4]` little-endian byte output.

/claim #560

## Validation

- `cargo test -p proof-of-sql --lib --no-default-features --features test offset_to_bytes`
- `rustfmt --check crates/proof-of-sql/src/proof_primitive/dory/offset_to_bytes.rs`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/offset_to_bytes.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/offset_to_bytes.rs
@@ -60,3 +60,73 @@ impl OffsetToBytes<32> for [u64; 4] {
         bytemuck::cast(*self)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::OffsetToBytes;
+
+    #[test]
+    fn signed_offsets_map_min_zero_and_max_edges() {
+        assert_eq!(i8::MIN.offset_to_bytes(), [0]);
+        assert_eq!((-1_i8).offset_to_bytes(), [127]);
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow `/claim #560` test-only PR in `crates/proof-of-sql/src/proof_primitive/dory/offset_to_bytes.rs` (`+70/-0`). The new tests cover signed offset biasing at min/zero/max boundaries for `i8`, `i16`, `i32`, `i64`, and `i128`, plus raw byte output for `u8`, `u64`, `bool`, and `[u64; 4]` little-endian limbs.

The main review point I would check is whether the signed coverage intentionally uses boundary-value assertions rather than a roundtrip helper; as written, it should catch endian or bias regressions directly, but it does not add randomized/property coverage. The array-limb case is useful because it verifies contiguous limb ordering rather than just one scalar conversion.

I did not run the Rust suite locally. Visible Orca checks are green; the GitHub Actions runs for `CI-Lint-And-Test` and `CI-Code-Coverage` currently show `action_required`, and the combined commit status is `pending` with no individual statuses returned by the status API.

JayWhite2357:

PR #1484, opened 2026-05-12
test: cover committable column conversions

himanalot:
## Summary
- add non-blitzar-gated coverage for borrowed `Column` to `CommittableColumn` conversion arms
- cover primitive `&[u8]` and `&[bool]` slice conversions

## Verification
- `cargo test -p proof-of-sql base::commitment::committable_column --no-default-features --features=test`
- `cargo llvm-cov -p proof-of-sql --no-default-features --features=test --lcov --output-path /tmp/sxt-committable-after.lcov -- base::commitment::committable_column`
- `cargo fmt --check`
- `git diff --check`
- `source scripts/check_commits.sh`

/claim #560

PR #1945, opened 2026-05-21
test: cover committable column conversions

alexgduarte:
/claim #560

## Summary
- add non-blitzar unit coverage for committable-column length and type helpers
- cover borrowed Column conversions across primitive, scalar, decimal, string, binary, and timestamp variants
- cover owned Uint8, VarChar, and VarBinary conversions without requiring blitzar

## Validation
- rustfmt --check crates/proof-of-sql/src/base/commitment/committable_column.rs
- cargo test -p proof-of-sql --no-default-features --features arrow,test tests_without_blitzar
- cargo test -p proof-of-sql --no-default-features --features arrow,test committable_column
- git diff --check

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -247,6 +247,173 @@ impl<'a, 'b> From<&'a CommittableColumn<'b>> for Sequence<'a> {
     }
 }
 
+#[cfg(all(test, not(feature = "blitzar")))]
+mod tests_without_blitzar {
+    use super::*;
+    use crate::base::scalar::test_scalar::TestScalar;
+    use alloc::{string::String, vec};
+
+    #[test]
+    fn we_can_get_type_and_length_without_blitzar() {
+        let bools = [true, false, true];
+        let uint8s = [1_u8, 2, 3];
+        let tinyints = [-1_i8, 0, 1];
+        let smallints = [-2_i16, 0, 2];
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist for `#560`:

I checked this against the adjacent open committable-column conversion PRs (`#1484`, `#1890`, and `#2201`). This version is the broader non-blitzar slice: it covers `len` / `is_empty` / `column_type` across the primitive, decimal, scalar, var-char, var-binary, and timestamp variants, then adds owned-column conversion coverage for `Uint8`, `VarChar`, and `VarBinary`.

Suggested maintainer checks before choosing between the overlapping entries:

1. This module is gated with `#[cfg(all(test, not(feature = "blitzar")))]`, while the nearby alternatives use plain `#[cfg(test)]`. The PR validation path should include the no-default-features / non-blitzar test target so these tests are actually compiled.
2. The owned `VarBinary` case currently asserts only the output limb count. If this is the version to keep, an exact limb-value assertion would make that regression check stronger.
3. Because several open PRs touch the same `committable_column.rs` coverage area, compare the final accepted patch against `#1484` / `#1890` to avoid merging duplicate coverage with weaker assertions.

Visible state at the checked head `25de9996524b03c9ab495ec432346f9199856814`: open, mergeable, `BLOCKED` / `REVIEW_REQUIRED`, with Orca checks green and no prior top-level comments or reviews visible.

JayWhite2357:

PR #1946, opened 2026-05-21
test: cover byte matrix constant inputs

alexgduarte:
## Summary
- add direct coverage for empty input to compute_varying_byte_matrix
- add direct coverage for constant columns with no varying byte matrix entries
- keep the slice scoped to byte_matrix_utils tests only

## Validation
- cargo test -p proof-of-sql --no-default-features --features arrow,test byte_matrix
- rustfmt --check crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs
- git diff --check

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs
+++ b/crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs
@@ -39,6 +39,32 @@ mod tests {
     use bumpalo::Bump;
 
     #[test]
+    fn we_can_compute_varying_byte_matrix_for_empty_columns() {
+        let alloc = Bump::new();
+        let scalars = Vec::<TestScalar>::new();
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only `/claim #560` change in `crates/proof-of-sql/src/base/byte/byte_matrix_utils.rs` (`+26/-0`). It adds explicit coverage for `compute_varying_byte_matrix` on empty input and on constant scalar columns where no byte positions vary, checking that the returned byte distribution matches `ByteDistribution::new`, that `varying_byte_count()` is zero, and that the varying-column matrix is empty.

The main review point I would confirm is whether the constant-column case should also cover a non-zero higher byte or a negative/signed `TestScalar` value, since this currently pins only the all-42 positive fixture. The current tests are still a useful no-varying-byte regression guard, but one additional fixture could make the intended byte-distribution boundary clearer.

JayWhite2357:

PR #1949, opened 2026-05-21
Remove string conversion bounds from Scalar

NAEL592:
## Summary
- remove string conversion bounds from `Scalar`
- add explicit `ScalarExt::from_str_via_hash`
- route generic VarChar/string scalar conversions through the explicit hash helper

Partially addresses #228 (string conversion bounds only).

## Verification
- `cargo fmt --check`
- `cargo check -p proof-of-sql --no-default-features --features "std test"`
- `cargo test -p proof-of-sql --no-default-features --features "std test" from_str_via_hash_matches_byte_hashing`

Note: default `cargo test -p proof-of-sql ...` is blocked on Windows because the default `perf` feature enables `blitzar-sys`, whose build script supports Linux only.

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -162,7 +162,7 @@ impl<'a, S: Scalar> From<&'a OwnedColumn<S>> for CommittableColumn<'a> {
             OwnedColumn::VarChar(strings) => CommittableColumn::VarChar(
                 strings
                     .iter()
-                    .map(Into::<S>::into)
+                    .map(|s| S::from_str_via_hash(s))
                     .map(Into::<[u64; 4]>::into)
                     .collect(),
             ),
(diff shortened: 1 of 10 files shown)


MyTH-zyxeon:
Review-assist for #228 string-conversion item

I checked this as the explicit string-conversion slice of #228.

What looks aligned:
- Removes the generic `From<&str>`, `From<&String>`, and `From<String>` bounds from `Scalar`, which directly addresses the implicit string conversion part of #228.
- Adds `ScalarExt::from_str_via_hash(&str)` and routes VarChar/string scalar conversion call sites through it, preserving the existing byte-hash semantics via `from_byte_slice_via_hash(val.as_bytes())`.
- Updates the string inner-product path from generic ref casting to `inner_product_with_strings`, so VarChar data is hashed explicitly before multiplying by the scalar vector.
- Adds a focused regression test asserting `from_str_via_hash("abc") == from_byte_slice_via_hash(b"abc")`.

Validation/status I can see:
- Visible rollup has Orca IaC, Secrets, and Vulnerabilities passing.
- `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required` for head `d76c169ac3142a0173ae06c3c4c90305ca700a9f`, so I do not see maintainer-run CI output yet.
- Author-reported validation includes `cargo fmt --check`, `cargo check -p proof-of-sql --no-default-features --features "std test"`, and the targeted `from_str_via_hash_matches_byte_hashing` test.

Suggested maintainer next step: approve/rerun CI for this head and compare it with the other #228 string-conversion candidates. If CI is clean, this looks like a narrow second-checkbox candidate; if anything needs follow-up, the likely area is broader VarChar proof-result decoding coverage around the new `decode_and_convert::<&str, &str>` path.

JayWhite2357:

PR #1911, opened 2026-05-21
test: add TableRef parser/serde coverage slice

liangtovi-debug:
/claim #560

## Summary
- add dedicated tests for `TableRef::new` with and without schema
- add parser helper coverage for `from_strs` and `TryFrom<&str>` success/error paths
- add serde round-trip and invalid input coverage for `TableRef`
- wire `table_ref_test` into `base/database/mod.rs`

## Validation
- attempted: `cargo test -p proof-of-sql table_ref_ -- --nocapture`
- blocked in this macOS environment by upstream build constraints: `blitzar-sys` requires Linux and `halo2curves` asm gate requires x86_64
- attempted fallback: `cargo test -p proof-of-sql --no-default-features table_ref_ -- --nocapture`
- fallback build also fails due unrelated upstream no-std test compile errors across many files

This PR is intentionally a small non-overlapping test-only coverage slice for bounty issue #560.

PR #1967, opened 2026-05-22
test: cover table reference parsing and serde

JacKane21:
/claim #560

## Scope

Adds a small, test-only coverage slice for `base::database::table_ref`.

Covered behavior:
- `TableRef::new` with and without schema names
- `TableRef::from_idents`
- one-part and two-part parsing via `TryFrom<&str>`, `FromStr`, and `from_strs`
- invalid references with too many components
- `Display` output
- serde string round-trip and invalid serde input

## Validation

- `cargo test -p proof-of-sql --no-default-features --features test table_ref::tests -- --nocapture` -> 4 passed
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/database/table_ref.rs
+++ b/crates/proof-of-sql/src/base/database/table_ref.rs
@@ -142,3 +142,84 @@ impl<'d> Deserialize<'d> for TableRef {
         TableRef::from_str(&string).map_err(serde::de::Error::custom)
     }
 }
+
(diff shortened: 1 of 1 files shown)


JacKane21:
Closing this because I found an earlier overlapping TableRef coverage slice (#1911) after opening. I do not want to add duplicate noise to the coverage bounty; I will take a genuinely non-overlapping slice instead.

PR #1970, opened 2026-05-22
test: cover permutation comparator and errors

JacKane21:
/claim #560

## Scope
Adds a small, test-only coverage slice for `base::math::permutation`.

Covered behavior:
- `unchecked_new_from_cmp` orders indices using the supplied comparator
- stable Display output for duplicate-element invalid permutation errors
- stable Display output for permutation size mismatch errors

No production behavior changes, aside from making the existing dead-code lint expectation apply only outside test builds now that the comparator constructor is directly covered.

## Validation
- `cargo test -p proof-of-sql --no-default-features --features test base::math::permutation::test -- --nocapture` -> 5 passed
- `cargo fmt --check`
- `git diff --check`


JacKane21:

PR #1971, opened 2026-05-22
test: cover varint decode edge cases

JacKane21:
/claim #560

## Scope
Adds a small, test-only coverage slice for `base::encode::varint_trait`.

Covered behavior:
- bool varint decoding rejects non-boolean values
- narrow integer decoding rejects overflow values
- `u64` decoding rejects incomplete and overwide varints
- `u128` decoding rejects encoded values that require the high `U256` limb
- large `u128`/`i128` values round-trip through varint encoding

## Validation
- `cargo test -p proof-of-sql --no-default-features --features test base::encode::varint_trait::tests -- --nocapture` -> 5 passed
- `cargo fmt --check`
- `git diff --check`


JacKane21:

PR #1977, opened 2026-05-22
test: cover decimal division edge cases

JacKane21:
## Summary
- add focused coverage for decimal division by zero returning `ColumnOperationError::DivisionByZero`
- add coverage for decimal division where the internal applied scale is negative

/claim #560

## Validation
- `cargo test -p proof-of-sql --no-default-features --features test slice_decimal_operation -- --nocapture`
- `cargo fmt --check`
- `git diff --check`


JacKane21:

PR #1983, opened 2026-05-22
test: cover timestamp scale casts

JacKane21:
## Summary
- Adds coverage for `scale_cast_binary_op` when the left timestamp needs to be cast to the right timestamp type.
- Adds coverage for the symmetric right-side timestamp cast branch.
- Keeps the change test-only and scoped to `sql::scale`.

/claim #560

## Validation
- `cargo test -p proof-of-sql --no-default-features --features test sql::scale::tests -- --nocapture` -> 7 passed
- `git diff --check`


JacKane21:

PR #1994, opened 2026-05-23
Add FixedSizeBinary support

SkyForThinking:
## Summary
- add `FixedSizeBinary` support across Arrow conversion, owned/borrowed columns, committable columns, query operations, and provable result handling
- make fixed-size binary commitment/inner-product paths use the fixed-size scalar embedding instead of varbinary hashing
- validate decoded fixed-size binary result lengths against the declared byte width

## Notes
- `ColumnType::FixedSizeBinary(_)::byte_size()` now reports the 32-byte scalar-limb representation used by commitments, matching `[u64; 4]` packing rather than the logical Arrow byte width
- `FixedSizeBinary` metadata/bounds are treated like other unordered binary/string-like columns

## Testing
- `cargo fmt --check`
- `cargo check -p proof-of-sql --lib --no-default-features`
- `cargo test -p proof-of-sql --lib --no-default-features --features arrow,test fixed_size_binary -- --nocapture`
- `cargo test -p proof-of-sql --lib --no-default-features --features arrow,test column_commitment_metadata -- --nocapture`
- `cargo test -p proof-of-sql --lib --no-default-features --features arrow,test base::arrow -- --nocapture`

Also attempted broader `cargo test -p proof-of-sql --lib --no-default-features --features arrow,test`; it progressed through the suite but timed out locally after 600s on macOS before a final summary.

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/arrow_array_to_column_conversion.rs
+++ b/crates/proof-of-sql/src/base/arrow/arrow_array_to_column_conversion.rs
@@ -7,9 +7,10 @@ use crate::base::{
 };
 use arrow::{
     array::{
-        Array, ArrayRef, BooleanArray, Decimal128Array, Decimal256Array, Int16Array, Int32Array,
-        Int64Array, Int8Array, LargeBinaryArray, StringArray, TimestampMicrosecondArray,
-        TimestampMillisecondArray, TimestampNanosecondArray, TimestampSecondArray, UInt8Array,
+        Array, ArrayRef, BooleanArray, Decimal128Array, Decimal256Array, FixedSizeBinaryArray,
+        Int16Array, Int32Array, Int64Array, Int8Array, LargeBinaryArray, StringArray,
+        TimestampMicrosecondArray, TimestampMillisecondArray, TimestampNanosecondArray,
+        TimestampSecondArray, UInt8Array,
     },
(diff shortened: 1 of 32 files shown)


SkyForThinking:
One implementation detail I wanted to call out: I intentionally kept FixedSizeBinary separate from VarBinary in the scalar embedding paths.

VarBinary uses the hash-based byte embedding, while FixedSizeBinary is treated more like fixed-width scalar-backed data elsewhere. Reusing the VarBinary path there would make the inner-product / commitment behavior inconsistent, so this PR uses the fixed-size byte embedding for FixedSizeBinary instead.

I also added width checks when decoding FixedSizeBinary query results, so FixedSizeBinary(n) fails if the decoded value is not exactly n bytes.

Local checks run:

- cargo fmt --check
- cargo check -p proof-of-sql --lib --no-default-features
- cargo test -p proof-of-sql --lib --no-default-features --features arrow,test fixed_size_binary -- --nocapture
- cargo test -p proof-of-sql --lib --no-default-features --features arrow,test column_commitment_metadata -- --nocapture
- cargo test -p proof-of-sql --lib --no-default-features --features arrow,test base::arrow -- --nocapture

I also tried the broader no-default-features arrow/test lib suite, but it hit my 600s local timeout on macOS before finishing.

JayWhite2357:

PR #2004, opened 2026-05-23
test: cover standard binary deserialization edges

JJ-Lin:
## Summary
- add coverage that standard binary deserialization reports the consumed byte count when trailing data is present
- add coverage that truncated serialized payloads return a decode error

/claim #560

## Test plan
- cargo fmt --check
- cargo test -p proof-of-sql --no-default-features --features test standard_serializations::binary::tests --lib
- git diff --check

Note: running the focused test without `--features test` fails in this crate because the lib test target compiles no-std test modules that require the repository test feature.

### DIFF
--- a/crates/proof-of-sql/src/base/standard_serializations/binary.rs
+++ b/crates/proof-of-sql/src/base/standard_serializations/binary.rs
@@ -51,4 +51,37 @@ mod tests {
         let reserialized = try_standard_binary_serialization(deserialized).unwrap();
         assert_eq!(serialized, reserialized);
     }
+
+    #[test]
+    fn deserialization_reports_consumed_bytes_with_trailing_data() {
+        let obj = SerdeTestType {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

- Scope is test-only in `standard_serializations/binary.rs` (`+33/-0`), adding focused deserialization edge coverage without production-path changes.
- The two useful assertions are: trailing bytes preserve a successful decode while reporting the consumed byte count, and a one-byte-truncated payload returns an error.
- PR state currently looks mergeable but still approval-gated: `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`; visible Orca checks are green and I do not see public Rust Actions check-runs on the PR.
- The PR body includes the issue 560 claim marker; this comment is only review context, not an additional claim submission.

JayWhite2357:

PR #2006, opened 2026-05-23
test: run proof expr test without blitzar

Yassinbrine:
/claim #560

## Summary
- enable the existing `proof_expr_test` module for plain `#[cfg(test)]`
- switch that test from `InnerProductProof` to the test-only `NaiveEvaluationProof`
- keep the coverage slice independent of the `blitzar` feature so it can run under the no-default `test` feature path

## Validation
- `git diff --check` passed
- Not run: `cargo test -p proof-of-sql --lib --no-default-features --features test we_can_compute_the_correct_result_of_a_complex_bool_expr_using_first_round_evaluate` because `cargo` is not installed in this local environment

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/mod.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/mod.rs
@@ -2,7 +2,7 @@
 mod proof_expr;
 pub(crate) use proof_expr::DecimalProofExpr;
 pub use proof_expr::ProofExpr;
-#[cfg(all(test, feature = "blitzar"))]
+#[cfg(test)]
 mod proof_expr_test;
 
 mod aliased_dyn_proof_expr;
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:

- Scope is a small test-path enablement (`+3/-3` across 2 files): `proof_expr_test` moves from `#[cfg(all(test, feature = "blitzar"))]` to plain `#[cfg(test)]`, and the test setup swaps `InnerProductProof` for test-only `NaiveEvaluationProof`.
- The useful coverage effect is that the existing complex boolean proof-expression test can run under the no-default `test` feature path instead of requiring `blitzar`.
- Review point: this appears to avoid production runtime behavior changes, but maintainers may want to confirm `NaiveEvaluationProof` is the intended test-only replacement for this coverage lane.
- PR state is approval-gated from the public view: `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`; raw check-runs show only the three Orca checks and all are green.
- The PR body includes `/claim #560`; this comment is review context only, not an additional claim submission.

PR #2010, opened 2026-05-23
Fix HyperKZG public setup short-read deserialization

JJ-Lin:
/claim #560

## Summary
- make `deserialize_flat_compressed_hyperkzg_public_setup_from_reader` accumulate bytes until a full compressed point is available
- return `UnexpectedEof` for truncated setup data instead of attempting to deserialize a partial point buffer
- add regression coverage for reader implementations that return short reads

## Tests
- `cargo test -p proof-of-sql --no-default-features --features std we_can_deserialize_ppot_02_setup_from_reader_with_short_reads -- --nocapture`
- `cargo test -p proof-of-sql --no-default-features --features std hyperkzg::public_setup -- --nocapture`
- `cargo fmt --check --all`

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/hyperkzg/public_setup.rs
+++ b/crates/proof-of-sql/src/proof_primitive/hyperkzg/public_setup.rs
@@ -24,21 +24,31 @@ pub fn deserialize_flat_compressed_hyperkzg_public_setup_from_reader<R: std::io:
     mut reader: R,
     validate: Validate,
 ) -> Result<HyperKZGPublicSetupOwned, SerializationError> {
-    std::iter::repeat_with(|| {
+    let mut setup = Vec::new();
+    loop {
         let mut buffer = [0; COMPRESSED_SIZE];
-        let num_bytes_read = reader.read(&mut buffer);
-        (buffer, num_bytes_read)
-    })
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a concrete `/claim #560` HyperKZG deserialization fix, not just coverage: `deserialize_flat_compressed_hyperkzg_public_setup_from_reader` now keeps reading until each `COMPRESSED_SIZE` point buffer is full, handles `Interrupted`, and returns `UnexpectedEof` instead of trying to deserialize a partially filled point. The added test uses a `Read` implementation capped at 7 bytes/read and compares the reader path with the existing slice deserializer, so it specifically covers short-read behavior that ordinary byte-slice tests would miss.

I checked nearby public-setup PRs and this looks distinct from the slice/error coverage lanes (`#1625`, `#2383`): this one changes the streaming reader contract. The main merge-readiness item I would still confirm is the body's truncated-data claim: the patch implements `UnexpectedEof` on a partial final point, but the new regression test only covers successful short reads. A small test for truncated reader input would pin that error path before merge.

No local Rust suite was run from my side. Visible Orca checks are green; combined statuses are empty/pending and this PR is still gated on `reviewDecision=REVIEW_REQUIRED` / `mergeStateStatus=BLOCKED`.

JayWhite2357:

PR #2022, opened 2026-05-24
test: cover commitment evaluation proof forwarding

michaelapollopimentel-svg:
/claim #560

## Summary
- add a focused unit test for the default `CommitmentEvaluationProof::verify_proof()` helper
- verify the helper forwards a single commitment, product, and batching factor into `verify_batched_proof()` exactly as documented
- keep the change fully test-only

## Testing
- `cargo test -p proof-of-sql --lib --no-default-features --features test base::commitment::commitment_evaluation_proof::tests::verify_proof_forwards_single_inputs_to_verify_batched_proof -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/commitment_evaluation_proof.rs
+++ b/crates/proof-of-sql/src/base/commitment/commitment_evaluation_proof.rs
@@ -71,3 +71,75 @@ pub trait CommitmentEvaluationProof {
         setup: &Self::VerifierPublicSetup<'_>,
     ) -> Result<(), Self::Error>;
 }
+
+#[cfg(test)]
+mod tests {
+    use super::CommitmentEvaluationProof;
+    use crate::base::{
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only `/claim #560` change in `crates/proof-of-sql/src/base/commitment/commitment_evaluation_proof.rs` (`+72/-0`). It adds a local `ForwardingProof` implementation to check that the default `verify_proof()` helper forwards the single commitment, `ONE` batching factor, evaluation, `b_point`, generator offset, and table length into `verify_batched_proof()`.

The main review point I would confirm is whether this should also assert transcript behavior. The test is good at pinning argument forwarding, but because `ForwardingProof::verify_batched_proof` ignores the transcript, it does not detect accidental transcript mutation/order changes inside the helper path.

JayWhite2357:

PR #2025, opened 2026-05-24
test: cover bit matrix ordering

michaelapollopimentel-svg:
/claim #560

## Summary
- add coverage for the ordering of multiple varying bits across more than two rows
- add coverage for cross-limb varying bits so the output columns stay in bit-position order

## Testing
- cargo test -p proof-of-sql --lib --no-default-features --features test bit_matrix_test -- --nocapture
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/proof-of-sql/src/base/bit/bit_matrix_test.rs
+++ b/crates/proof-of-sql/src/base/bit/bit_matrix_test.rs
@@ -74,6 +74,45 @@ fn we_can_compute_the_bit_matrix_for_data_with_varying_bits_and_constant_bits() 
 }
 
 #[test]
+fn we_preserve_varying_bit_order_across_multiple_rows() {
+    let data: Vec<TestScalar> = vec![
+        TestScalar::zero(),
+        TestScalar::one(),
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer review assist for `/claim #560`:

- Scope is a single test-only file, `crates/proof-of-sql/src/base/bit/bit_matrix_test.rs`, with two focused cases and no production-path changes.
- The added cases exercise two ordering invariants that are easy to regress: preserving varying-bit column order across more than two rows, and preserving cross-limb ordering when a high-limb bit varies alongside a low bit.
- Current PR metadata looks merge-friendly from the contributor side: open, non-draft, mergeable, no existing comments/reviews/review threads before this note, and Orca IaC/Secrets/Vulnerabilities checks are green. The remaining gate appears to be maintainer review / Actions approval rather than a visible test failure.

This looks like a low-risk coverage slice that complements the existing bit matrix tests and should be suitable for maintainer review/rerun when queue capacity allows.

JayWhite2357:

PR #2026, opened 2026-05-24
test: optimize test profile for faster CI runtimes

liangtovi-debug:
/claim #557

## Summary
- add `[profile.test]` with `opt-level = 3` at the workspace root
- keep all test logic and assertions unchanged

## Why
Issue #557 explicitly allows CI/runtime improvements without weakening tests. This change targets compute-heavy cryptographic tests by compiling tests with optimization while preserving debuginfo and test behavior.

## Validation
- `cargo test -p proof-of-sql --no-run --no-default-features --features std`

## Notes
- This is intentionally small and low-risk so maintainers can quickly compare CI timing deltas against baseline.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,3 +89,5 @@ allow_attributes = "warn"
 # patch for sqlparser no_std compatibility with the serde feature enabled.
 # required until proof-of-sql upgrades to at least sqlparser 0.56.0.
 sqlparser = { git = "https://github.com/tlovell-sxt/datafusion-sqlparser-rs.git", rev = "a828cbea22cf19bb6b4596f902bdd6f4d14a00b8" }
+[profile.test]
+opt-level = 3


MyTH-zyxeon:
Review-assist for #557 triage:

I checked this PR against the current body, diff, visible checks, comments, reviews, and review-thread state. It is a very small `Cargo.toml`-only runtime slice: it keeps the `sqlparser` patch unchanged, restores the trailing newline, and adds workspace `[profile.test] opt-level = 3` so compute-heavy tests can run with optimized test binaries while leaving test assertions and source logic untouched.

Current visible state from my side: no prior comments, no reviews, no review threads; Orca Security checks are green. The GitHub Actions runs for this head SHA show `CI-Code-Coverage` and `CI-Lint-And-Test` as `action_required`, so this still looks like the fork-workflow approval gate rather than an observed code failure.

Acceptance note: because this changes the workspace test profile globally, the useful maintainer check is the timing/behavior comparison already implied by the PR body's `cargo test -p proof-of-sql --no-run --no-default-features --features std`, plus one normal CI run after approval. If CI approves the runtime improvement, this is distinct from the Dory setup-size PRs because it changes compiler optimization for tests rather than weakening or deleting any test cases.

JayWhite2357:

PR #2031, opened 2026-05-24
test: cover record batch conversion error paths

michaelapollopimentel-svg:
/claim #560

Adds focused coverage for record-batch conversion failure paths without changing production logic.

Scope:
- cover null-containing Arrow batches bubbling through `batch_to_columns`
- cover identifier mismatch errors when appending record batches to an existing table commitment
- cover column-count mismatch errors for record-batch append flows

Validation:
- `cargo test -p proof-of-sql record_batch_conversion --no-default-features --features "arrow test"`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/record_batch_conversion.rs
+++ b/crates/proof-of-sql/src/base/arrow/record_batch_conversion.rs
@@ -177,3 +177,108 @@ mod tests {
         assert_eq!(commitment, expected_commitment);
     }
 }
+
+#[cfg(test)]
+mod error_tests {
+    use super::*;
+    use crate::base::{
+        arrow::arrow_array_to_column_conversion::ArrowArrayToColumnConversionError,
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only `/claim #560` change in `crates/proof-of-sql/src/base/arrow/record_batch_conversion.rs` (`+105/-0`). It adds coverage for three record-batch conversion failure paths: null-containing Arrow arrays bubbling out of `batch_to_columns`, identifier mismatches while appending a record batch to an existing table commitment, and column-count mismatches in the same append flow.

The main review point I would confirm is whether these tests should also assert the mismatched identifier details, not just the `ColumnCommitmentsMismatch::Ident { .. }` shape. The current assertions are useful for preserving error classification without over-pinning wording, but if downstream users rely on which identifiers are reported, that payload may be worth locking down too.

JayWhite2357:

PR #2041, opened 2026-05-26
test: cover column type operation guards

meteop-x:
/claim #560

## Summary
- add focused tests for equality and inequality column type guard behavior
- cover scaled comparison guard behavior across string, binary, boolean, numeric, decimal, scalar, and timestamp cases
- cover logical `and`/`or` and `not` type guard helpers

## Validation
- `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=gcc RUSTFLAGS="" cargo test -p proof-of-sql --lib --no-default-features --features test column_type_operation -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/database/column_type_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_type_operation.rs
@@ -1401,4 +1401,148 @@ mod test {
             matches!(try_neg_type(ColumnType::VarChar).unwrap_err(), ColumnOperationError::UnaryOperationInvalidColumnType { operator, operand_type } if operator == "negation" && operand_type == ColumnType::VarChar)
         );
     }
+
+    #[test]
+    fn we_can_validate_equality_column_types() {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer review assist for `/claim #560`:

- Scope is a single test-only file, `crates/proof-of-sql/src/base/database/column_type_operation.rs`, with no production-path changes.
- The added tests cover equality/inequality guard behavior, scaled comparison guards across string/binary/boolean/numeric/decimal/scalar/timestamp cases, and logical `and`/`or`/`not` helpers.
- Current PR metadata looks review-ready from the contributor side: open, non-draft, mergeable, zero comments/reviews/review threads before this note, and Orca IaC/Secrets/Vulnerabilities checks are green. The remaining gate appears to be maintainer review / Actions approval rather than a visible code failure.

This is a larger single-file coverage slice than the tiny helper tests, but it stays focused on column type operation guard behavior and should be useful for maintainer review/rerun when queue capacity allows.

JayWhite2357:

PR #2042, opened 2026-05-26
test: cover U256 scalar varint boundaries

CharlieLZ:
/claim #560

## Summary
- add direct U256 varint round-trip coverage across low/high word boundaries
- cover the `read_u256_varint` shift-126 split path
- assert `u256_varint_size` matches the byte count written by `write_u256_varint`
- verify exact boundary bytes and non-terminating U256 varint rejection

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features test scalar_varint -- --nocapture` (19 passed)
- `git diff --check`

Earlier branch validation before this follow-up also ran `cargo test -p proof-of-sql --lib --no-default-features --features test` (962 passed).

### DIFF
--- a/crates/proof-of-sql/src/base/encode/scalar_varint_test.rs
+++ b/crates/proof-of-sql/src/base/encode/scalar_varint_test.rs
@@ -1,6 +1,7 @@
 use super::scalar_varint::{
-    read_scalar_varint, read_scalar_varints, scalar_varint_size, scalar_varints_size,
-    write_scalar_varint, write_scalar_varints,
+    read_scalar_varint, read_scalar_varints, read_u256_varint, scalar_varint_size,
+    scalar_varints_size, u256_varint_size, write_scalar_varint, write_scalar_varints,
+    write_u256_varint,
 };
 use crate::base::{encode::U256, scalar::test_scalar::TestScalar};
 use alloc::vec;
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

I checked this as a focused test-only candidate. The change stays in `crates/proof-of-sql/src/base/encode/scalar_varint_test.rs` and adds direct coverage for `read_u256_varint`, `write_u256_varint`, and `u256_varint_size`.

- Boundary coverage spans 1-byte, 2-byte, 18-byte, 19-byte, and 37-byte encodings, including both low-word and high-word transitions.
- `read_u256_varint_handles_the_shift_126_low_high_split` exercises the low/high split around bit 126, which is the easiest place for a varint implementation to lose bits.
- The added tests also assert exact boundary marker bytes for `1 << 126` and high-word `1`, plus the non-terminating `0xff` buffer rejection path.
- Scope is narrow: one test file, `+69/-2`, no production behavior changes.

Deconfliction notes: before posting this, the PR had zero top-level comments and zero reviews. Searches for `read_u256_varint shift 126`, `non-terminating U256 varint`, and `U256 scalar varint boundaries` only surfaced this PR. Current PR state is `OPEN` / `BLOCKED` / `REVIEW_REQUIRED` at head `f3d467ed64fa3c3563b10ba0ecd35f1b6d710848`; Orca checks are green. I did not run local cargo because the `/tmp/sxt-proof-of-sql` checkout is not present in this environment.

PR #2043, opened 2026-05-26
test: reduce repeated Dory public parameter generation

CharlieLZ:

/claim #557

## Summary
- generate one max-size Dory `PublicParameters` value for the setup size tests
- derive the `nu=0..4` test inputs from prefixes of that value instead of calling `PublicParameters::test_rand` for every size
- keep the existing prover setup, verifier setup, and verifier setup serialization assertions over the same size range

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features test setup_test -- --nocapture`
- `cargo test -p proof-of-sql --lib --no-default-features --features test`
- `git diff --check`

Note: `cargo nextest` is not installed in my local environment, so I could not run the exact nextest command from the issue.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/setup_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/setup_test.rs
@@ -2,6 +2,32 @@ use super::{test_rng, ProverSetup, PublicParameters, VerifierSetup};
 use ark_ec::pairing::Pairing;
 use std::{fs, path::Path};
 
+const SETUP_TEST_MAX_NU: usize = 4;
+
+fn public_parameters_by_nu() -> Vec<PublicParameters> {
+    let mut rng = test_rng();
+    let public_parameters = PublicParameters::test_rand(SETUP_TEST_MAX_NU, &mut rng);
+    // Reuse prefixes from one max-size setup so each test still covers every nu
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer-facing review assist for `/claim #557`:

- Scope is focused on one performance-sensitive test file: `crates/proof-of-sql/src/proof_primitive/dory/setup_test.rs` (`+32/-12`).
- The change generates one max-size `PublicParameters` value and derives `nu=0..4` prefixes from it, avoiding repeated `PublicParameters::test_rand` calls across the prover setup, verifier setup, and verifier setup serialization tests.
- The existing assertions over the same size range remain in place; the helper changes only how the test fixtures are produced.
- The PR has no prior conversation comments or latest reviews visible before this note.
- PR state at review time: `OPEN`, `MERGEABLE`, `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`.
- Orca IaC/Secrets/Vulnerabilities checks are green. No GitHub Actions runs were visible for head `6c44ac47a61ab9893569ef4218b72de911627906`, and combined status returned `statuses=[]`, so there is no failing CI log to diagnose yet.

I do not see a code-change request here; this looks like a narrow `#557` performance candidate awaiting maintainer review and CI/Actions visibility.

PR #2049, opened 2026-05-26
test: cover byte distribution high-byte variance

glasses666:
/claim #560

## Summary
- Add focused test coverage for `ByteDistribution::new` when the highest byte column (index 248) varies.
- Verify the top-byte vary-mask bit, reported varying byte indices, and preservation of a constant low byte.
- Keep production logic unchanged.

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features="test" base::byte::byte_distribution -- --nocapture`
  - 10 passed; 951 filtered
- `git diff --check`

Bounty: #560

### DIFF
--- a/crates/proof-of-sql/src/base/byte/byte_distribution.rs
+++ b/crates/proof-of-sql/src/base/byte/byte_distribution.rs
@@ -201,6 +201,24 @@ mod tests {
     }
 
     #[test]
+    fn we_can_get_byte_distribution_when_highest_byte_varies() {
+        let constant_low_byte = U256::from(0x44u8);
+        let column = [
+            constant_low_byte | U256::from(0x11u8).shl(8) | U256::from(0x01u8).shl(248),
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only change in `crates/proof-of-sql/src/base/byte/byte_distribution.rs` (`+18/-0`). It adds a boundary case where byte index 1 and the highest byte lane at index 31 vary while the low byte remains constant, then asserts the `vary_mask`, `constant_mask()`, `varying_byte_count()`, and byte index iterator output.

The main review point I would confirm is that the expected iterator indices `[8, 248]` intentionally expose bit offsets rather than byte-lane numbers. That matches the existing helper style if consumers treat those values as shifted bit positions, but this test will make that contract explicit at the high-byte boundary.

I did not run the Rust suite locally in this pass. The PR body records targeted cargo/fmt/diff checks by the contributor. Visible Orca checks are green; the two GitHub Actions runs for this head currently show `action_required`, and the combined commit status API returned `state=pending` with no statuses.

PR #2054, opened 2026-05-27
test: cover scaling cast expression helpers

thefuturrjfhejh:
/claim #560

## Summary
- Add no-default unit coverage for `ScalingCastExpr` constructor/accessor behavior.
- Cover invalid target-type rejection through the `DataTypeMismatch` path.
- Cover column-reference delegation from a wrapped column expression.

## Validation
- `cargo +1.91.1-x86_64-pc-windows-gnu test -p proof-of-sql --lib --no-default-features --features test sql::proof_exprs::scaling_cast_expr::tests -- --nocapture` (3 passed)
- `git diff --check`
- `bash -lc "tr -d ''\r'' < scripts/check_commits.sh | bash"`

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/scaling_cast_expr.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/scaling_cast_expr.rs
@@ -110,3 +110,71 @@ impl ProofExpr for ScalingCastExpr {
         self.from_expr.get_column_references(columns);
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{DynProofExpr, ProofExpr, ScalingCastExpr};
+    use crate::{
+        base::{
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only `/claim #560` change in `crates/proof-of-sql/src/sql/proof_exprs/scaling_cast_expr.rs` (`+68/-0`). It adds coverage for `ScalingCastExpr::try_new` constructor/accessor behavior, rejection of a non-scalable Boolean target through `AnalyzeError::DataTypeMismatch`, and column-reference delegation from a wrapped `ColumnExpr`.

The main review point I would confirm is whether the `DataTypeMismatch` assertion should also pin the higher-level analyze context that produces this expression, not only the direct constructor error. The current tests are useful for the helper boundary, but a planner/analyzer-level fixture would catch accidental bypasses around `try_new`.

PR #2057, opened 2026-05-27
test: cover planner filter pushdown support

bilbop1:
## Summary
- add focused coverage for `PoSqlTableSource::supports_filters_pushdown`
- verify both the empty-filter result and multiple-filter `Exact` pushdown responses

## Tests
- `cargo test -p proof-of-sql-planner --no-default-features context::tests::we_can_report_exact_filter_pushdown_support -- --exact`
- `git diff --check`

/claim #560

### DIFF
--- a/crates/proof-of-sql-planner/src/context.rs
+++ b/crates/proof-of-sql-planner/src/context.rs
@@ -127,6 +127,7 @@ mod tests {
     use ahash::AHasher;
     use alloc::vec;
     use core::any::TypeId;
+    use datafusion::common::ScalarValue;
     use indexmap::indexmap_with_default;
     use proof_of_sql::base::database::{ColumnType, SchemaAccessorImpl, TableRef};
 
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This looks like a focused #560 coverage slice. I checked the current diff and it is test-only in `crates/proof-of-sql-planner/src/context.rs` (+22/-0), adding coverage for:

- `PoSqlTableSource::supports_filters_pushdown` with an empty filter list returning an empty response.
- multiple boolean filter expressions each reporting `TableProviderFilterPushDown::Exact`.

That makes it complementary to the other open #560 coverage PRs because it targets DataFusion planner table-source filter-pushdown behavior rather than `df_util` helpers, proof-plan slices like EmptyExec/DemoMockPlan, verification-builder accounting, fold utilities, bit-mask boundaries, table helpers, commitment-vector gating, equality helpers, or nullable/Arrow paths.

Current gate note: the PR is open/mergeable but still blocked on review/workflow approval. The head SHA `50164bedb9655c99a8a60fc55b80f0e2d650c36e` has `CI-Lint-And-Test` and `CI-Code-Coverage` at `action_required`, so I do not see a code-test failure signal from the available checks yet. The listed targeted `supports_filters_pushdown` cargo command is the useful validation once workflow approval is available.

PR #2059, opened 2026-05-27
test: cover big decimal bigint conversion

giaphutran12:
/claim #560

## Summary

- Adds focused tests for `BigDecimalExt::try_into_bigint_with_precision_and_scale`
- Covers the successful conversion path for matching precision and scale
- Covers the `LossyCast` path when scaled digits exceed the allowed precision
- Adds parser coverage for ignoring non-`CREATE TABLE` statements and decimals at the precision threshold

## Tests

- `cargo test -p proof-of-sql --lib --no-default-features --features test big_decimal_ext -- --nocapture`
- `cargo test -p proof-of-sql --lib --no-default-features --features test find_bigdecimals -- --nocapture`

Note: `npm run build` was attempted, but this repo has no `build` script in `package.json`.

### DIFF
--- a/crates/proof-of-sql/src/base/math/big_decimal_ext.rs
+++ b/crates/proof-of-sql/src/base/math/big_decimal_ext.rs
@@ -89,4 +89,24 @@ mod tests {
             .unwrap_err();
         assert!(matches!(err, IntermediateDecimalError::ConversionFailure));
     }
+
+    #[test]
+    fn we_can_convert_decimal_to_bigint_with_matching_precision_and_scale() {
+        let big_decimal: BigDecimal = "123.45".parse::<BigDecimal>().unwrap().normalized();
+        let bigint = big_decimal
+            .try_into_bigint_with_precision_and_scale(5, 2)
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:

- Scope is test-only (`+38/-0` across 2 files), focused on decimal conversion and parser coverage; I do not see production-path changes in the diff.
- The `BigDecimalExt::try_into_bigint_with_precision_and_scale` additions cover the successful `123.45 -> 12345` conversion when precision/scale match and the `IntermediateDecimalError::LossyCast` path when scaled digits exceed precision.
- The `find_bigdecimals` parser test checks that a leading non-`CREATE TABLE` statement is ignored and that only a DECIMAL above the precision threshold is reported.
- PR state is approval-gated from the public view: `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`; raw check-runs show only the three Orca checks and all are green.
- The PR body includes `/claim #560`; this comment is review context only, not an additional claim submission.

PR #1337, opened 2026-05-11
test: cover database error displays

vegapunk9527:
## Summary

- add focused display tests for database column operation errors
- add display tests for table operation, owned column, coercion, and parse errors
- keep the change scoped to non-overlapping `base/database/*_error.rs` coverage

/claim #560

## Testing

- `cargo test -p proof-of-sql --no-default-features --features test we_can_display --lib` (passes: 8 tests)
- `rustfmt --check crates/proof-of-sql/src/base/database/column_operation_error.rs crates/proof-of-sql/src/base/database/table_operation_error.rs crates/proof-of-sql/src/base/database/owned_column_error.rs crates/proof-of-sql/src/base/database/error.rs`
- `git diff --check`

Note: `cargo test -p proof-of-sql we_can_display --lib` with default features is blocked on this Mac arm64 environment by upstream platform constraints (`blitzar-sys` only supports Linux and `halo2curves/asm` is x86_64-only).

PR #2062, opened 2026-05-27
test: cover database parse error display

adliebe:
/claim #560

Adds a focused test-only coverage slice for `ParseError::InvalidTableReference` in `crates/proof-of-sql/src/base/database/error.rs`.

Scope:
- covers the display output for invalid table-reference parse errors
- verifies equality remains keyed by the rejected reference text
- keeps production code unchanged
- avoids the current `table_ref.rs` coverage PRs and other active #560 file slices

Validation:
- `CARGO_TARGET_DIR=/tmp/sxt-proof-of-sql-target CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=cc cargo test -p proof-of-sql --lib --no-default-features --features test base::database::error::tests -- --nocapture`
- `git diff --check`
- `tr -d '\r' < scripts/check_commits.sh | bash`

### DIFF
--- a/crates/proof-of-sql/src/base/database/error.rs
+++ b/crates/proof-of-sql/src/base/database/error.rs
@@ -11,3 +11,37 @@ pub enum ParseError {
         table_reference: String,
     },
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloc::string::ToString;
+
+    #[test]
+    fn invalid_table_reference_display_includes_the_rejected_reference() {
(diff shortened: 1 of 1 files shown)


adliebe:
Closing after deeper #560 collision review: older open PR #1337 already touches crates/proof-of-sql/src/base/database/error.rs for database error display coverage. This PR is test-only and harmless, but likely overlaps prior bounty accounting, so I am withdrawing it rather than adding duplicate claim noise.

PR #2082, opened 2026-05-28
Add FixedSizeBinary column support

### LINKED ISSUE #229: Add `FixedSizeBinary` with support up to at least 32 bytes.

maiqiu-cat:
Closes #229.

## Summary
- Add `ColumnType::FixedSizeBinary(i32)` and preserve the fixed width through `Column`, `OwnedColumn`, `CommittableColumn`, Arrow conversion, serialization, indexing, filtering, union, repetition, ordering, commitments, and result decoding paths.
- Add scalar conversion for fixed-size binary values: <=31 bytes are naturally embedded, 32 bytes are hashed, and larger widths are rejected.
- Support equality comparisons for fixed-size binary columns while keeping ordering comparisons rejected.

## Tests
- `cargo fmt --all`
- `cargo check -p proof-of-sql --lib --no-default-features --features arrow,test`
- `cargo test -p proof-of-sql --no-default-features --features arrow,test fixed_size_binary`
- `cargo test -p proof-of-sql --no-default-features --features arrow,test base::arrow::`

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/arrow_array_to_column_conversion.rs
+++ b/crates/proof-of-sql/src/base/arrow/arrow_array_to_column_conversion.rs
@@ -7,9 +7,10 @@ use crate::base::{
 };
 use arrow::{
     array::{
-        Array, ArrayRef, BooleanArray, Decimal128Array, Decimal256Array, Int16Array, Int32Array,
-        Int64Array, Int8Array, LargeBinaryArray, StringArray, TimestampMicrosecondArray,
-        TimestampMillisecondArray, TimestampNanosecondArray, TimestampSecondArray, UInt8Array,
+        Array, ArrayRef, BooleanArray, Decimal128Array, Decimal256Array, FixedSizeBinaryArray,
+        Int16Array, Int32Array, Int64Array, Int8Array, LargeBinaryArray, StringArray,
+        TimestampMicrosecondArray, TimestampMillisecondArray, TimestampNanosecondArray,
+        TimestampSecondArray, UInt8Array,
     },
     datatypes::{i256, DataType, TimeUnit as ArrowTimeUnit},
 };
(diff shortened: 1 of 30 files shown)


JayWhite2357:

PR #1966, opened 2026-05-22
test: cache Dory setups in evaluation proof tests

nozprod:
Please be sure to look over the pull request guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md#submit-pr.

/claim #557

# Please go through the following checklist

- [ ] The PR title and commit messages adhere to guidelines here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/CONTRIBUTING.md. In particular `!` is used if and only if at least one breaking change has been introduced.
- [ ] I have run the ci check script with `source scripts/run_ci_checks.sh`.
- [ ] I have run the clean commit check script with `source scripts/check_commits.sh`, and the commit history is certified to follow clean commit guidelines as described here: https://github.com/spaceandtimelabs/sxt-proof-of-sql/blob/main/COMMIT_GUIDELINES.md
- [x] The latest changes from `main` have been incorporated to this PR by simple rebase if possible, if not, then conflicts are resolved appropriately.

# Rationale for this change

Addresses #557.

The Dory commitment evaluation proof tests repeatedly construct equivalent public parameters, prover setups, and verifier setups for the same `nu` values. `VerifierSetup::from` performs expensive pairing work, and `ProverSetup::from` may also prepare setup-specific state, so rebuilding these values in each test adds avoidable runtime.

# What changes are included in this PR?

- Add a test-only `test_setup.rs` helper with `OnceLock`-backed cached Dory setups for the `nu` values used by these tests.
- Wire the helper into the Dory test module.
- Update the static and dynamic Dory commitment evaluation proof tests to reuse cached prover/verifier setup values.

# Are these changes tested?

Partially. `git diff --check` passes locally.

I could not run `cargo test`, `source scripts/run_ci_checks.sh`, or `source scripts/check_commits.sh` in this Codex runtime because `cargo`/`rustfmt`/Rust toolchain managers are not installed on PATH here.

PR #1771, opened 2026-05-19
test: shrink Dory state setup without dropping cases

MyTH-zyxeon:
## Summary
- reduce over-provisioned Dory public parameter setup generation in the state and extended-state multi-nu tests from max_nu=5 to max_nu=4
- preserve the exact tested nu values by changing the loops from 0..max_nu to 0..=max_nu
- keep all assertions and verifier-state checks unchanged

/claim #557

## Validation
- cargo fmt -- --check
- cargo test -p proof-of-sql --no-default-features --features="test" proof_primitive::dory::state_test
- cargo test -p proof-of-sql --no-default-features --features="test" proof_primitive::dory::extended_state_test
- git diff --check

PR #2083, opened 2026-05-28
perf(test): cache deterministic Dory PublicParameters across the test suite

sravan27:
## Summary

The Dory test suite calls `PublicParameters::test_rand(N, &mut test_rng())` **95 times**, and **51 of those** use a freshly-created `test_rng()` (`ark_std::test_rng()`) — which is a *fixed-seed deterministic* RNG. Every one of those 51 calls produces a bit-identical `PublicParameters` for the same `max_nu`, so today's tests are paying the full EC-point-generation cost dozens of times for the same handful of values (in particular, `nu = 5` is used **40 times**, `nu = 4` is used 7 times).

This PR introduces a process-shared cache for `PublicParameters` keyed by `max_nu`, and migrates the safe call sites to use it.

## Why this targets the bounty

The issue (#557) identifies `PublicParameters::test_rand`, `ProverSetup::from`, and `VerifierSetup::from` as the dominant cost (`>50%`). Of those, `PublicParameters::test_rand` is the one that can be shared across threads/processes safely (no `blitzar` handle, no thread-affinity), so it's the right first target — and on its own it collapses ~51 expensive constructions per run into ≤ ~5 per process (one per distinct `max_nu`).

## Changes

- **`dory::rand_util`**: adds `shared_test_public_parameters(max_nu) -> &'static PublicParameters` backed by `OnceLock<Mutex<HashMap<usize, &'static PublicParameters>>>` (test-only; leaked once per `max_nu`, once per process). The function is documented and guaranteed bit-identical to `PublicParameters::test_rand(max_nu, &mut test_rng())`.
- **`dory::mod`**: re-exports `shared_test_public_parameters` at the dory module level alongside `test_rng`.
- **51 test call sites** across 8 files migrated to the cached helper. Only `PublicParameters::test_rand(_, &mut test_rng())` direct calls were migrated; the 44 `&mut rng` call sites — where `rng` may carry advanced state from prior use — are deliberately left untouched to preserve their RNG-state semantics.

## Conservative-by-design

- Cache is only consulted via `shared_test_public_parameters`; `PublicParameters::test_rand` is unchanged.
- `ProverSetup` and `VerifierSetup` are *not* cached (the `blitzar` `MsmHandle` field has thread-affinity considerations; a follow-up can extend the same pattern to those once a Send/Sync story is in place).
- Cache is `#[cfg(test)]`-gated and uses `std`, which is already available in the test cfg.

## Verification

- `cargo test -p proof-of-sql --no-default-features --features="arrow rayon" --lib "proof_primitive::dory::"` → **171 passed; 0 failed**.
- `cargo fmt -- --check` clean on all changed files.
- Clippy clean on the changed code (the helper carries a `# Panics` doc for lock-poisoning).

## Follow-ups (separate PRs welcome)

- Extend the same pattern to `ProverSetup`/`VerifierSetup` once thread-safety of the optional `blitzar` field is settled.
- Migrate the remaining 44 `&mut rng` call sites where a quick local audit confirms the rng wasn't advanced before the `test_rand` call.

Refs: #557

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/query_commitments.rs
+++ b/crates/proof-of-sql/src/base/commitment/query_commitments.rs
@@ -136,7 +136,7 @@ mod tests {
         },
         proof_primitive::dory::{
             test_rng, DoryCommitment, DoryEvaluationProof, DoryProverPublicSetup, ProverSetup,
-            PublicParameters,
+            PublicParameters, shared_test_public_parameters,
         },
     };
 
@@ -319,8 +319,8 @@ mod tests {
     #[expect(clippy::similar_names)]
     #[test]
     fn we_can_get_query_commitments_from_accessor() {
-        let public_parameters = PublicParameters::test_rand(4, &mut test_rng());
-        let prover_setup = ProverSetup::from(&public_parameters);
+        let public_parameters = shared_test_public_parameters(4);
+        let prover_setup = ProverSetup::from(public_parameters);
         let setup = DoryProverPublicSetup::new(&prover_setup, 3);
 
         let column_a_id: Ident = "column_a".into();
--- a/crates/proof-of-sql/src/proof_primitive/dory/dory_commitment.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dory_commitment.rs
@@ -103,7 +103,7 @@ mod tests {
             database::{Column, OwnedColumn},
             scalar::test_scalar_constants,
         },
-        proof_primitive::dory::{rand_util::test_rng, ProverSetup, PublicParameters},
+        proof_primitive::dory::{ProverSetup, shared_test_public_parameters},
     };
     use ark_ec::pairing::Pairing;
     use ark_ff::UniformRand;
@@ -116,8 +116,8 @@ mod tests {
 
     #[test]
     fn we_can_convert_from_columns() {
-        let public_parameters = PublicParameters::test_rand(5, &mut test_rng());
-        let prover_setup = ProverSetup::from(&public_parameters);
+        let public_parameters = shared_test_public_parameters(5);
+        let prover_setup = ProverSetup::from(public_parameters);
         let setup = DoryProverPublicSetup::new(&prover_setup, 2);
         let Gamma_1 = &public_parameters.Gamma_1;
         let Gamma_2 = &public_parameters.Gamma_2;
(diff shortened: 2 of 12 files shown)


MyTH-zyxeon:
Maintainer-facing review-assist/deconfliction note for the #557 test-performance lane (no new claim from me):

I checked this PR as a zero-comment #557 performance candidate. Current state is `OPEN`, `MERGEABLE`, with `mergeStateStatus=BLOCKED` and `reviewDecision=REVIEW_REQUIRED` at head `2307c8a7fe3b1a95a0f0096a4398f20c23487e4c`. At the time checked, there were no issue comments, reviews, or review-thread comments. The visible Orca checks are green. The combined commit status API has no legacy statuses (`total_count=0`), and GitHub Actions has `action_required` suites with no check runs, which looks like the usual fork workflow approval gate rather than a code/test failure.

The patch is performance-focused/test-only in the Dory test surface: 12 files, `+241/-203`. It adds `shared_test_public_parameters(max_nu)` in `crates/proof-of-sql/src/proof_primitive/dory/rand_util.rs`, re-exports it from `dory/mod.rs`, and migrates deterministic `PublicParameters::test_rand(..., &mut test_rng())` call sites to the cached helper. The PR body says it collapses repeated deterministic PublicParameters generation from 51 fresh `test_rng()` call sites plus 7 audited `&mut rng` cases while leaving the remaining stateful RNG call sites untouched, with reported local validation of 171 Dory tests.

For overlap hygiene, this is in the same #557 family as #1771/#1774 and the later cached-Dory setup PRs such as #1966/#2069/#2204/#2234. I would treat #2083 as the broad `PublicParameters` caching candidate: it touches the shared Dory helper plus many call sites. The narrower PRs should be considered competing/incremental depending on whether they cache only `PublicParameters`, broader `ProverSetup`/`VerifierSetup`, or only reduce setup sizes in specific test files.

Routing suggestion: if maintainers want the broad process-shared `PublicParameters` cache, #2083 is the most direct candidate I saw. If they prefer lower-risk smaller diffs, route through the narrower #1966/#2069/#2204-style slices and ask #2083 to split or limit scope.

I did not rerun local cargo in this pass because the local `/tmp/sxt-proof-of-sql` checkout was unavailable; this note is remote triage/deconfliction only.

JayWhite2357:

PR #2100, opened 2026-05-29
test: cover column repetition variants

Jorel97:
/claim #560

## Summary
- Adds unit coverage for `ColumnRepeatOp` across tiny/small/big/int128, scalar, decimal, and timestamp columns.
- Adds matching unit coverage for `ElementwiseRepeatOp` across the same column variants.
- Adds coverage for zero repetitions returning an empty column.
- Keeps the change test-only with no runtime behavior changes.

## Validation
- `cargo +1.91.1-x86_64-pc-windows-gnu test -p proof-of-sql --lib --no-default-features --features test column_repetition_operation -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/database/column_repetition_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_repetition_operation.rs
@@ -153,7 +153,11 @@ impl RepetitionOp for ElementwiseRepeatOp {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::base::scalar::test_scalar::TestScalar;
+    use crate::base::{
+        math::decimal::Precision,
+        posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
+        scalar::test_scalar::TestScalar,
+    };
 
     #[test]
     fn test_column_repetition_op() {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow test-only change in `crates/proof-of-sql/src/base/database/column_repetition_operation.rs` (`+137/-1`). It extends coverage for both `ColumnRepeatOp` and `ElementwiseRepeatOp` across TinyInt, SmallInt, BigInt, Int128, Scalar, Decimal75, and TimestampTZ columns, plus a zero-repetition empty-column case.

Main pre-merge points I would check:

- Claim hygiene: this PR carries `/claim #560`, so it should be treated as supplemental coverage in the same bounty pool as the existing #560 submissions unless maintainers decide this is part of the accepted merge set.
- The zero-repetition assertion only uses `Column::Int`; that is fine as a minimal shared-path smoke test if both repetition modes dispatch through the same empty-allocation path, but it does not prove each newly covered Decimal75/TimestampTZ shape preserves its variant when `repetitions == 0`.
- TimestampTZ metadata is covered with non-UTC `ColumnRepeatOp` and UTC `ElementwiseRepeatOp`; if timezone preservation is the intended focus, reviewers may want one non-UTC elementwise assertion too.
- The PR body reports targeted cargo/fmt/diff checks. I did not run the Rust suite locally in this pass.

Visible state: `OPEN` / `MERGEABLE`, but `mergeStateStatus=BLOCKED` and `reviewDecision=REVIEW_REQUIRED`. Orca checks are green; `CI-Lint-And-Test` and `CI-Code-Coverage` are still `action_required`, consistent with the fork workflow approval gate, and there are no visible review threads.

JayWhite2357:

PR #1757, opened 2026-05-19
test: cover naive commitment column variants

applego:
## Summary
- add direct coverage for the remaining `NaiveCommitment::compute_commitments` committable column variants
- cover Boolean, Uint8, TinyInt, SmallInt, Int, Int128, Decimal75, and TimestampTZ inputs with an offset

/claim #560

## Validation
- `cargo fmt --all -- --check`
- `cargo test -p proof-of-sql --no-default-features --features="test" we_can_compute_commitments_from_remaining_committable_column_variants`
- attempted `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql --all-features we_can_compute_commitments_from_remaining_committable_column_variants`; local aarch64 macOS build is blocked by `halo2curves` requiring x86_64 for feature `asm`

PR #2106, opened 2026-05-29
test: cover bool varint encoding

chinook1001:
/claim #560

## Summary
- Add bool varint encoding coverage for previously untested true/false slice paths.

## Coverage
This covers 13 previously uncovered lines, or approximately $1.30 at $0.10/line.

crates/proof-of-sql/src/base/encode/varint_trait.rs: 136, 137, 138, 140, 141, 143, 144, 145, 146, 148, 150, 151, 152

## Tests
- `cargo test --package proof-of-sql --lib bool_values --no-default-features --features test`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/encode/varint_trait_test.rs
+++ b/crates/proof-of-sql/src/base/encode/varint_trait_test.rs
@@ -238,6 +238,13 @@ fn we_can_encode_and_decode_small_u64_values() {
 }
 
 #[test]
+fn we_can_encode_and_decode_bool_values() {
+    test_encode_decode(false, [0]);
+    test_encode_decode(true, [1]);
+    assert_eq!(bool::decode_var(&[2]), None);
+}
+
+#[test]
 fn we_can_encode_and_decode_large_u64_values() {
     test_encode_decode(
         u64::MAX,


MyTH-zyxeon:
Review-assist note for the #560 coverage bounty:

I checked this PR as a very small VarInt coverage slice. It is currently `OPEN`, `MERGEABLE`, `BLOCKED`, and `REVIEW_REQUIRED` at `3575070a4a230e7e7135ab200b144d24a24a6be1`; I do not see prior top-level comments, reviews, or review threads. The visible Orca checks are green. I also do not see public GitHub Actions runs for this head via the current API surface, so review/workflow visibility remains the main gate rather than a visible test failure.

What looks useful here:
- scope is one file only: `crates/proof-of-sql/src/base/encode/varint_trait_test.rs`, `+7/-0`, tests only
- it covers bool VarInt round trips for `false -> [0]` and `true -> [1]`
- it adds the negative boundary that `bool::decode_var(&[2])` returns `None`

Suggested pre-merge check: confirm the intended bool VarInt public contract is exactly single-byte `0`/`1` encoding with other byte values rejected. This appears separable from the active #1757/#1820 nullable-head lane because it targets primitive encoding coverage only.

JayWhite2357:

PR #2115, opened 2026-05-29
perf: cache dory commitment test setup

ajinkyaexams-design:
Refs #557

Problem
- The Dory commitment computation tests rebuild random public parameters and prover setup in every test case.
- That repeated setup work adds avoidable runtime to the test suite targeted by the performance bounty.

Root cause
- Each static and dynamic Dory commitment test constructs equivalent test setup independently before exercising the commitment logic.

Solution
- Add a test-only thread-local Dory setup helper.
- Reuse the cached public parameters and prover setup across Dory commitment computation tests while keeping each assertion unchanged.
- Keep the cache thread-local so feature-gated setup internals do not need to be shared across test threads.

Testing
- cargo fmt --check
- cargo test -p proof-of-sql --lib dory_compute_commitments_test --no-default-features --features="std"

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dory_compute_commitments_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dory_compute_commitments_test.rs
@@ -1,30 +1,27 @@
+use super::test_setup::test_dory_setup;
 use crate::{
     base::{
         commitment::CommittableColumn,
         math::decimal::Precision,
         posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
     },
-    proof_primitive::dory::{
-        compute_dory_commitments, DoryProverPublicSetup, ProverSetup, PublicParameters, F, GT,
-    },
+    proof_primitive::dory::{compute_dory_commitments, DoryProverPublicSetup, F, GT},
 };
 use ark_ec::pairing::Pairing;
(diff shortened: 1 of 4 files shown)


MyTH-zyxeon:
Maintainer-facing review-assist for the #557 performance lane:

This is a focused test-only speedup candidate. It moves repeated `PublicParameters::test_rand(5, ...)` plus `ProverSetup::from(...)` construction in the static and dynamic Dory commitment suites into a thread-local `OnceCell`, while leaving the expected pairing assertions in each case tied to the same cached parameters.

Checks I would use before landing:

- Run the two touched Dory commitment test modules together, not independently, so the cached setup is exercised across the full static + dynamic suite.
- Confirm the shared `PublicParameters` / `ProverSetup` values stay immutable. The leaked `'static` setup is fine for test-only reuse, but any future test that mutates setup state would make this order-dependent.
- Keep the `sigma=0`, offset, empty-column, mixed-column, and signed-value cases in the same run; those are the cases that prove the helper is not only covering the common positive path.

Visible repo state: Orca checks are green, while `CI-Lint-And-Test` and `CI-Code-Coverage` are still `action_required` on the fork head, so the remaining blocker looks like maintainer workflow approval rather than a failing test signal.

JayWhite2357:

PR #2136, opened 2026-05-29
test: cover bigdecimal parsing edge cases

Jorel97:
/claim #560

Summary:
- add focused coverage for `utils::parse::find_bigdecimals`
- verify non-CREATE statements are ignored while CREATE TABLE statements are still collected
- verify DECIMAL columns without `(precision, scale)` and precision <= 38 are ignored
- keep the change test-only with no runtime behavior changes

Validation:
- `cargo +1.91.1-x86_64-pc-windows-gnu test -p proof-of-sql --lib --no-default-features --features test utils::parse::tests -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/utils/parse.rs
+++ b/crates/proof-of-sql/src/utils/parse.rs
@@ -92,4 +92,26 @@ mod tests {
             &empty_vec
         );
     }
+
+    #[test]
+    fn we_can_ignore_non_bigdecimal_columns_and_non_create_statements() {
+        let sql = "SELECT 1;
+
+        CREATE TABLE IF NOT EXISTS ETHEREUM.TRACES(
+            TRACE_ID BIGINT NOT NULL,
+            STANDARD_DECIMAL DECIMAL(38, 2),
+            UNSCALED_DECIMAL DECIMAL(78),
+            BIG_DECIMAL DECIMAL(39, 4),
+            PRIMARY KEY(TRACE_ID)
+        );
+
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist note for the #560 coverage bounty:

I checked this PR as a very small `utils::parse` coverage slice. It is currently `OPEN`, `MERGEABLE`, `BLOCKED`, and `REVIEW_REQUIRED` at `98fde151b500cb2cef760d158735259e945fa0ad`; I do not see prior top-level comments, reviews, or review threads. The visible Orca checks are green. I also do not see public GitHub Actions runs for this head via the current API surface, so review/workflow visibility remains the main gate rather than a visible test failure.

What looks useful here:
- scope is one file only: `crates/proof-of-sql/src/utils/parse.rs`, `+22/-0`, tests only
- it covers a mixed SQL string where non-`CREATE TABLE` statements and non-BigDecimal decimal columns should be ignored
- it keeps the expected extraction narrow: only `DECIMAL(39, 4)` is reported as `BIG_DECIMAL`, while `DECIMAL(38, 2)` and `DECIMAL(78)` stay out of the result

Suggested pre-merge check: confirm this matches the intended `find_bigdecimals` boundary for precision/scale detection, especially the exclusion of regular decimal columns and statements outside the CREATE TABLE parse path. This appears separable from the active #1757/#1820 nullable-head lane because it targets parser helper coverage only.

JayWhite2357:

PR #2137, opened 2026-05-29
test: cover ref into blanket conversion

Jorel97:
/claim #560

Summary:
- add focused coverage for the `RefInto` blanket implementation
- verify a reference conversion dispatches through `ref_into` without consuming the original value
- keep the change test-only with no runtime behavior changes

Validation:
- `cargo +1.91.1-x86_64-pc-windows-gnu test -p proof-of-sql --lib --no-default-features --features test ref_into -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/ref_into.rs
+++ b/crates/proof-of-sql/src/base/ref_into.rs
@@ -23,3 +23,26 @@ where
         self.into()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::RefInto;
+
+    struct LocalValue(u64);
+
+    impl From<&LocalValue> for u64 {
+        fn from(value: &LocalValue) -> Self {
+            value.0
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

I checked this PR as a small, test-only coverage addition. The diff adds a local `From<&LocalValue> for u64` implementation and then calls `value.ref_into()`, which directly exercises the blanket `RefInto` path while keeping the original value readable afterward. That makes the intent narrower than a generic conversion test: it specifically guards the reference-based dispatch this helper promises.

Current external signal: the Orca security checks are green, while the repo CI workflows are still `action_required` for maintainer approval on the fork. I do not see existing PR comments or reviews to answer yet, so the next unblocker appears to be maintainer review / CI approval rather than a code change.

JayWhite2357:

PR #2143, opened 2026-05-29
test: reduce append rows bench setup size

Jorel97:
/claim #557

## Summary
- Reduce the append rows benchmark Dory setup size from max_nu 10 to max_nu 6.
- Keep the parameter named in a local constant so the benchmark's required capacity is explicit.

## Rationale
The largest benchmark case starts with 1000 rows and appends 1000 more rows, so max_nu 6 still covers the 2000-row workload while avoiding extra unused setup generation during local benchmark runs.

## Validation
- `git diff --check`

`cargo +1.91.1-x86_64-pc-windows-gnu check -p proof-of-sql-benches --bench bench_append_rows --no-default-features` was attempted locally, but the Windows environment stops in `blitzar-sys` with `Unsupported OS. Only Linux is supported.`

### DIFF
--- a/crates/proof-of-sql-benches/benches/bench_append_rows.rs
+++ b/crates/proof-of-sql-benches/benches/bench_append_rows.rs
@@ -27,6 +27,8 @@ use proof_of_sql::{
 };
 use rand::Rng;
 
+const APPEND_ROWS_BENCH_MAX_NU: usize = 6;
+
 /// Bench dory performance when appending rows to a table. This includes the computation of
 /// commitments. Chose the number of columns to randomly generate across supported `PoSQL`
 /// data types, and choose the number of rows to append at a time.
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer-facing review assist for `/claim #557`:

- Scope is clean and narrow: one bench-only file, introducing `APPEND_ROWS_BENCH_MAX_NU` and lowering the Dory public parameter setup used by `bench_append_rows` from `10` to `6`.
- Main acceptance check I would want before merge: confirm that `APPEND_ROWS_BENCH_MAX_NU = 6` is still sufficient for the active benchmark cases, especially `bench_append_rows_10x1000`, where the commitment is built from 1000 initial rows and then appends another 1000-row table inside the iteration. If `max_nu` maps to the row-capacity/log2 bound used by the Dory setup, this deserves an explicit validation note or a targeted bench/check run.
- CI status looks like the same fork-gated shape as the other #557/#560 PRs: Orca checks are green, while `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required` for head `57229ba8ce263818965c70db9178114ab16a6ff3`.

Suggested maintainer path: approve/rerun the fork workflows, then either accept this as a small #557 bench-runtime improvement if the `max_nu=6` bound is confirmed, or ask the author to raise the constant / document the exact capacity relation before merge.

JayWhite2357:

PR #2146, opened 2026-05-30
test: reduce dory inner product setup work

Jorel97:
/claim #557

## Summary
- Consolidates related malformed-message Dory inner-product rejection checks so they reuse one generated `PublicParameters`/setup/proof.
- Keeps the same rejection assertions for modified GT messages, missing/extra GT and G1 messages, and an incorrect verifier commitment.
- Derives `Clone` for `VerifierState` only in test builds so the shared fixture can be reused safely.

## Validation
- `cargo +1.91.1-x86_64-pc-windows-gnu test -p proof-of-sql --lib --no-default-features --features test dory_inner_product -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dory/dory_inner_product_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dory/dory_inner_product_test.rs
@@ -1,6 +1,6 @@
 use super::{
     dory_inner_product_prove, dory_inner_product_verify, rand_G_vecs, test_rng, DoryMessages,
-    G1Affine, ProverState, PublicParameters, GT,
+    G1Affine, ProverState, PublicParameters, VerifierSetup, VerifierState, GT,
 };
 use ark_std::UniformRand;
 use merlin::Transcript;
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:
Maintainer-facing review assist for `/claim #557` on this PR:

I checked the current head `27d4848124153a11efaed49df984f9756f6de09c`. This looks like a focused performance slice rather than a behavior change:

- The malformed Dory inner-product rejection cases still cover modified GT, too few/too many GT messages, too few/too many G1 messages, and wrong verifier commitment.
- The expensive `PublicParameters` / prover setup / verifier setup generation is now shared across those malformed-message assertions instead of repeated per test.
- `VerifierState` only gains `Clone` under `cfg(test)`, so the production type surface is not widened.
- The separate transcript-difference and setup-difference tests remain separate, which is the right shape because those cases need independent transcript/setup construction.

1. Keep `/claim #557` hygiene explicit: #1771 is still the older canonical claim PR, so maintainers should decide whether this PR is an alternative/supplemental acceptance path or whether the claim marker should stay consolidated elsewhere.
2. Confirm the listed no-default-features Dory test command on the target toolchain, since the value of the change is runtime reduction without dropping the negative-path assertions.
3. In review, verify each mutation uses a fresh cloned `base_messages` / `verifier_state` so one malformed case cannot leak into the next assertion.

Current visible gate from GitHub metadata is still maintainer-side: `mergeStateStatus=BLOCKED`, `reviewDecision=REVIEW_REQUIRED`, Orca checks green. If the hidden Actions runs are the usual fork `action_required` gate, approving/rerunning those workflows should expose the real CI result.

JayWhite2357:

PR #2200, opened 2026-06-01
test: cover vec commitment extension

stmr:
## Summary
- run `VecCommitmentExt` tests under normal test builds instead of gating them on `blitzar`
- adds no-blitzar coverage for commitment conversion, append, extend, add, sub, and mismatch paths

/claim #560

## Verification
- cargo --config 'target.x86_64-unknown-linux-gnu.linker="cc"' --config 'target.x86_64-unknown-linux-gnu.rustflags=[]' test -p proof-of-sql --lib --no-default-features --features test vec_commitment_ext -- --nocapture\n- cargo fmt --all -- --check\n- git diff --check\n- bash scripts/check_commits.sh

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/vec_commitment_ext.rs
+++ b/crates/proof-of-sql/src/base/commitment/vec_commitment_ext.rs
@@ -171,7 +171,7 @@ impl<C: Commitment> VecCommitmentExt for Vec<C> {
     }
 }
 
-#[cfg(all(test, feature = "blitzar"))]
+#[cfg(test)]
 mod tests {
     use super::*;
     use crate::base::{


MyTH-zyxeon:

This looks like a very narrow #560 coverage slice. I checked the current diff and it is test-only in `crates/proof-of-sql/src/base/commitment/vec_commitment_ext.rs` (1 line changed), widening the existing `VecCommitmentExt` test module from `#[cfg(all(test, feature = "blitzar"))]` to `#[cfg(test)]`.

That makes it complementary to the other open #560 coverage PRs because it does not add another helper/assertion path; it specifically asks whether the existing commitment-vector tests should run in the no-Blitzar test matrix as well.

Current gate note: the PR is open/mergeable but still blocked on review/workflow approval. The head SHA `c8e059f16c2c2cabb72b0eec493a1de840be6dd5` has `CI-Lint-And-Test` and `CI-Code-Coverage` at `action_required`, so I do not see a code-test failure signal from the available checks yet.

Main validation point before merge: because the change only broadens the compile/test gate, the useful maintainer check is the listed no-default-features cargo command to confirm none of the existing tests still require Blitzar-only symbols at compile time.

stmr:
Thanks for the review-assist pass. Intent note for maintainers: this stays intentionally tiny, widening vec commitment extension coverage without changing runtime logic.

JayWhite2357:

PR #2192, opened 2026-06-01
test: cover dynamic vec panic path

stmr:
## Summary
- add focused coverage for compute_dynamic_vecs when a point coordinate is exactly 1
- assert the existing panic message for the documented invalid input path

/claim #560

## Validation
- cargo --config 'target.x86_64-unknown-linux-gnu.linker="cc"' --config 'target.x86_64-unknown-linux-gnu.rustflags=[]' test -p proof-of-sql --lib --no-default-features --features test we_cannot_compute_dynamic_vecs_from_point_with_one -- --nocapture
- cargo fmt --all -- --check
- git diff --check
- bash scripts/check_commits.sh

## Deconfliction
Checked current open #560 PRs for standard_basis_helper, dynamic standard basis, compute_dynamic_vecs point=1, and the exact panic message. No active overlapping slice found.

PR #2207, opened 2026-06-02
test: cover dynamic vec zero coordinates

hp283260133-bit:
## Summary
- add a deterministic `compute_dynamic_vecs` coverage case with zero coordinates in the evaluation point
- verify the dynamic row/column factorization still matches the full evaluation vector for those zero-coordinate inputs
- keep production code unchanged

/claim #560

## Validation
- `git diff --check`

I could not run the focused cargo test locally because this Windows environment does not currently have `cargo`/`rustc` on PATH.

## Deconfliction
Checked current open PRs and #560 comments for `standard_basis_helper`, `standard basis`, `compute_dynamic_vecs`, and the existing dynamic-vector panic slice. This PR does not touch the already-open panic-path coverage in #2192.

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/dynamic_matrix_utils/standard_basis_helper.rs
+++ b/crates/proof-of-sql/src/proof_primitive/dynamic_matrix_utils/standard_basis_helper.rs
@@ -723,4 +723,22 @@ pub(crate) mod tests {
             }
         }
     }
+
+    #[test]
+    fn we_can_compute_dynamic_vecs_that_match_evaluation_vec_with_zero_coordinates() {
+        let point = [
+            DoryScalar::from(0),
+            DoryScalar::from(2),
+            DoryScalar::from(0),
+            DoryScalar::from(3),
+            DoryScalar::from(5),
+        ];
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist note for #560 triage.

This is a very narrow test-only coverage PR: 1 changed file, +18/-0 lines, and no production-code changes.

Useful coverage added:
- adds a deterministic `compute_dynamic_vecs` case where the evaluation point contains zero coordinates
- checks that dynamic row/column factorization still matches the full evaluation vector for that zero-coordinate input
- keeps the scope confined to `standard_basis_helper.rs`

Pre-merge check I would ask maintainers to confirm: the PR body says this is deconflicted from #2192's dynamic-vector panic slice. That seems plausible from the visible diff because this PR exercises the non-panic zero-coordinate factorization path, but it is worth confirming the two dynamic-vector coverage claims are treated as separate non-overlapping #560 slices.

Public status surface: Orca checks are green; GitHub Actions `CI-Lint-And-Test` and `CI-Code-Coverage` are both `action_required` for head `702130991a2643088358fee1dc76dc52558e1c22`, so this looks workflow-approval gated rather than blocked on a visible test failure.

PR #2214, opened 2026-06-03
refactor: blanket impl VarInt for Scalar

barton87:
# Rationale for this change

Addresses the remaining `VarInt` item from #228 by removing `VarInt` as an explicit `Scalar` supertrait requirement and replacing the concrete `MontScalar` implementation with a blanket `impl<S: Scalar> VarInt for S`.

/claim #228

- Removed the `+ VarInt` bound from `Scalar`.
- Added explicit `Scalar::from_limbs` / `Scalar::to_limbs` helpers with default implementations backed by the existing limb conversions.
- Reimplemented scalar varint encoding generically for any `Scalar`, using those explicit limb helpers.
- Preserved the existing scalar zig-zag/varint behavior while avoiding a `MontScalar`-specific `VarInt` impl.

# Are these changes tested?

Yes.

- `rustfmt --check crates/proof-of-sql/src/base/scalar/scalar.rs crates/proof-of-sql/src/base/encode/varint_trait.rs`
- `BLITZAR_BACKEND=cpu cargo check -p proof-of-sql --lib --no-default-features`
- `BLITZAR_BACKEND=cpu cargo test -p proof-of-sql base::encode::scalar_varint --lib` — 15 passed

Note: I also tried the targeted scalar-varint test under `--no-default-features`, but the repo test harness currently fails before reaching these tests due unrelated missing `vec`/`Box` imports across many test modules. The library check under `--no-default-features` passes, and the scalar varint behavior tests pass under the repo's normal feature set.

### DIFF
--- a/crates/proof-of-sql/src/base/encode/varint_trait.rs
+++ b/crates/proof-of-sql/src/base/encode/varint_trait.rs
@@ -10,16 +10,12 @@
 // There were significant code changes to simplify the code
 // ---------------------------------------------------------------------------------------------------------------
 use super::{
-    scalar_varint::{
-        read_scalar_varint, read_u256_varint, scalar_varint_size, u256_varint_size,
-        write_scalar_varint, write_u256_varint,
-    },
+    scalar_varint::{read_u256_varint, u256_varint_size, write_u256_varint},
     U256,
 };
-use crate::base::scalar::MontScalar;
+use crate::base::scalar::Scalar;
 #[cfg(test)]
 use alloc::{vec, vec::Vec};
-use ark_ff::MontConfig;
 
 /// Most-significant byte, == 0x80
 pub const MSB: u8 = 0b1000_0000;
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:
Maintainer-facing review assist for the `#228` blanket-`VarInt` slice:

- Visible state from here: `OPEN` / `MERGEABLE`, but `mergeStateStatus=BLOCKED` and `reviewDecision=REVIEW_REQUIRED`. Orca checks are green; `CI-Code-Coverage` and `CI-Lint-And-Test` are still `action_required` with `jobs=[]`, so this looks like the usual fork workflow approval/rerun gate rather than a code-test failure.
- Scope looks aligned with the third checkbox in #228: remove the `VarInt` supertrait from `Scalar` and provide `impl<S: Scalar> VarInt for S` using limb conversions.
- Main landing-order item to confirm: this PR adds default `Scalar::from_limbs` / `Scalar::to_limbs` helpers while the first checkbox of #228 separately wants the implicit limb-conversion bounds removed. If a limb-conversion PR lands first or is selected as canonical, this PR likely needs a small rebase so the default helpers are either kept only while the old bounds remain, or moved to concrete scalar impls once those bounds are removed.
- Suggested acceptance check: compare the new blanket scalar varint encoding/decoding against the prior `MontScalar` path for boundary values such as zero, one, negative one, max-signed, and a high-limb value. The PR body reports `base::encode::scalar_varint` passing; those byte-for-byte boundary checks are the clearest maintainer signal that the generic zig-zag helpers preserved the old wire format.

No local checkout was available in this automation run, so I did not rerun Rust tests; this is a review-routing note based on the visible PR diff and CI state.

JayWhite2357:

PR #2218, opened 2026-06-03
test: cover provable result guards

stmr:
## Summary
- add `ProvableQueryResult::evaluate` coverage for mismatched result-field counts
- add `ProvableQueryResult::to_owned_table` coverage for mismatched result-field counts
- keep the change test-only in the existing provable query-result test module

/claim #560

## Non-overlap
Checked current open PRs for `ProvableQueryResult`, `provable_query_result`, `InvalidColumnCount`, and result-field count guards before taking this slice; no active overlapping claim was found.

## Verification
- `cargo --config 'target.x86_64-unknown-linux-gnu.linker="cc"' --config 'target.x86_64-unknown-linux-gnu.rustflags=[]' test -p proof-of-sql --lib --no-default-features --features 'arrow test' provable_query_result_test -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/sql/proof/provable_query_result_test.rs
+++ b/crates/proof-of-sql/src/sql/proof/provable_query_result_test.rs
@@ -167,6 +167,22 @@ fn evaluation_fails_if_extra_data_is_included() {
 }
 
 #[test]
+fn evaluation_fails_if_the_column_field_count_is_invalid() {
+    let cols: [Column<TestScalar>; 1] = [Column::BigInt(&[10, 12])];
+    let res = ProvableQueryResult::new(2, &cols);
+    let evaluation_point = [TestScalar::from(10u64), TestScalar::from(100u64)];
+    let column_fields = [
+        ColumnField::new("a".into(), ColumnType::BigInt),
+        ColumnField::new("b".into(), ColumnType::BigInt),
+    ];
+
+    assert!(matches!(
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This looks like a small, non-overlapping #560 coverage slice. I checked the current diff and it is test-only in `crates/proof-of-sql/src/sql/proof/provable_query_result_test.rs` (+31/-0), adding two focused guards around `ProvableQueryResult` result-field mismatch behavior:

- `evaluate(...)` returns `QueryError::InvalidColumnCount` when the supplied `ColumnField` count does not match the encoded result columns.
- `to_owned_table(...)` returns the same error for the conversion path.

That makes it complementary to the other open #560 coverage PRs because it targets the provable-query-result boundary rather than commitment, scalar, table utility, or dynamic-vector helpers.

Current gate note: the PR is open/mergeable but still blocked on review/workflow approval. The head SHA `ff76acd4d7d0dc1b3f211fbbebb5203e21e73ccf` has `CI-Code-Coverage` and `CI-Lint-And-Test` at `action_required`, so I do not see a code-test failure signal from the available checks yet. A maintainer rerun/approval plus the listed targeted cargo command should be enough to validate this slice.

stmr:
Thanks for the review-assist pass. Intent note for maintainers: this PR stays test-only and covers provable result guard behavior without changing query/result execution paths.

JayWhite2357:

PR #2223, opened 2026-06-03
test: cover math log helpers

stmr:
## Summary
- add direct `is_pow2` coverage across unsigned integer widths
- cover `log2_down`/`log2_up` boundary behavior for max-width unsigned values
- add focused `is_pow2_bytes` and `log2_down_bytes` little-endian edge coverage

/claim #560

## Non-overlap
Checked current open #560 PRs for math log helpers, `is_pow2`, `log2_down`, `log2_up`, `is_pow2_bytes`, and `log2_down_bytes` before taking this slice; no active overlapping claim was found.

## Verification
- `cargo --config 'target.x86_64-unknown-linux-gnu.linker="cc"' --config 'target.x86_64-unknown-linux-gnu.rustflags=[]' test -p proof-of-sql --lib --no-default-features --features test base::math::log -- --nocapture`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/math/log.rs
+++ b/crates/proof-of-sql/src/base/math/log.rs
@@ -71,6 +71,52 @@ mod tests {
         assert_eq!(log2_up(4u32), 2);
     }
 
+    #[test]
+    fn test_is_pow2_for_unsigned_widths() {
+        assert!(is_pow2(1u8));
+        assert!(is_pow2(128u8));
+        assert!(is_pow2(1u16 << 15));
+        assert!(is_pow2(1u64 << 63));
+        assert!(!is_pow2(3u8));
+        assert!(!is_pow2(u16::MAX));
+        assert!(!is_pow2(u64::MAX));
+    }
+
+    #[test]
+    fn test_log2_for_unsigned_width_boundaries() {
+        assert_eq!(log2_down(u8::MAX), 7);
+        assert_eq!(log2_up(u8::MAX), 8);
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist note for #560 triage.

This is a narrow test-only coverage PR: one file, `crates/proof-of-sql/src/base/math/log.rs`, with +46 lines and no production-code changes.

Useful coverage added:
- `is_pow2` across unsigned widths, including high-bit boundaries and non-powers
- `log2_down` / `log2_up` boundary behavior for `u8`, `u16`, `u64`, and `u128::MAX`
- byte-slice helpers for little-endian power-of-two detection and floor-log cases

Pre-merge check I would ask maintainers to confirm: the byte helper tests intentionally treat `[0, 0, 0, 0]` as power-of-two / floor-log zero because that matches current behavior. If zero is meant to be a special invalid case in a future cleanup, these tests would lock in the current semantics.

I do not see a public Actions run for head `925c3724262053b2b6cd81a9c90ad731d7d9d33b` and the combined status API returns no statuses, so from the public surface this looks review-gated rather than blocked on a visible failing check.

stmr:
Thanks for the review-assist pass. Intent note for maintainers: the byte-helper assertions intentionally pin the current zero-slice behavior because it is what the existing helpers return today. This PR is test-only and does not try to redefine zero as an invalid/special case; if maintainers want that semantics change, I can update the test with the production behavior change rather than hiding it inside this coverage slice.

JayWhite2357:

PR #2226, opened 2026-06-03
test: cover timestamp scale casts

stmr:
Adds focused test-only coverage for scale_cast_binary_op timestamp upcast paths in both directions. It verifies TimestampTZ second-to-millisecond and millisecond-to-second scaling casts use DynProofExpr::try_new_scaling_cast with the opposite timestamp type.

/claim #560

Tested:
- cargo --config target.x86_64-unknown-linux-gnu.linker="cc" --config target.x86_64-unknown-linux-gnu.rustflags=[] test -p proof-of-sql --lib --no-default-features --features test sql::scale -- --nocapture
- cargo fmt --all -- --check
- git diff --check
- scripts/check_commits.sh

### DIFF
--- a/crates/proof-of-sql/src/sql/scale.rs
+++ b/crates/proof-of-sql/src/sql/scale.rs
@@ -73,7 +73,10 @@ pub fn scale_cast_binary_op(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::base::database::{ColumnRef, TableRef};
+    use crate::base::{
+        database::{ColumnRef, TableRef},
+        posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
+    };
 
     #[expect(non_snake_case)]
     fn COLUMN1_BOOLEAN() -> DynProofExpr {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

I checked the diff against the stated coverage goal. This is a focused test-only addition in `sql::scale` that adds `TimestampTZ` second and millisecond proof expressions, then covers both `scale_cast_binary_op` directions. The assertions are useful because they compare against `DynProofExpr::try_new_scaling_cast(...)` with the opposite timestamp type, so the reviewer can see exactly which side should be upcast.

Non-overlap note for the reviewer queue: this targets timestamp scale-cast behavior, not the neighboring proof-error, sumcheck, or row-comparison #560 coverage PRs. One small pre-merge check is whether UTC-only timestamp fixtures are sufficient for this path, since the new helpers use `PoSQLTimeZone::utc()`; if time-zone semantics are intentionally irrelevant to scale selection, this looks narrowly scoped while CI remains fork-gated.

stmr:
Thanks for the review-assist pass. Intent note for maintainers: the UTC fixtures are intentional here because this coverage targets timestamp scale-selection behavior, not time-zone interpretation. The tests compare the expected `DynProofExpr::try_new_scaling_cast(...)` direction for second/millisecond `TimestampTZ` expressions while keeping timezone semantics unchanged.

JayWhite2357:

PR #2227, opened 2026-06-03
test: cover sumcheck round guards

stmr:
Adds focused test-only coverage for prove_round guard paths: verifier message on the first round, missing verifier message after the first round, and inactive prover state. No production behavior changes.

/claim #560

Tested:
- cargo --config target.x86_64-unknown-linux-gnu.linker="cc" --config target.x86_64-unknown-linux-gnu.rustflags=[] test -p proof-of-sql --lib --no-default-features --features test prove_round -- --nocapture
- cargo fmt --all -- --check
- git diff --check
- scripts/check_commits.sh

### DIFF
--- a/crates/proof-of-sql/src/proof_primitive/sumcheck/proof_test.rs
+++ b/crates/proof-of-sql/src/proof_primitive/sumcheck/proof_test.rs
@@ -12,7 +12,7 @@ use crate::{
     },
     proof_primitive::{
         inner_product::curve_25519_scalar::Curve25519Scalar,
-        sumcheck::{ProverState, SumcheckProof},
+        sumcheck::{prove_round, ProverState, SumcheckProof},
     },
 };
 use alloc::rc::Rc;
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

I checked the diff against the stated coverage goal. This is a small test-only PR that imports `prove_round`, adds a reusable `simple_prover_state()`, and pins three guard paths: verifier input on the first prover round, missing verifier input after the first round, and an inactive prover state. That gives the reviewer a clear non-overlap claim because it targets sumcheck round-state validation rather than the display, row-comparison, or planner coverage in the neighboring #560 PRs.

Pre-merge note: these tests intentionally assert the current panic messages (`first round should be prover first.`, `verifier message is empty`, `Prover is not active`). If those panic strings are considered stable enough for regression coverage, this is narrowly scoped and easy to review. If the project treats them as internal-only wording, the maintainer may prefer `#[should_panic]` without exact text, but the current form is stronger for coverage clarity while CI is still fork-gated.

stmr:
Thanks for the review-assist pass. Intent note for maintainers: the exact panic-message assertions are deliberate current-behavior regression coverage for these sumcheck guard paths. This PR is test-only and keeps production behavior unchanged; if panic wording is not considered stable enough to pin, I can loosen the assertions while preserving the guard-path coverage.

JayWhite2357:

PR #2228, opened 2026-06-03
test: cover proof error displays

stmr:
Adds focused test-only coverage for ProofError, ProofSizeMismatch, and PlaceholderError display paths, including transparent source-error display forwarding. No production behavior changes.

/claim #560

Tested:
- cargo --config target.x86_64-unknown-linux-gnu.linker="cc" --config target.x86_64-unknown-linux-gnu.rustflags=[] test -p proof-of-sql --lib --no-default-features --features test base::proof::error -- --nocapture
- cargo fmt --all -- --check
- git diff --check
- scripts/check_commits.sh

### DIFF
--- a/crates/proof-of-sql/src/base/proof/error.rs
+++ b/crates/proof-of-sql/src/base/proof/error.rs
@@ -90,3 +90,133 @@ pub enum PlaceholderError {
 
 /// Result type for placeholder errors
 pub type PlaceholderResult<T> = Result<T, PlaceholderError>;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn proof_errors_display_expected_messages() {
+        assert_eq!(
+            ProofError::VerificationError {
+                error: "bad transcript"
+            }
+            .to_string(),
+            "Verification error: bad transcript"
+        );
+        assert_eq!(
+            ProofError::UnsupportedQueryPlan {
+                error: "cross join"
+            }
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

I checked the diff against the stated coverage goal. This is a focused test-only addition in `base::proof::error` that covers the display strings for `ProofError`, `ProofSizeMismatch`, and `PlaceholderError`, plus the transparent source-error forwarding paths for nested `ProofSizeMismatch` and `PlaceholderError` variants. That gives the reviewer a clear non-overlap story: the PR is not changing production behavior, it is pinning the public error text currently emitted by the existing derives/annotations.

Small pre-merge note: `ProofSizeMismatch::TooFewChiLengths` is asserted as `Proof has too few one lengths`. If that wording is the intended current contract from the existing display annotation, this is a useful regression guard. If the project would rather rename that user-facing text to `chi lengths`, this test would be the place to make that decision explicit before merging.

stmr:
Thanks for the review-assist pass. Intent note for maintainers: this PR is test-only and deliberately pins the current display output from the existing error annotations, including `Proof has too few one lengths` for `ProofSizeMismatch::TooFewChiLengths`. I did not intend to rename production/user-facing text in this coverage slice; if maintainers prefer changing that wording to `chi lengths`, I can adjust the test together with the production display text in a follow-up revision.

JayWhite2357:

PR #2229, opened 2026-06-03
test: cover varbinary row comparison

stmr:
## Summary
- add the missing VarBinary branch for compare_single_row_of_tables
- add regression coverage for VarChar-tied rows where VarBinary decides ordering

## Tests
- cargo --config 'target.x86_64-unknown-linux-gnu.linker="cc"' --config 'target.x86_64-unknown-linux-gnu.rustflags=[]' test -p proof-of-sql --lib --no-default-features --features test order_by_util -- --nocapture
- cargo fmt --all -- --check
- git diff --check
- scripts/check_commits.sh

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/base/database/order_by_util.rs
+++ b/crates/proof-of-sql/src/base/database/order_by_util.rs
@@ -88,6 +88,9 @@ pub(crate) fn compare_single_row_of_tables<S: Scalar>(
             (Column::VarChar((left_col, _)), Column::VarChar((right_col, _))) => {
                 left_col[left_row_index].cmp(right_col[right_row_index])
             }
+            (Column::VarBinary((left_col, _)), Column::VarBinary((right_col, _))) => {
+                left_col[left_row_index].cmp(right_col[right_row_index])
+            }
             // Should never happen since we checked the column types
             _ => unreachable!(),
         };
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:
Review-assist pass, since this is tied to `/claim #560`.

I checked the diff against the stated coverage goal. The implementation adds the missing `Column::VarBinary` match arm in `compare_single_row_of_tables`, and the new test does a nice job forcing the `VarChar` column to tie so the `VarBinary` column actually decides ordering. The three assertions cover greater, less, and equal outcomes, so the regression target is clear and reviewer-friendly.

One small note for the maintainer/reviewer queue: the comparison here is byte-slice lexicographic ordering via `.cmp(...)`, matching the existing `VarChar` branch shape rather than scalar encoding order. If that ordering contract is intended for `VarBinary`, this looks narrowly scoped and claim-relevant. If there is any concern, a short note in the PR body that the branch intentionally compares the raw byte slices would make the intent explicit while CI remains fork-gated.

stmr:
Thanks for the review-assist pass. Intent note for maintainers: this branch deliberately matches the existing VarChar branch shape and compares raw VarBinary byte slices with `.cmp(...)` in `compare_single_row_of_tables`; the test forces the preceding VarChar value to tie so the VarBinary branch is the ordering decision. No scalar/encoding-order change is intended.

JayWhite2357:

PR #2262, opened 2026-06-06
test: cover composite polynomial trace annotation

seekynn:
/claim #560

## Summary
- add focused coverage for `CompositePolynomial::annotate_trace`
- capture only `Product #...` tracing info events with a scoped test subscriber
- assert the exact trace payload for the coefficient and multiplicand indexes, not just the product numbering
- enable `tracing/std` for dev-dependencies only so tests can install a scoped default subscriber

AI-assisted with OpenAI Codex; I reviewed the diff and validation before submitting.

## Validation
- `cargo test -p proof-of-sql --lib base::polynomial::composite_polynomial_test --no-default-features --features test --offline -- --nocapture`
- `cargo check -p proof-of-sql --lib --no-default-features --offline`
- `git diff --check HEAD~1..HEAD`

`cargo llvm-cov` was attempted locally, but this machine could not install `llvm-tools-preview` for toolchain `1.91.1-aarch64-apple-darwin` because rustup reported a missing manifest.

### DIFF
--- a/crates/proof-of-sql/Cargo.toml
+++ b/crates/proof-of-sql/Cargo.toml
@@ -67,7 +67,7 @@ merlin = { workspace = true }
 rand = { workspace = true, default-features = false, features = ["std"] }
 rand_core = { workspace = true, default-features = false }
 serde_json = { workspace = true }
-tracing = { workspace = true }
+tracing = { workspace = true, features = ["std"] }
 proptest = { workspace = true }
 proptest-derive = { workspace = true }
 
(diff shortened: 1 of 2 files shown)


MyTH-zyxeon:
I did a credential-free review pass on this #560 coverage PR. The shape looks narrow and reviewable: one dev-dependency feature tweak plus one test around `CompositePolynomial::annotate_trace`; Orca checks are green, while the repo CI runs are still in the usual fork `action_required` state.

Maintainer checklist I would use before approving/landing:

- Strengthen the new assertion beyond event count and `Product #0/#1`. The production trace includes the coefficient and multiplicand index vector, so checking for the `7 * [0, 1]` and `11 * [2]` payload shape would make the test catch regressions in the useful part of the trace, not only the product numbering.
- Consider filtering recorded events to messages that actually start with `Product #` before pushing into `events`. As written, any future INFO event from the same target with no `message` field would be stored as an empty string and could make this test fail for noise rather than for a trace-format regression.
- Keep the `tracing/std` feature addition clearly dev-only. The PR body already includes a no-default `cargo check`; that is the right guard so this test helper does not accidentally mask normal `proof-of-sql` no-std compilation.

No secrets, local checkout mutation, claim submission, or live workflow action from my side.

JayWhite2357:

PR #2264, opened 2026-06-06
test: cover numerical arithmetic helpers

ajinkyaexams-design:
## Rationale
- Adds direct unit coverage for the shared dd_subtract_columns and multiply_columns helpers.
- Covers mixed column variants, signed arithmetic results, decimal-backed scalar multiplication, and multiply length-mismatch panic behavior.

Refs #560
/claim #560

## Testing
- cargo fmt --check`n- cargo test -p proof-of-sql --lib sql::proof_exprs::numerical_util::tests --no-default-features --features=`"std`" (34 passed)

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs
@@ -662,8 +662,8 @@ pub fn cast_column_with_scaling<'a, S: Scalar>(
 #[cfg(test)]
 mod tests {
     use super::{
-        cast_bool_column_to_signed_int_column, cast_column, cast_int_slice_to_int_column,
-        divide_columns, divide_integer_columns,
+        add_subtract_columns, cast_bool_column_to_signed_int_column, cast_column,
+        cast_int_slice_to_int_column, divide_columns, divide_integer_columns, multiply_columns,
     };
     use crate::{
         base::{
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainer-facing review assist for the `#560` queue:

- Scope: this is a one-file test-only slice in `crates/proof-of-sql/src/sql/proof_exprs/numerical_util.rs` (`+62/-2`) covering the direct `add_subtract_columns` add/subtract paths, direct `multiply_columns`, and the multiply length-mismatch panic.
- Claim hygiene: the PR body and top-level comment both carry `/claim #560`; this looks distinct from the recent #560 slices around serialization, table/record-batch errors, filter/sumcheck gadgets, and query/commitment helpers because it targets the numerical arithmetic helper entry points directly.
- Current visible gate: GitHub shows `OPEN` / `MERGEABLE`, but `mergeStateStatus=BLOCKED` and `reviewDecision=REVIEW_REQUIRED`. Orca checks are green. `CI-Lint-And-Test` and `CI-Code-Coverage` are both `action_required` with `jobs=[]` on head `1ff5f71c55e033f6d9878a3c1a2a110cf036b1f5`, so this appears workflow-approval gated rather than blocked on a surfaced code failure.

- Approve/rerun the fork workflows so the listed `numerical_util::tests` command actually executes on GitHub. The PR body's Testing line has visible formatting escapes (`` `n`` / escaped feature quotes), so CI output is the clean source of truth.
- If the goal includes the shared length assertion for both helper families, consider adding the same mismatched-length panic coverage for `add_subtract_columns`; right now that explicit panic test exists only for `multiply_columns`.
- The `Decimal75` multiply case uses scale `0`, which is fine for scalar-backed multiplication coverage. If maintainers expect decimal scale semantics to be covered too, that would need a separate nonzero-scale assertion.

No local checkout mutation, claim submission, or live workflow action from my side.

JayWhite2357:

PR #2269, opened 2026-06-06
test: cover planner conversion edge cases

ibobgunardi:
Adds two planner conversion tests:

- invalid three-part table names return `InvalidTableReference`
- `optimizer()` keeps `common_sub_expression_eliminate` out of the rule set

Tests:
- `git diff --check`
- Tried `cargo test -p proof-of-sql-planner --no-default-features conversion::tests -- --nocapture`, but this Windows host is missing `link.exe`; Docker Desktop also failed locally, so CI is the compile gate.

/claim #560

### DIFF
--- a/crates/proof-of-sql-planner/src/conversion.rs
+++ b/crates/proof-of-sql-planner/src/conversion.rs
@@ -112,12 +112,12 @@ pub fn get_table_refs_from_statement(
 
 #[cfg(test)]
 mod tests {
-    use super::get_table_refs_from_statement;
+    use super::{get_table_refs_from_statement, optimizer};
     use crate::{conversion::sql_to_posql_plans, PlannerResult};
     use datafusion::{config::ConfigOptions, logical_expr::LogicalPlan};
     use indexmap::IndexSet;
     use proof_of_sql::{
-        base::database::{TableRef, TableTestAccessor},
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

I checked this PR against the current body, diff, visible checks, comments, reviews, and review-thread state. The change is a narrow `crates/proof-of-sql-planner/src/conversion.rs` test-only slice: it adds coverage that three-part table names return `ParseError::InvalidTableReference`, and that `optimizer()` does not include `common_sub_expression_eliminate` in its rule set.

I do not see comments, reviews, or unresolved review threads on this PR right now. Orca checks are green; `CI-Code-Coverage` and `CI-Lint-And-Test` are `action_required`, which matches the author's note that CI is the compile gate after local Windows `link.exe`/Docker issues, rather than an observed code failure.

Non-overlap note: this is distinct from the recent #560 coverage lanes around planner filter pushdown, AnalyzeError conversion, df_util helper coverage, error display formatting, table/column accessor tests, bit-distribution coverage, evaluation-vector bounds, and standard serialization coverage.

ibobgunardi:
Happy to adjust if CI or review turns up anything.

JayWhite2357:

PR #2243, opened 2026-06-06
Make opaque Scalar conversions explicit rather than using From/Into trait

heickerv1001-dev:
## Summary

- add explicit `Scalar::from_limbs` and `Scalar::to_limbs` methods
- remove opaque limb conversion bounds from `Scalar`
- remove concrete limb `From` conversions for `MontScalar`
- update production call sites and tests to use the explicit limb methods

## Notes

This now addresses only the limb-conversion checkbox from #228. I split the blanket `VarInt` refactor back out so this PR lines up with the issue's request for separate PRs.

## Testing

Not run locally: this workspace does not currently have `cargo` / `rustc` installed.

/claim #228

PR #2189, opened 2026-05-31
refactor: blanket impl VarInt for Scalar

meaqua9420:
## Summary
- remove the `VarInt` supertrait bound from `Scalar`
- add a blanket `impl<S: Scalar> VarInt for S`
- make scalar varint and zigzag helpers generic over `Scalar` while preserving the existing U256 varint representation

## Validation
- `cargo fmt --check`
- `cargo check -p proof-of-sql --lib --no-default-features --features="arrow rayon ark-ec/parallel ark-poly/parallel ark-ff/asm"`
- `cargo test -p proof-of-sql zigzag --no-default-features --features="arrow rayon ark-ec/parallel ark-poly/parallel ark-ff/asm"`
- `cargo test -p proof-of-sql scalar_varint --no-default-features --features="arrow rayon ark-ec/parallel ark-poly/parallel ark-ff/asm"`
- `cargo test -p proof-of-sql varint_trait --no-default-features --features="arrow rayon ark-ec/parallel ark-poly/parallel ark-ff/asm"`
- `source scripts/check_commits.sh`

Part of #228: this PR handles the `VarInt` bullet only.

/claim #228

PR #2272, opened 2026-06-06
Make scalar string hashing explicit

heickerv1001-dev:
# PR Body Draft for `#228` String-Hash Slice

## Summary

- add `Scalar::from_str_via_hash(&str)` with a default implementation
- remove opaque string-conversion bounds from `Scalar`
- update varchar conversion call sites to use the explicit string-hash helper

## Notes

This PR intentionally covers only the second checkbox from `#228`.
The limb-conversion PR is already split out separately, and the blanket `VarInt` refactor stays out of this slice.

## Testing

Not run locally in this workspace because `cargo` / `rustc` are unavailable here.

/claim #228

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/arrow_array_to_column_conversion.rs
+++ b/crates/proof-of-sql/src/base/arrow/arrow_array_to_column_conversion.rs
@@ -279,7 +279,9 @@ impl ArrayRefExt for ArrayRef {
                     let scals = if let Some(scals) = precomputed_scals {
                         &scals[range.start..range.end]
                     } else {
-                        alloc.alloc_slice_fill_with(vals.len(), |i| -> S { vals[i].into() })
+                        alloc.alloc_slice_fill_with(vals.len(), |i| -> S {
+                            S::from_str_via_hash(vals[i])
+                        })
                     };
 
                     Ok(Column::VarChar((vals, scals)))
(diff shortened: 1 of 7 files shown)


MyTH-zyxeon:
Maintainer-facing review assist for the #228 string-hash slice.

Visible state from the PR metadata: `OPEN` / `MERGEABLE`, but `mergeStateStatus=BLOCKED` and `reviewDecision=REVIEW_REQUIRED`. Orca checks are green. `CI-Lint-And-Test` and `CI-Code-Coverage` are still `action_required`, so maintainer workflow approval/rerun is still needed before the real Rust suite can validate this.

The main thing I would resolve before review is scope alignment. The PR body says this is only the second #228 checkbox (`from_str_via_hash`), but the diff against current `main` also removes the limb-conversion bounds (`From<[u64; 4]>`, `Into<[u64; 4]>`, `RefInto<[u64; 4]>`), adds `from_limbs` / `to_limbs`, and removes the `VarInt` super-bound. That overlaps the first and third #228 checkboxes, plus the existing focused candidates #2243 and #2189. If this should be the clean second-slice PR, I would either base it after the selected limb PR lands or remove the limb/VarInt changes from this branch.

One concrete blocker in the current diff: `OwnedColumn::Decimal75(_, _, Vec<S>)` is changed to feed each scalar through `S::from_str_via_hash(s)`. Those entries are already `S`, not strings, and decimal commitments should preserve the scalar value/limbs rather than hash text. Against current `main`, that also looks type-incompatible. The Decimal75 arm should stay on scalar limb conversion (`to_limbs` after the limb slice lands, or the existing `Into<[u64; 4]>` path until then).

Suggested pre-merge checks once the branch is scoped:
- Confirm the PR contains only the string-hash checkbox and leaves limb conversion / blanket VarInt to their selected PRs.
- Add or keep focused coverage proving old varchar string hashing equals `Scalar::from_str_via_hash` for `LiteralValue`, `Column::VarChar`, Arrow string conversion with and without precomputed scalars, and `CommittableColumn::VarChar`.
- Re-run maintainer CI after approving fork workflows; contributor-visible checks currently do not show the Rust compile/test result.

PR #2278, opened 2026-06-07
fix: detect numeric and dec bigdecimal columns

qiuning100876-commits:
/claim #560

> Disclosure: This PR was authored with assistance from Codex.

## Summary
- Detect `NUMERIC(p, s)` and `DEC(p, s)` columns as bigdecimal columns alongside `DECIMAL(p, s)`.
- Keep the existing `precision > 38` threshold, including excluding `DECIMAL(38, 0)`.
- Cover SQL exact numeric aliases in `test_find_bigdecimals`.

## Non-overlap check
Before opening this PR, I checked the open #560 PRs touching `parse.rs`. Those cover non-`CREATE TABLE` handling and `DECIMAL` threshold/boundary cases. This PR intentionally does not change the current `DECIMAL(p)` precision-only behavior and focuses on SQL aliases (`NUMERIC` and `DEC`) instead.

## Validation
- `cargo f --check`
- `cargo test --manifest-path crates/proof-of-sql/Cargo.toml --no-default-features --features std test_find_bigdecimals --lib`
- `cargo clippy --manifest-path crates/proof-of-sql/Cargo.toml --no-default-features --features std --lib -- -A warnings -D clippy::cast_possible_truncation`
- `cargo test --manifest-path crates/proof-of-sql/Cargo.toml --no-default-features --features std --lib`

### DIFF
--- a/crates/proof-of-sql/src/utils/parse.rs
+++ b/crates/proof-of-sql/src/utils/parse.rs
@@ -10,6 +10,20 @@ use sqlparser::{
 };
 
 #[expect(clippy::cast_possible_truncation)]
+fn bigdecimal_spec(data_type: &DataType) -> Option<(u8, i8)> {
+    match data_type {
+        DataType::Decimal(exact_number_info)
+        | DataType::Numeric(exact_number_info)
+        | DataType::Dec(exact_number_info) => match exact_number_info {
+            ExactNumberInfo::PrecisionAndScale(precision, scale) if *precision > 38 => {
+                Some((*precision as u8, *scale as i8))
+            }
+            _ => None,
+        },
+        _ => None,
+    }
+}
+
 /// Parse a DDL file and return a map of table names to bigdecimal columns
 ///
 /// # Panics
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist for #560 triage (not a new claim):

I checked PR #2278 at head `65336d2f483cf421208141cccbb7c29a9b5a058c`. This is a narrow one-file change in `crates/proof-of-sql/src/utils/parse.rs`: it factors the existing `DECIMAL(p, s)` detection into `bigdecimal_spec()` and applies the same `precision > 38` rule to `NUMERIC(p, s)` and `DEC(p, s)` aliases.

Useful maintainer checks before merge:

- The scope is intentionally small: `NUMERIC(39, 0)` and `DEC(40, 2)` are newly included, while `DECIMAL(38, 0)` remains excluded by the existing threshold.
- The PR deliberately does not change the current `DECIMAL(p)` precision-only behavior; if that is still intended, this is cleanly scoped to exact-number aliases with explicit scale.
- Because `precision` and `scale` are still cast to `u8`/`i8`, the existing parser assumptions remain unchanged. If SQL dialects can surface scale outside the signed-byte range here, that should be handled separately rather than blocking this alias coverage slice.

Current gates look maintainer-side: GitHub marks the PR `OPEN` / `MERGEABLE`, but `mergeStateStatus=BLOCKED` and `reviewDecision=REVIEW_REQUIRED`; Orca checks are green, while `CI-Lint-And-Test` and `CI-Code-Coverage` are both `action_required` for this fork head. I did not see prior comments or review threads before this note, and I did not run local Cargo validation from this environment.

PR #2280, opened 2026-06-07
test: cover panic and unsupported-feature branches in SortMergeJoinExec

64johnlee:
## Summary

Adds 5 `#[should_panic]` tests to `sort_merge_join_exec_test.rs` targeting the 16 uncovered lines in `sort_merge_join_exec.rs` (currently at 96.3% coverage):

- **Constructor guard — left index out of bounds**: triggers `panic!("Join column index out of bounds")` when a left join column index ≥ number of left columns
- **Constructor guard — right index out of bounds**: symmetric case for the right table
- **Constructor guard — join column count mismatch**: triggers `assert!` when `left_join_column_indexes.len() ≠ right_join_column_indexes.len()`
- **Constructor guard — wrong result_idents count**: triggers `assert!` when `result_idents.len() ≠ left_cols + right_cols − join_cols`
- **Multi-column join not supported**: passes a 2-column join through the constructor (valid), then runs `VerifiableQueryResult::new` which hits `assert!((num_columns_u == 1), "Join on multiple columns not supported yet")` in `first_round_evaluate`

Relates to coverage issue #560.

## Test plan

- [ ] Rust CI passes (all existing tests green, new `#[should_panic]` tests pass)
- [ ] Codecov shows `sort_merge_join_exec.rs` coverage increase toward 100%

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_plans/sort_merge_join_exec_test.rs
+++ b/crates/proof-of-sql/src/sql/proof_plans/sort_merge_join_exec_test.rs
@@ -496,3 +496,127 @@ fn we_can_prove_and_get_the_correct_empty_result_from_a_sort_merge_join_if_one_o
     ]);
     assert_eq!(res, expected_res);
 }
+
+#[test]
+#[should_panic(expected = "Join column index out of bounds")]
+fn we_panic_when_left_join_column_index_is_out_of_bounds() {
+    let table_left: TableRef = "sxt.left".parse().unwrap();
+    let table_right: TableRef = "sxt.right".parse().unwrap();
+    sort_merge_join(
+        table_exec(table_left, vec![column_field("id", ColumnType::BigInt)]),
+        table_exec(table_right, vec![column_field("id", ColumnType::BigInt)]),
+        vec![1], // index 1 >= 1 column
+        vec![0],
+        vec![Ident::new("id")],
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Maintainers: quick review-assist for the #560 coverage lane.

I checked this as a one-file, test-only change in `crates/proof-of-sql/src/sql/proof_plans/sort_merge_join_exec_test.rs` (`+124/-0`). The useful review points are:

- constructor guards now have negative coverage for left/right out-of-bounds join indexes, mismatched join-column counts, and the wrong `result_idents` length;
- the multi-column join path uses valid constructor input first, then reaches `VerifiableQueryResult::new` to exercise the currently unsupported multi-column join assertion;
- the added cases are `#[should_panic]` tests only and do not change production `SortMergeJoinExec` behavior.

Current status observed at head `ed990ceeba4993fdedfb2cd6b70be7461b2a0244`: zero prior comments/reviews, Orca checks pass, combined commit statuses are empty, `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required`, and GitHub reports `MERGEABLE` but `mergeStateStatus=BLOCKED` / `reviewDecision=REVIEW_REQUIRED`. From the visible surface, the unblock looks like fork workflow approval/rerun plus maintainer review.

PR #2322, opened 2026-06-08
test: cover equals expression verifier evaluation

elianguitarra:
/claim #560

## Summary
- Adds direct verifier-level coverage for `EqualsExpr::verifier_evaluate`.
- Uses a 4-row BigInt table with both equal and non-equal row cases.
- Verifies the returned equality scalar and the generated identity results.

## Scope / Deconfliction
- File: `crates/proof-of-sql/src/sql/proof_exprs/equals_expr_test.rs`
- Refresh98 open PR metadata showed zero open PR file hits for this test file.
- Local claim ledger has no previous claim against this file.
- `/attempt #560`: https://github.com/spaceandtimefdn/sxt-proof-of-sql/issues/560#issuecomment-4644914566

## Evidence
- MP4 evidence release: https://github.com/elianguitarra/sxt-proof-of-sql/releases/tag/evidence-sxt-560-equals-expr-verifier-refresh98

## Validation
- `git diff --check`
- `cargo test -p proof-of-sql --lib --no-default-features --features test equals_expr -- --quiet` -> 2 passed, 0 failed

### DIFF
--- a/crates/proof-of-sql/src/sql/proof_exprs/equals_expr_test.rs
+++ b/crates/proof-of-sql/src/sql/proof_exprs/equals_expr_test.rs
@@ -2,16 +2,21 @@ use crate::{
     base::{
         commitment::InnerProductProof,
         database::{
-            owned_table_utility::*, table_utility::*, Column, ColumnType, OwnedTable,
+            owned_table_utility::*, table_utility::*, Column, ColumnRef, ColumnType, OwnedTable,
             OwnedTableTestAccessor, Table, TableRef, TableTestAccessor,
         },
+        map::indexmap,
         math::decimal::Precision,
-        scalar::Scalar,
+        polynomial::MultilinearExtension,
+        scalar::{test_scalar::TestScalar, Scalar},
     },
     proof_primitive::inner_product::curve_25519_scalar::Curve25519Scalar,
     sql::{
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only delta in `crates/proof-of-sql/src/sql/proof_exprs/equals_expr_test.rs` (+68/-4) that directly exercises `EqualsExpr::verifier_evaluate` rather than only the first/final-round evaluator path. The new case builds a four-row BigInt table with both equal rows (`3 == 3`, `0 == 0`) and non-equal rows (`-2 != 4`, `5 != -5`), feeds the final-round output into `run_verify_for_each_row`, and compares the verifier result against the expected equality scalar from the multilinear evaluations.

The identity assertion (`vec![vec![true, true]; 4]`) is also useful for review because it checks the verifier-side identity outputs for every row, not just the returned scalar. Current external gate state from the PR metadata is still maintainer-side: Orca checks are green, while `CI-Code-Coverage` and `CI-Lint-And-Test` are `action_required` on the fork head `1284613eedbe3e6322ddfbc91c5f1e00b2b95882`. I did not see prior conversation comments or reviews on this PR at the time of this note.

JayWhite2357:

PR #2345, opened 2026-06-08
test: cover byte log floor and powers

elianguitarra:
/claim #560

## Summary
- Adds direct coverage for `is_pow2_bytes`, including zero-as-power, true powers, and multi-bit non-powers.
- Adds direct floor coverage for `log2_down_bytes` across zero, single-byte, multibyte, and all-ones byte arrays.

## Evidence
- MP4: https://github.com/elianguitarra/sxt-proof-of-sql/releases/tag/evidence-sxt-560-log-byte-floor-pow2-refresh120
- Commit: 685cd86ef647

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features test log --target-dir C:\Users\malow\AppData\Local\Temp\sxt-target-116 -- --quiet`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/math/log.rs
+++ b/crates/proof-of-sql/src/base/math/log.rs
@@ -71,6 +71,25 @@ mod tests {
         assert_eq!(log2_up(4u32), 2);
     }
 
+    #[test]
+    fn test_is_pow2_bytes() {
+        assert!(is_pow2_bytes(&[0, 0, 0, 0]));
+        assert!(is_pow2_bytes(&[0, 0, 1, 0]));
+        assert!(is_pow2_bytes(&[128, 0, 0, 0]));
+        assert!(!is_pow2_bytes(&[1, 1, 0, 0]));
+        assert!(!is_pow2_bytes(&[3, 0, 0, 0]));
+    }
+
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist for #560:

I checked this against the current `log.rs` implementation and the patch looks intentionally narrow: it only adds tests under `#[cfg(test)]` for the existing byte-log helpers, with no production-path changes.

A couple of validation notes that may help review:
- `is_pow2_bytes(&[0, 0, 0, 0])` looking "true" is consistent with the existing doc comment: zero is explicitly treated as a power of two here to keep `log2_up_bytes(0)` from panicking.
- The new `log2_down_bytes` cases cover the zero guard, low-byte floor, byte-boundary floor, high-bit-in-a-later-byte offset, and all-ones upper bound.
- The cases complement the existing `log2_up_bytes` ceiling tests rather than just duplicating them, so this is a useful coverage slice for the little-endian byte interpretation.

No blocking issue found from static review. I could not run Cargo locally in this heartbeat because there is no local `sxt-proof-of-sql` checkout available.

JayWhite2357:

PR #2350, opened 2026-06-08
test: cover inner product ref cast

elianguitarra:
/claim #560

## Summary
- Adds direct coverage for `inner_product_ref_cast`.
- Verifies the normal equal-length path and the truncation behavior when the right-hand slice is longer.

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features test inner_product_ref_cast --target-dir C:\Users\malow\AppData\Local\Temp\sxt-target-116 -- --quiet`
- `cargo fmt --check`
- `git diff --check`

## Evidence
- MP4 evidence: https://github.com/elianguitarra/sxt-proof-of-sql/releases/tag/evidence-sxt-560-inner-product-ref-cast-refresh125
- Commit: 23720a6f
- Payout wallet EVM/Base USDC/FNDRY: `0xa3d7745af6E77ce825f0AF6DB94bA8073355E022`

### DIFF
--- a/crates/proof-of-sql/src/base/slice_ops/inner_product_test.rs
+++ b/crates/proof-of-sql/src/base/slice_ops/inner_product_test.rs
@@ -123,6 +123,20 @@ fn test_inner_product_different_lengths() {
     assert_eq!(TestScalar::from(40), inner_product(&a, &b));
 }
 
+#[test]
+fn test_inner_product_ref_cast() {
+    let a = [1, 2, 3, 4].map(TestScalar::from).to_vec();
+    let b = [2, 3, 4, 5].map(TestScalar::from).to_vec();
+    assert_eq!(TestScalar::from(40), inner_product_ref_cast(&a, &b));
+}
+
+#[test]
+fn test_inner_product_ref_cast_different_lengths() {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow, test-only delta in `crates/proof-of-sql/src/base/slice_ops/inner_product_test.rs` (+14/-0) that adds direct coverage for `inner_product_ref_cast`. The new tests cover the normal equal-length case and the different-length case where the right-hand slice has an extra element.

The useful review point is that both cases assert the same expected `TestScalar::from(40)` result for `[1, 2, 3, 4]` dotted with `[2, 3, 4, 5]`, so the added different-length test documents that `inner_product_ref_cast` follows the same truncation behavior as the adjacent `inner_product` length-mismatch test rather than consuming the extra right-hand element. Current external gate state is maintainer-side: Orca checks are green, while `CI-Code-Coverage` and `CI-Lint-And-Test` are `action_required` on fork head `23720a6f325c198917a00f9529476d8e97969e5b`. I saw only the author's `/claim #560` comment and no reviews before this note.

JayWhite2357:

PR #2356, opened 2026-06-08
test: cover naive commitment column variants

elianguitarra:
/claim #560

## Summary
- Adds focused coverage for `NaiveCommitment::compute_commitments` across additional committable column variants.
- Covers Boolean, Uint8, TinyInt, SmallInt, Int, Int128, Decimal75, Scalar, and TimestampTZ inputs.
- Verifies an offset prepends `TestScalar::ZERO` and that committed scalar values match the source data for each variant.

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features test primitive_time_and_limb_columns --target-dir C:\Users\malow\AppData\Local\Temp\sxt-target-116 -- --quiet`
- Result: 1 passed, 0 failed, 960 filtered out.
- `cargo fmt --check`
- `git diff --check`

## Evidence
- MP4 evidence: https://github.com/elianguitarra/sxt-proof-of-sql/releases/tag/evidence-sxt-560-naive-commitment-types-refresh131
- Commit: 6c6cd2e2

## Payout
- EVM/Base/USDC/FNDRY wallet: `0xa3d7745af6E77ce825f0AF6DB94bA8073355E022`

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/naive_commitment.rs
+++ b/crates/proof-of-sql/src/base/commitment/naive_commitment.rs
@@ -225,6 +225,11 @@ fn we_can_compute_commitments_from_committable_varbinary_column_with_offset() {
 
 #[cfg(test)]
 mod tests {
+    use crate::base::{
+        math::decimal::Precision,
+        posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
+    };
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist for #560:

I checked this against the current `NaiveCommitment::compute_commitments` implementation. The patch is a narrow `#[cfg(test)]` addition in `naive_commitment.rs`, and it exercises the offset path by expecting the leading `TestScalar::ZERO` before each converted column value.

Useful coverage this adds:
- covers several primitive branches plus limb-backed `Decimal75`/`Scalar` and `TimestampTZ`;
- validates that offset padding is applied consistently across each returned commitment, not only the older `BigInt`/`VarBinary` examples;
- keeps the expectation explicit by comparing each raw commitment vector instead of only checking lengths or transcript bytes.

One non-blocking follow-up gap: this still leaves `VarChar`/`VarBinary` coverage in the existing tests rather than this combined variant matrix, so I would treat this as complementary coverage rather than a full replacement for the older tests.

No blocking issue found from static review. I could not run Cargo locally in this heartbeat because there is no local `sxt-proof-of-sql` checkout available.

JayWhite2357:

PR #2433, opened 2026-06-08
test: cover empty wrapped transcript binding

elianguitarra:
/claim #560

## Summary
- Adds test-only coverage for empty `wrap_transcript` closures on `Keccak256Transcript`.
- Verifies two transcripts with an empty wrapped transcript derive matching challenges.
- Verifies that wrapped transcripts still differ from an untouched transcript, proving the outer transcript was bound even when the wrapped closure is a no-op.

## Evidence
- MP4 evidence: https://github.com/elianguitarra/sxt-proof-of-sql/releases/tag/evidence-sxt-560-empty-wrapped-transcript-refresh208
- Attempt marker: https://github.com/spaceandtimefdn/sxt-proof-of-sql/issues/560#issuecomment-4651191315

## Validation
- `cargo test -p proof-of-sql --lib --no-default-features --features test empty_wrapped_transcripts_are_still_bound_to_the_outer_transcript -- --quiet`
- `cargo test -p proof-of-sql --lib --no-default-features --features test transcript::tests -- --quiet`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/proof-of-sql/src/base/proof/transcript.rs
+++ b/crates/proof-of-sql/src/base/proof/transcript.rs
@@ -141,6 +141,22 @@ mod tests {
     }
 
     #[test]
+    fn empty_wrapped_transcripts_are_still_bound_to_the_outer_transcript() {
+        let mut untouched_transcript: Keccak256Transcript = Transcript::new();
+        let mut transcript1: Keccak256Transcript = Transcript::new();
+        let mut transcript2: Keccak256Transcript = Transcript::new();
+
+        transcript1.wrap_transcript(|_transcript: &mut merlin::Transcript| ());
+        transcript2.wrap_transcript(|_transcript: &mut merlin::Transcript| ());
+
+        assert_eq!(transcript1.challenge_as_le(), transcript2.challenge_as_le());
+        assert_ne!(
+            transcript1.challenge_as_le(),
+            untouched_transcript.challenge_as_le()
+        );
+    }
+
+    #[test]
     fn we_can_extend_transcript_with_extend_as_be_from_refs() {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

- Current visible state: `OPEN` / `MERGEABLE`, with `mergeStateStatus=BLOCKED` and `reviewDecision=REVIEW_REQUIRED` at head `4e3925a5613136b5b5c16bc2ebfb0533ec13aa8e`.
- CI visibility: Orca IaC/Secrets/Vulnerabilities are green. `CI-Lint-And-Test` and `CI-Code-Coverage` both show `conclusion=action_required`; their jobs lists are empty, which looks like the usual fork workflow approval gate rather than a code failure.
- Review surface: this is a one-file, test-only slice in `crates/proof-of-sql/src/base/proof/transcript.rs` (`+16/-0`). It verifies that an empty `wrap_transcript` closure is still bound into the outer `Keccak256Transcript`: two identically wrapped transcripts match, while a wrapped transcript differs from an untouched transcript.
- Pre-merge check: if `challenge_as_le()` advances transcript state, consider caching the first `transcript1` challenge before the `assert_eq!` and comparing that cached value against the untouched transcript. As written, the second assertion calls `transcript1.challenge_as_le()` again, so it may be comparing the next challenge rather than the same bound challenge checked in the equality assertion. If the API intentionally supports repeated challenge draws here, the current shape is fine; otherwise caching would make the test intent sharper.

Claim hygiene: PR body and the first comment both carry `/claim #560`; I see no #557 overlap in this slice.

JayWhite2357:

PR #2438, opened 2026-06-08
refactor: make Scalar limb conversions explicit

### LINKED ISSUE #228: Make opaque `Scalar` conversions explicit rather than using `From`/`Into` trait.

SINISTER1999:

# Rationale for this change

Fixes #228.

This PR addresses the limb-conversion portion of the issue by making scalar limb conversions explicit on the `Scalar` trait. The goal is to replace implicit limb conversions through generic trait bounds with explicit conversion methods.

- Adds `Scalar::from_limbs`.
- Adds `Scalar::to_limbs`.
- Removes the `Into<[u64; 4]>`, `From<[u64; 4]>`, and `RefInto<[u64; 4]>` bounds from `Scalar`.
- Updates generic call sites to use the explicit methods.
- Keeps the existing `MontScalar` `From` impls for compatibility with existing concrete-code call sites.

# Are these changes tested?

Partially.

Verification completed:
- `git diff --check`

I could not run `cargo check -p proof-of-sql`, `source scripts/run_ci_checks.sh`, or `source scripts/check_commits.sh` in my local environment because `cargo` is not installed there.

I am happy to adjust the implementation if CI or maintainers identify required changes.

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/scalar_and_i256_conversions.rs
+++ b/crates/proof-of-sql/src/base/arrow/scalar_and_i256_conversions.rs
@@ -14,7 +14,7 @@ const MAX_SUPPORTED_I256: i256 = i256::from_parts(
 pub fn convert_scalar_to_i256<S: Scalar>(val: &S) -> i256 {
     let is_negative = val > &S::MAX_SIGNED;
     let abs_scalar = if is_negative { -*val } else { *val };
-    let limbs: [u64; 4] = abs_scalar.into();
+    let limbs = abs_scalar.to_limbs();
 
     let low = u128::from(limbs[0]) | (u128::from(limbs[1]) << 64);
     let high = i128::from(limbs[2]) | (i128::from(limbs[3]) << 64);
@@ -47,7 +47,7 @@ pub fn convert_i256_to_scalar<S: Scalar>(value: &i256) -> Option<S> {
         ];
 
         // Convert limbs to Scalar and adjust for sign
-        let scalar: S = limbs.into();
+        let scalar = S::from_limbs(limbs);
         Some(if value.is_negative() { -scalar } else { scalar })
     }
 }
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -3,7 +3,6 @@ use crate::base::{
     if_rayon,
     math::decimal::Precision,
     posql_time::{PoSQLTimeUnit, PoSQLTimeZone},
-    ref_into::RefInto,
     scalar::{Scalar, ScalarExt},
 };
 use alloc::vec::Vec;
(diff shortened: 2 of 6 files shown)


MyTH-zyxeon:
Quick review from the #228 scalar-conversion angle:

This is a focused limb-conversion slice: it removes the generic `Scalar: Into<[u64; 4]> + From<[u64; 4]> + RefInto<[u64; 4]>` requirements, adds explicit `Scalar::from_limbs` / `Scalar::to_limbs`, updates the generic call sites, and keeps the concrete `MontScalar` `From` impls as compatibility shims. That lines up with the issue direction without pulling in the string-hash subtasks.

Maintainer checks I would use before approving:

- CI is still fork-gated on head `b3330c4d7a264e651b4a4a667b8a8ec03162576f`: Orca is green, but `CI-Lint-And-Test` and `CI-Code-Coverage` are `action_required`, both runs expose `jobs=[]`, and the combined commit status returns `statuses=[]`. This looks like approval/rerun needed rather than a proven code failure.
- The compatibility decision is worth confirming: `MontScalar` still implements `From<MontScalar>` and `From<&MontScalar>` for `[u64; 4]`, so existing concrete-code conversions continue to compile while generic callers are forced through the explicit trait methods.
- One small code-review point: `range_check.rs` now uses `Into::<S>::into(*scalar).to_limbs()` even though `scalar` is already an `&S`; `scalar.to_limbs()` should be equivalent and simpler if there is no type-inference reason for the extra conversion.

No claim action from me; this just scopes why #228 can review this as a narrow explicit-limb-conversion implementation slice once the fork-gated workflows are approved.

SINISTER1999:
Thanks for the careful review. That makes sense.

I’ll check the `range_check.rs` call site and simplify it to
`scalar.to_limbs()` if there is no inference issue. I’ll push a small
follow-up commit once verified.

The fork-gated CI note also matches what I’m seeing: the Vercel/GitHub
Actions checks appear to need maintainer approval rather than failing on
the patch itself.

SINISTER1999:
Thanks for the review. I simplified the `range_check.rs` call site by replacing the explicit `Into::<S>::into(*scalar)` call with `(*scalar).into()`.

I kept the conversion before `to_limbs()` because this helper is generic over `T: Copy + Into<S>`, so the loop variable is `&T` rather than `&S`.

MyTH-zyxeon:
Follow-up after the author response on the `range_check.rs` point:

I rechecked head `349e7e03c82fac071dddb5f7e11fe9df5400dd64`. The author did push the intended cleanup: the range-check limb extraction now uses `(*scalar).into().to_limbs()` instead of the more explicit `Into::<S>::into(*scalar).to_limbs()`. Given the helper iterates over `&T` where `T: Copy + Into<S>`, this keeps the required `T -> S` conversion while removing the noisy fully qualified form; it is not the same as `scalar.to_limbs()` because the loop item is not already `&S`.

Current review state:
- PR metadata remains `OPEN` / `MERGEABLE` / `mergeStateStatus=BLOCKED` / `reviewDecision=REVIEW_REQUIRED`.
- Visible Orca checks are green.
- The two GitHub Actions suites for the current head are still `action_required` with `latest_check_runs_count=0`, so the Rust validation remains maintainer-approval gated.
- Diff remains a narrow #228 limb-conversion slice: `Scalar::from_limbs` / `Scalar::to_limbs`, explicit conversions in Arrow/commitment/scalar-ext paths, and compatibility `MontScalar` `From` impls routed through `to_limbs`.

So the earlier code-review concern about the overly explicit `Into::<S>::into(...)` form is resolved. The remaining acceptance checks are maintainer workflow approval plus comparison against the broader #228 candidates before choosing the canonical landing path.

No competing claim action from me; this is only status/routing help for the existing `/claim #228` PR.

JayWhite2357:

PR #2456, opened 2026-06-12
test: cover RefInto reference conversion

Shreshth3:
## Summary
- Adds a focused unit test for `RefInto`
- Verifies the blanket implementation uses `From<&T>` without consuming the original value

/claim #560

## Validation
- `git diff --check`

I could not run `cargo test` locally because this environment does not have `cargo` or `rustc` installed.

### DIFF
--- a/crates/proof-of-sql/src/base/ref_into.rs
+++ b/crates/proof-of-sql/src/base/ref_into.rs
@@ -23,3 +23,25 @@ where
         self.into()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::RefInto;
+
+    #[derive(Debug, PartialEq, Eq)]
+    struct Wrapper(u8);
+
+    impl From<&Wrapper> for u16 {
+        fn from(value: &Wrapper) -> Self {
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:

This is a narrow `/claim #560` test-only PR in `crates/proof-of-sql/src/base/ref_into.rs` (`+22/-0`). It adds a local `Wrapper(u8)` test fixture, implements `From<&Wrapper> for u16`, then verifies `RefInto::ref_into()` uses the reference conversion and leaves the original value comparable afterward.

The main review point I would check is whether this unit-test placement is the intended home for `RefInto` blanket-impl coverage versus a nearby shared conversion/test module. The assertion pair does cover both the return conversion and non-consuming behavior, so it looks scoped to the claim without changing production code.

I did not run the Rust test suite locally. GitHub currently shows the security scan check runs succeeded, while `CI-Code-Coverage` and `CI-Lint-And-Test` are `action_required` for the head commit rather than completed pass/fail results.

JayWhite2357:

PR #2461, opened 2026-06-13
ci: parallelize Foundry and Slither checks

absalonCRC:

- [x] The PR title and commit message adhere to the conventional commit guidelines.
- [x] I have run the clean commit check script with `source scripts/check_commits.sh`.
- [x] The latest changes from `main` have been incorporated to this PR.

# Rationale for this change

Issue #557 asks for test and CI runtime improvements without removing or weakening tests. The Foundry CI job currently runs four independent Forge test modes serially, and later runs three independent Slither analyses serially.

This PR keeps the same test and analyzer coverage, but reduces wall-clock time by running independent work concurrently inside the existing `foundrycheck` job:

- preprocess Solidity sources once before the Forge test modes;
- run the four existing Forge test modes in parallel;
- give each Forge mode its own artifact and compiler-cache directories via `--out` and `--cache-path` to avoid shared output/cache races;
- cap each Forge mode at `--threads 2` so the four modes fit the existing 8-core runner instead of oversubscribing it;
- run the three existing Slither target analyses in parallel and preserve failure propagation.

/claim #557

- Replace the four serial Foundry test steps with one parallel `Run Foundry test modes` step.
- Replace the three serial Slither steps with one parallel `Run Slither checks` step.
- Keep coverage, formatting, solhint, dependency installation, and the Rust EVM job unchanged.

# Are these changes tested?

Local/static validation:

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/lint-and-test.yml"); puts "yaml ok"'`
- `python3 - <<'PY' ... bash -n -c step['run'] ... PY` for the two changed shell blocks
- `git diff --check`

I did not complete `source scripts/run_ci_checks.sh` locally. This is a workflow-only Linux CI change, and this local macOS/Apple Silicon environment does not match the Linux runner/toolchain assumptions. I validated the changed YAML and shell blocks directly and left full runtime validation to GitHub Actions / maintainer CI.

Disclosure: prepared with AI assistance in Codex. I reviewed the workflow behavior, checked for overlap with existing #557 PRs, and validated the changed YAML/shell blocks before submission.

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -288,14 +288,41 @@ jobs:
           version: v1.0.0
       - name: Install dependencies
         run: solidity/scripts/install_deps.sh
-      - name: Run tests without via-ir or optimization
-        run: solidity/scripts/pre_forge.sh test
-      - name: Run tests with via-ir but without optimization
-        run: solidity/scripts/pre_forge.sh test --via-ir
-      - name: Run tests with optimization but without via-ir
-        run: solidity/scripts/pre_forge.sh test --optimize
-      - name: Run tests with via-ir and optimization
-        run: solidity/scripts/pre_forge.sh test --via-ir --optimize
+      - name: Run Foundry test modes
+        run: |
+          set -euo pipefail
+          cd solidity
+
+          find . -type f -name '*.post.sol' -delete
+          python3 preprocessor/yul_preprocessor.py .
+
+          pids=()
+          run_mode() {
+            local label="$1"
+            local out_dir="$2"
+            local cache_dir="$3"
+            shift 3
+
+            echo "Running Foundry tests: ${label}"
+            forge test --threads 2 --out "${out_dir}" --cache-path "${cache_dir}" "$@"
+          }
+
+          run_mode "without via-ir or optimization" out-test-default cache-test-default &
+          pids+=($!)
+          run_mode "with via-ir but without optimization" out-test-via-ir cache-test-via-ir --via-ir &
+          pids+=($!)
+          run_mode "with optimization but without via-ir" out-test-optimize cache-test-optimize --optimize &
+          pids+=($!)
+          run_mode "with via-ir and optimization" out-test-via-ir-optimize cache-test-via-ir-optimize --via-ir --optimize &
+          pids+=($!)
+
+          failed=0
+          for pid in "${pids[@]}"; do
+            if ! wait "${pid}"; then
+              failed=1
+            fi
+          done
+          exit "${failed}"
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist for `/claim #557`:

I checked the workflow delta against the current `foundrycheck` job and `solidity/scripts/pre_forge.sh`. The PR keeps the same four Forge modes and the same three Slither targets, but moves the independent work into parallel shell blocks. The manual Foundry setup mirrors the existing `pre_forge.sh` behavior (`cd solidity`, delete `*.post.sol`, run the Yul preprocessor) and adds separate `--out` / `--cache-path` directories per mode, which is the key race-avoidance detail for running the four modes concurrently.

Useful maintainer checks before merge:
- approve/rerun the forked GitHub Actions jobs, because the visible `CI-Lint-And-Test` and `CI-Code-Coverage` runs are still `action_required` on head `4f464a625478b9b316d9750b63dad8958315e4ab`;
- confirm the 8-core runner has enough headroom for four `forge test --threads 2` processes plus compiler memory pressure;
- confirm Slither writes no shared cache/output that would make the three target analyses race when run in parallel.

No blocking issue found from static review. I could not run the Linux CI locally because there is no local `sxt-proof-of-sql` checkout available in this environment, and this is a runner-specific workflow change.

No competing claim action from me; this is only review routing for the existing #557 claim.

absalonCRC:
I see the full `CI-Lint-And-Test` and `CI-Code-Coverage` workflow runs are currently `action_required` on the forked PR head (`4f464a625478b9b316d9750b63dad8958315e4ab`).

Once a maintainer approves/reruns those workflows, I’ll follow up on any failures. The intent of this change is to keep the same Forge and Slither coverage while only reducing wall-clock time through isolated parallel jobs.

MyTH-zyxeon:
Follow-up after the author response:

I rechecked head `4f464a625478b9b316d9750b63dad8958315e4ab`. The visible PR checks are still only the three Orca scans, all green, while the two GitHub Actions check suites for the head SHA are still `action_required` with `latest_check_runs_count=0`. That matches the author's note that the full `CI-Lint-And-Test` / `CI-Code-Coverage` runs still need maintainer approval or rerun before the actual wall-clock/runtime effect can be measured.

The patch remains narrowly scoped to `.github/workflows/lint-and-test.yml` (`+52/-14`) and still preserves the same four Forge modes plus the three Slither targets while isolating Forge `--out` and `--cache-path` directories per parallel mode. So the remaining gate I see is not a code change request from static review; it is maintainer approval/rerun of the forked Actions suites, then review of any real runtime failures if those suites produce logs.

No competing claim action from me; this is only status/routing help for the existing `/claim #557` PR.

PR #2463, opened 2026-06-14
test: cover uppercase column visitor

mawxcodehub:
## Summary
- Add direct coverage for `uppercase_identifier` value conversion and quote preservation.
- Add a regression test for compound identifiers across qualified select columns and join tables.

## Validation
- `cargo test -p proof-of-sql-planner uppercase_column_visitor --lib`
- `cargo fmt --check`

Part of #560.

### DIFF
--- a/crates/proof-of-sql-planner/src/uppercase_column_visitor.rs
+++ b/crates/proof-of-sql-planner/src/uppercase_column_visitor.rs
@@ -47,14 +47,50 @@ pub fn statement_with_uppercase_identifiers(mut statement: Statement) -> Stateme
 
 #[cfg(test)]
 mod tests {
-    use super::statement_with_uppercase_identifiers;
-    use sqlparser::{dialect::GenericDialect, parser::Parser};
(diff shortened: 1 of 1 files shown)


MyTH-zyxeon:
Review-assist for the #560 coverage queue:

I checked this diff against the current `main` implementation. It is test-only in `crates/proof-of-sql-planner/src/uppercase_column_visitor.rs` (`+38/-2`) and covers two useful surfaces that were not directly asserted before:

- `uppercase_identifier` uppercases `Ident.value` while preserving `quote_style` via the existing `Ident { value, ..ident }` rebuild.
- `statement_with_uppercase_identifiers` handles compound identifiers across qualified select columns and joined table references, matching the visitor plus `visit_relations_mut` paths.

Current public status for head `f76569d00467bf661e5901135bddcd4052d905ba`: PR is open/mergeable, no prior comments or reviews were visible, and the raw check-runs only show the three Orca checks passing (IaC, Secrets, Vulnerabilities). I do not see a code-side blocker in this patch; the remaining gate appears to be maintainer review/acceptance.

PR #2504, opened 2026-06-22
test: cover PoSQL-compatible Arrow schema conversion

alexcvr:
## Summary
- add direct coverage for Float16, Float32, and Float64 conversion to Decimal256(20, 10)
- verify field nullability is preserved during conversion
- verify non-floating-point fields remain unchanged

## Testing
- isolated Arrow 51 harness: 2 passed, 0 failed
- the full workspace cannot build on Windows because blitzar-sys supports Linux only; project CI can provide the Linux validation

/claim #560

### DIFF
--- a/crates/proof-of-sql/src/base/database/arrow_schema_utility.rs
+++ b/crates/proof-of-sql/src/base/database/arrow_schema_utility.rs
@@ -35,3 +35,39 @@ pub fn get_posql_compatible_schema(schema: &SchemaRef) -> SchemaRef {
 
     Arc::new(Schema::new(new_fields))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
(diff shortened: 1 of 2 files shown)


alexcvr:
Added a second non-overlapping coverage slice for `Table` accessors and inner views.

Validation run locally on Windows with Linux-only `blitzar` disabled:

```text
cargo +stable-x86_64-pc-windows-gnu test -p proof-of-sql --no-default-features --features arrow,std table_ --offline
```

Result: 58 passed, 0 failed, including the new `table_accessors_preserve_column_order_and_types` and `table_inner_views_expose_the_same_ordered_columns` tests.

Full default-feature test is still blocked locally because `blitzar-sys` only supports Linux in its build script, so CI/Linux should be the source of truth for default features.

PR #2507, opened 2026-06-23
feat: add `IN` operator via product-form `InExpr`

rishitesh-snt:
# Rationale for this change

SQL `IN` / `NOT IN` weren't supported by the planner. This adds them by lowering `Expr::InList` into expressions we already have, so there's no new proof expr.

`a IN (v1, …, vk)` is lowered by operand type:
- **numeric** → `(a - v1) * … * (a - vk) = 0` (product is zero iff one factor is)
- **varchar** → `a = v1 OR … OR a = vk`
- **empty** `IN ()` → `false`
- **`NOT IN`** → the above wrapped in `NOT`

# Are these changes tested?

Yes — planner unit tests (product/or-chain/empty/negation) and end-to-end prove/verify for numeric and varchar `IN`/`NOT IN`.

### DIFF
--- a/crates/proof-of-sql-planner/src/expr.rs
+++ b/crates/proof-of-sql-planner/src/expr.rs
@@ -3,13 +3,16 @@ use super::{
     PlannerError, PlannerResult,
 };
 use datafusion::logical_expr::{
-    expr::{Alias, Cast, Placeholder},
+    expr::{Alias, Cast, InList, Placeholder},
     BinaryExpr, Expr, Operator,
 };
 use indexmap::IndexSet;
 use proof_of_sql::{
-    base::database::ColumnType,
-    sql::{proof_exprs::DynProofExpr, scale_cast_binary_op},
+    base::database::{ColumnType, LiteralValue},
+    sql::{
+        proof_exprs::{DynProofExpr, ProofExpr},
+        scale_cast_binary_op,
+    },
 };
 use sqlparser::ast::Ident;
 
@@ -27,6 +30,13 @@ pub(crate) fn get_column_idents_from_expr(expr: &Expr) -> IndexSet<Ident> {
             left_idents
         }
         Expr::Not(inner) => get_column_idents_from_expr(inner),
+        Expr::InList(InList { expr, list, .. }) => {
+            let mut idents = get_column_idents_from_expr(expr);
+            for value in list {
+                idents.extend(get_column_idents_from_expr(value));
+            }
+            idents
+        }
         Expr::Alias(Alias { expr, .. }) | Expr::Cast(Cast { expr, .. }) => {
             get_column_idents_from_expr(expr)
         }
@@ -119,10 +129,12 @@ fn binary_expr_to_proof_expr(
     }
 }
 
-/// Convert an [`datafusion::expr::Expr`] to [`DynProofExpr`]
+/// Convert a `DataFusion` [`Expr`] into a provable [`DynProofExpr`], using `schema`
+/// to resolve column references and their types.
 ///
-/// # Panics
-/// The function should not panic if Proof of SQL is working correctly
+/// # Errors
+/// Returns a [`PlannerError`] if the expression (or any subexpression) is unsupported,
+/// references an unknown column, or cannot be lowered to a `DynProofExpr`.
 pub fn expr_to_proof_expr(
     expr: &Expr,
     schema: &[(Ident, ColumnType)],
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -141,6 +151,59 @@ pub fn expr_to_proof_expr(
...
+        }) => {
+            // Lower `a IN (v_1, ..., v_k)`. The list length is uncapped in both branches.
+            //
+            //   - numeric `a`: use the product form `(a - v_1) * ... * (a - v_k) = 0`.
+            //     A field is an integral domain, so the product is zero iff some factor is.
+            //   - non-numeric `a` (e.g. varchar): the typed `-` operator rejects these, so
+            //     fall back to `a = v_1 OR ... OR a = v_k`. Equality subtracts the embedded
+            //     hashes internally, which is exactly what a membership zero-test needs.
+            let needle = expr_to_proof_expr(expr, schema)?;
+            let comparison = if list.is_empty() {
```
stuarttimwhite: A nice way to handle this sort of pattern is something like
```
let Some(expr) = list.first() else{
    return DynProofExpr::new_literal(LiteralValue::Boolean(false))
}
...
list.fold(expr, |acc, new_expr| {
    ...
})
```
The above isn't quite right, so you'd have to fiddle around with it a little.

I think the Rust sort of approach to consolidating a collection of objects into one object is to use the `fold` or `reduce` function, or something similar.

rishitesh-snt: Fixed it

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -141,6 +151,62 @@ pub fn expr_to_proof_expr(
...
+            let comparison = match list.split_first() {
+                // `a IN ()` is always false.
+                None => DynProofExpr::new_literal(LiteralValue::Boolean(false)),
+                Some((first, rest)) if needle.data_type().is_numeric() => {
+                    // Lower each `v_i` into the factor `a - v_i`, aligning operand
+                    // types/scales the same way `-` does before subtracting.
+                    let factor = |value| {
+                        let value_proof_expr = expr_to_proof_expr(value, schema)?;
+                        let (needle_cast, value_cast) =
+                            scale_cast_binary_op(needle.clone(), value_proof_expr)?;
```
stuarttimwhite: This particular block of code is repeated in both branches of logic. What you can do up above is something like:
```
list.map(|value| -> {
    let value_proof_expr = expr_to_proof_expr(value, schema)?;
    scale_cast_binary_op(needle.clone(), value_proof_expr)
}).try_collect()?.split_first()
...
```
Again, I probably haven't got the types quite right, but you get the idea.

rishitesh-snt: Changed the code

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -119,10 +129,30 @@ fn binary_expr_to_proof_expr(
...
+    first: DynProofExpr,
+    rest: impl Iterator<Item = Result<DynProofExpr, E>>,
+    combine: impl Fn(DynProofExpr, DynProofExpr) -> Result<DynProofExpr, E>,
+) -> Result<DynProofExpr, E> {
+    let mut acc = first;
+    for term in rest {
+        acc = combine(acc, term?)?;
+    }
+    Ok(acc)
+}
```
JayWhite2357: This is functionally identical to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_fold.

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -141,6 +171,59 @@ pub fn expr_to_proof_expr(
...
+                .collect::<PlannerResult<Vec<_>>>()?;
+            // Split off the first pair. `None` means `a IN ()`, which is always false;
+            // keeping that here (not an early return) lets the `negated` wrap below
+            // still apply, so `a NOT IN ()` correctly negates to `true`.
+            let comparison = match operands.split_first() {
+                None => DynProofExpr::new_literal(LiteralValue::Boolean(false)),
+                Some(((n, v), rest)) if needle.data_type().is_numeric() => {
+                    // product form: (a - v_1) * ... * (a - v_k) = 0
+                    let first = DynProofExpr::try_new_subtract(n.clone(), v.clone())?;
+                    let rest = rest
```
JayWhite2357: NIT: I try to avoid shadowing.
See https://rust-lang.github.io/rust-clippy/stable/index.html?search=shadow_unrelated#shadow_unrelated.

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -141,6 +171,59 @@ pub fn expr_to_proof_expr(
...
+            let needle = expr_to_proof_expr(expr, schema)?;
+            // Lower each `v_i` once, aligning its type/scale to the needle the same
+            // way `-`/`=` do. Both branches below consume these `(needle, v_i)` pairs.
+            let operands = list
+                .iter()
+                .map(|value| -> PlannerResult<_> {
+                    let value_proof_expr = expr_to_proof_expr(value, schema)?;
+                    Ok(scale_cast_binary_op(needle.clone(), value_proof_expr)?)
+                })
+                .collect::<PlannerResult<Vec<_>>>()?;
```
JayWhite2357: Avoid `collect`ing. It shouldn't be needed here.

Instead, you could apply `split_first` to `list`.

crates/proof-of-sql-planner/src/expr.rs
stuarttimwhite: This function is effectively the same as the `reduce` function. We can replace it with that.

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -141,6 +153,60 @@ pub fn expr_to_proof_expr(
...
+            //     fall back to `a = v_1 OR ... OR a = v_k`. Equality subtracts the embedded
+            //     hashes internally, which is exactly what a membership zero-test needs.
+            let needle = expr_to_proof_expr(expr, schema)?;
+            // Lower each `v_i` once, aligning its type/scale to the needle the same
+            // way `-`/`=` do. Both branches below consume these `(needle, v_i)` pairs.
+            let operands = list
+                .iter()
+                .map(|value| -> PlannerResult<_> {
+                    let value_proof_expr = expr_to_proof_expr(value, schema)?;
+                    Ok(scale_cast_binary_op(needle.clone(), value_proof_expr)?)
```
JayWhite2357: It doesn't make sense to apply `scale_cast_binary_op` to non-numeric types.
I don't even like the fact that `scale_cast_binary_op` accepts non-numeric types.

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -141,6 +153,60 @@ pub fn expr_to_proof_expr(
...
+                })
+                .collect::<PlannerResult<Vec<_>>>()?;
+            // Split off the first pair. `None` means `a IN ()`, which is always false;
+            // keeping that here (not an early return) lets the `negated` wrap below
+            // still apply, so `a NOT IN ()` correctly negates to `true`.
+            let comparison = match operands.split_first() {
+                None => DynProofExpr::new_literal(LiteralValue::Boolean(false)),
+                Some(((n, v), rest)) if needle.data_type().is_numeric() => {
```
JayWhite2357: NIT: This is mostly my preference, but I would prefer an `if-else` statement rather than an `if` guard here. That's because the "`else`" in the next branch is implicit, and I would rather it be explicit. Also, rearranging these match arms would make this incorrect, which always feels a bit fragile to me.

JayWhite2357 (changes requested):

crates/proof-of-sql-planner/src/expr.rs
rishitesh-snt: This function is removed now

crates/proof-of-sql-planner/src/expr.rs
rishitesh-snt: Fixed it in an earlier commit

crates/proof-of-sql-planner/src/expr.rs
rishitesh-snt: Fixed it for non-numeric tyoes

crates/proof-of-sql-planner/src/expr.rs
rishitesh-snt: Changed the flow of statement to mirror the above.

crates/proof-of-sql-planner/src/expr.rs
rishitesh-snt: Fixed

JayWhite2357 (approved):

PR #2615, opened 2026-06-26
feat: Changes to support BETWEEN operator using naive approach

sachin-snt:
# Rationale for this change

  SQL `BETWEEN` was not supported by the planner. This adds them by lowering `Expr::Between` into expressions we already have (inequality, `OR`, `NOT`), so there's no new proof expr.

  `a BETWEEN low AND high` is lowered to `NOT (a < low OR a > high)` (and `a NOT BETWEEN low AND high` to `a < low OR a > high`, without the outer `NOT`):
  - Added an `Expr::Between` arm to `expr_to_proof_expr`, handled by a new `between_to_proof_expr` helper.
  - Extracted the scale-casting binary-op logic out of `binary_expr_to_proof_expr` into a new `binary_proof_exprs_to_proof_expr`, so `between_to_proof_expr` can build the `<`/`>`/`OR` chain from already-lowered
  `DynProofExpr`s without going back through `Expr`.
  - Added a `Between` case to `get_column_idents_from_expr` so columns referenced in the expr, low, and high bounds are all picked up.

  # Are these changes tested?

  Yes — planner unit tests (`BETWEEN`, `NOT BETWEEN`, column-ident extraction) and end-to-end prove/verify tests covering inclusive boundaries and `BETWEEN` combined with `AND`/`OR` filters.

### DIFF
--- a/crates/proof-of-sql-planner/src/expr.rs
+++ b/crates/proof-of-sql-planner/src/expr.rs
@@ -3,7 +3,7 @@ use super::{
     PlannerError, PlannerResult,
 };
 use datafusion::logical_expr::{
-    expr::{Alias, Cast, InList, Placeholder},
+    expr::{Alias, Between, Cast, InList, Placeholder},
     BinaryExpr, Expr, Operator,
 };
 use indexmap::IndexSet;
@@ -45,15 +45,19 @@ pub(crate) fn get_column_idents_from_expr(expr: &Expr) -> IndexSet<Ident> {
             .iter()
             .flat_map(get_column_idents_from_expr)
             .collect(),
+        Expr::Between(Between {
+            expr, low, high, ..
+        }) => {
+            let mut idents = get_column_idents_from_expr(expr);
+            idents.extend(get_column_idents_from_expr(low));
+            idents.extend(get_column_idents_from_expr(high));
+            idents
+        }
         _ => IndexSet::new(),
     }
 }
 
 /// Convert a [`BinaryExpr`] to [`DynProofExpr`]
-#[expect(
-    clippy::missing_panics_doc,
-    reason = "Output of comparisons is always boolean"
-)]
 fn binary_expr_to_proof_expr(
     left: &Expr,
     right: &Expr,
@@ -62,7 +66,22 @@ fn binary_expr_to_proof_expr(
 ) -> PlannerResult<DynProofExpr> {
     let left_proof_expr = expr_to_proof_expr(left, schema)?;
     let right_proof_expr = expr_to_proof_expr(right, schema)?;
+    binary_proof_exprs_to_proof_expr(left_proof_expr, right_proof_expr, op)
+}
 
+/// Apply a binary [`Operator`] to two already-converted [`DynProofExpr`]s.
+///
+/// Scale-casting is performed here for operators that require it, so callers
+/// that already hold `DynProofExpr`s can skip the `Expr` conversion step.
+#[expect(
+    clippy::missing_panics_doc,
+    reason = "Output of comparisons is always boolean"
+)]
+fn binary_proof_exprs_to_proof_expr(
+    left_proof_expr: DynProofExpr,
+    right_proof_expr: DynProofExpr,
+    op: Operator,
+) -> PlannerResult<DynProofExpr> {
     let (left_proof_expr, right_proof_expr) = match op {
         Operator::Eq
         | Operator::NotEq
@@ -124,7 +143,6 @@ fn binary_expr_to_proof_expr(
             left_proof_expr,
             right_proof_expr,
         )?),
-        // Any other operator is unsupported
         _ => Err(PlannerError::UnsupportedBinaryOperator { op }),
     }
 }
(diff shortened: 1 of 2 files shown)


crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -119,7 +125,13 @@ fn binary_expr_to_proof_expr(
     }
 }
 
-/// Convert an [`datafusion::expr::Expr`] to [`DynProofExpr`]
+/// Convert a [`datafusion::logical_expr::Expr`] to [`DynProofExpr`].
+///
+/// `BETWEEN low AND high` is spit to `expr >= low AND expr <= high`,
+/// implemented as `NOT(expr < low) AND NOT(expr > high)` since the proof
```
stuarttimwhite: NIT: I think `NOT(expr < low OR expr > high)` is slightly more compact.

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -165,6 +177,46 @@ pub fn expr_to_proof_expr(
...
+            high,
+        }) => {
+            let expr_proof = expr_to_proof_expr(expr, schema)?;
+            let low_proof = expr_to_proof_expr(low, schema)?;
+            let high_proof = expr_to_proof_expr(high, schema)?;
+
+            let (expr_for_low, scaled_low) =
+                scale_cast_binary_op(expr_proof.clone(), low_proof)?;
+            let (expr_for_high, scaled_high) =
+                scale_cast_binary_op(expr_proof, high_proof)?;
```
stuarttimwhite: I think it might be a good idea to modify `binary_expr_to_proof_expr` to accept `DynProofExpr`s instead of datafusion `Expr`s. That could reduce the logic here by a decent amount. There are three or four binary exprs used here, and ideally we don't duplicate logic here.

crates/proof-of-sql-planner/src/expr.rs
sachin-snt: I think the comment matches the actual code structure , the code builds two separate NOT nodes joined by AND, not a single NOT over an OR. Changing it would make the comment inaccurate. refer line 203–217 . If you want i can add more detailed comment

stuarttimwhite: I would recommend changing both the comment and the code structure to make it more like `NOT(expr < low OR expr > high)`. It will make the resulting proof plan a little smaller. This has the added benefit in allowing the two branches of `if negated` to be more similar.

stuarttimwhite (dismissed):

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -35,91 +35,87 @@ pub(crate) fn get_column_idents_from_expr(expr: &Expr) -> IndexSet<Ident> {
...
-        Operator::Or => Ok(DynProofExpr::try_new_or(left_proof_expr, right_proof_expr)?),
-        Operator::Multiply => Ok(DynProofExpr::try_new_multiply(
-            left_proof_expr,
-            right_proof_expr,
-        )?),
-        Operator::Eq => Ok(DynProofExpr::try_new_equals(
-            left_proof_expr,
-            right_proof_expr,
-        )?),
+        Operator::And => Ok(DynProofExpr::try_new_and(left, right)?),
```
JayWhite2357: Renaming `left_proof_expr` and `right_proof_expr` to `left` and `right` make this diff substantially larger and difficult to read.

If you want to do the rename, make it a separate commit.

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -35,91 +35,87 @@ pub(crate) fn get_column_idents_from_expr(expr: &Expr) -> IndexSet<Ident> {
...
+fn binary_expr_to_proof_expr(
+    left: &Expr,
+    right: &Expr,
+    op: Operator,
+    schema: &[(Ident, ColumnType)],
+) -> PlannerResult<DynProofExpr> {
+    let left_proof_expr = expr_to_proof_expr(left, schema)?;
+    let right_proof_expr = expr_to_proof_expr(right, schema)?;
+    binary_proof_exprs_to_proof_expr(left_proof_expr, right_proof_expr, op)
+}
```
JayWhite2357: If you move this function above `binary_proof_exprs_to_proof_expr`, this diff will become substantially smaller and more readable.

You can move it back below in a separate commit if you think this position makes more sense (I'm indifferent).

JayWhite2357 (changes requested):

crates/proof-of-sql-planner/src/expr.rs
```diff
@@ -114,12 +133,17 @@ fn binary_expr_to_proof_expr(
...
         )?),
-        // Any other operator is unsupported
         _ => Err(PlannerError::UnsupportedBinaryOperator { op }),
     }
 }
 
-/// Convert an [`datafusion::expr::Expr`] to [`DynProofExpr`]
+/// Convert a [`datafusion::logical_expr::Expr`] to [`DynProofExpr`].
+///
+/// `BETWEEN low AND high` is implemented as `NOT(expr < low OR expr > high)`
```
JayWhite2357: Comments about the between don't belong here

JayWhite2357 (approved):

PR #2622, opened 2026-06-26
test: cover Arrow schema utility

xingdi614-alt:
/claim #560

## Summary

Adds a focused test-only coverage slice for `base::database::arrow_schema_utility`.

This PR verifies that `get_posql_compatible_schema`:

- converts `Float16`, `Float32`, and `Float64` fields to `Decimal256(20, 10)`
- preserves non-float Arrow field data types
- preserves field names and nullability for both converted and unchanged fields

It also corrects the function doc comment so it matches the current implementation’s `Decimal256(20, 10)` behavior.

## Non-overlap

This slice is limited to:

- `crates/proof-of-sql/src/base/database/arrow_schema_utility.rs`
- `crates/proof-of-sql/src/base/database/arrow_schema_utility_test.rs`
- the sibling test-module registration in `crates/proof-of-sql/src/base/database/mod.rs`

Before choosing this slice, I checked the visible #560 attempts and avoided the recently claimed areas around `zigzag`, `sql/proof/proof_plan.rs`, `sql/evm_proof_plan/proof_plan.rs`, `sql/proof_exprs/table_expr.rs`, and `sql/proof_plans/test_utility.rs`.

## Validation

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test -p proof-of-sql --lib --no-default-features --features arrow arrow_schema_utility -- --nocapture` — 3 passed

Note: the default feature set enables an x86_64-only `halo2curves/asm` path that fails on Apple Silicon, so the focused validation uses `--no-default-features --features arrow`, matching the narrow target of this Arrow utility coverage slice.

PR #2623, opened 2026-06-27
test: cover PoSQL timestamp errors

xingdi614-alt:
/claim #560

## Summary

Adds a focused coverage slice for `base::posql_time::PoSQLTimestampError`.

This PR verifies that timestamp errors:

- display the original invalid timezone input in the user-facing error string
- convert into `String` through the existing `From<PoSQLTimestampError> for String` implementation
- round-trip through `serde_json` for every enum variant, including the payload-carrying variants

## Non-overlap

This slice is limited to `crates/proof-of-sql/src/base/posql_time/error.rs`.

Before choosing it, I avoided the visible recent #560 attempt areas around `zigzag`, `sql/proof/proof_plan.rs`, `sql/evm_proof_plan/proof_plan.rs`, `sql/proof_exprs/table_expr.rs`, `sql/proof_plans/test_utility.rs`, and my separate Arrow schema utility PR #2622.

## Validation

- `git diff --check`
- `cargo test -p proof-of-sql --lib --no-default-features --features std posql_time::error -- --nocapture` — 3 passed

Note: the focused command uses `--features std` because this crate has existing tests and modules that require `std` when running the library test harness without default features.

### DIFF
--- a/crates/proof-of-sql/src/base/posql_time/error.rs
+++ b/crates/proof-of-sql/src/base/posql_time/error.rs
@@ -38,3 +38,46 @@ impl From<PoSQLTimestampError> for String {
         error.to_string()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::PoSQLTimestampError;
+    use alloc::{string::String, vec};
+
+    #[test]
+    fn we_display_invalid_timezone_with_the_original_input() {
+        let error = PoSQLTimestampError::InvalidTimezone {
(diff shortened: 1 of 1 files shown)


xingdi614-alt:
Closing this to keep the #560 bounty thread clean. After opening it, I found earlier #560 comments/claims covering `base/posql_time/error.rs`, so this slice is likely overlapping and should not compete for payout.

PR #2655, opened 2026-07-16
refactor: make Scalar VarInt implementation blanket (#228, third item)

### LINKED ISSUE #228: Make opaque `Scalar` conversions explicit rather than using `From`/`Into` trait.

01luyicheng:
/claim #228

## Summary

This PR implements the **third** checklist item of #228: removing the `VarInt` supertrait bound from `Scalar` and replacing the `MontScalar<T>`-specific `VarInt` implementation with a blanket `impl<S: Scalar> VarInt for S`.

Per the issue, the blanket implementation "relies on the limb conversions only" (`RefInto<[u64; 4]>` and `From<[u64; 4]>`). The `ZigZag` impls and scalar varint helpers are generalized from `MontScalar<T>` to `S: Scalar`, routing all conversions through the existing limb-based trait bounds.

This PR is intentionally scoped to the third checklist item only. It does **not** touch the limb-conversion bounds (first item) or the string-hash conversion (second item).

## Changes

- **`scalar.rs`**: Remove `+ VarInt` from the `Scalar` supertrait bounds and the unused `VarInt` import.
- **`varint_trait.rs`**: Replace `impl<T: MontConfig<4>> VarInt for MontScalar<T>` with a blanket `impl<S: Scalar> VarInt for S`.
- **`zigzag.rs`**: Generalize `ZigZag<U256> for MontScalar<T>` and `ZigZag<MontScalar<T>> for U256` to `ZigZag<U256> for S` and `ZigZag<S> for U256` where `S: Scalar`. Conversions now use `self.ref_into()` + `U256::from_limbs(...)` and `S::from(zig_val.to_limbs())`, relying on the existing limb-conversion bounds only.
- **`scalar_varint.rs`**: Generalize helper function signatures from `<T: MontConfig<4>>` / `MontScalar<T>` to `<S: Scalar>` / `S`.
- **`u256.rs`**: Add `U256::from_limbs([u64; 4])` and `U256::to_limbs() -> [u64; 4]` inherent methods to bridge between the limb representation and `U256` without relying on `MontScalar`-specific `From` impls.
- **`varint_trait_test.rs`**: Add a compile-time test verifying `TestScalar` receives `VarInt` via the blanket impl, and a roundtrip test for the new `U256` limb helpers.

## Semantic equivalence

The previous `ZigZag<S> for U256` impl used `From<&U256> for MontScalar<T>` (`from_le_bytes_mod_order`) to convert back to a scalar. The new impl uses `S::from(zig_val.to_limbs())` (`From<[u64; 4]>`, i.e. `Fp::new(BigInt(value))`). Both perform `value mod p` reduction and produce identical results. This is verified by the existing test `valid_varint_encoded_input_that_map_to_test_scalars_bigger_than_the_p_field_order_in_the_read_scalar_will_wrap_around_p` which constructs values larger than `p` and checks the wrapping behavior.

## Testing

- `cargo fmt --check` — clean
- `cargo check -p proof-of-sql --no-default-features` — passed
- `cargo check -p proof-of-sql --no-default-features --features="test"` — passed
- `cargo check -p proof-of-sql --no-default-features --features="arrow"` — passed
- `cargo test -p proof-of-sql --no-default-features --features="test" --lib base::encode` — 53 tests passed
- `cargo test -p proof-of-sql --no-default-features --features="test" --lib base::scalar` — 12 tests passed
- `cargo clippy` — no new warnings in modified files

## Notes

- The existing `From<&MontScalar<T>> for U256` and `From<&U256> for MontScalar<T>` impls in `u256.rs` are retained because they are still used by test code. They can be cleaned up in a follow-up PR once all three subtasks of #228 are addressed.
- This PR does not use `closes #228` since it only addresses the third of three checklist items.

Refs #228

### DIFF
--- a/crates/proof-of-sql/src/base/encode/u256.rs
+++ b/crates/proof-of-sql/src/base/encode/u256.rs
@@ -16,6 +16,26 @@ impl U256 {
     pub const fn from_words(low: u128, high: u128) -> Self {
         U256 { low, high }
     }
+
+    /// Creates a `U256` from little-endian 64-bit limbs.
+    #[inline]
+    pub const fn from_limbs(limbs: [u64; 4]) -> Self {
+        let low = limbs[0] as u128 | ((limbs[1] as u128) << 64);
+        let high = limbs[2] as u128 | ((limbs[3] as u128) << 64);
+        U256 { low, high }
+    }
+
+    /// Converts this `U256` into little-endian 64-bit limbs.
+    #[expect(clippy::cast_possible_truncation)]
+    #[inline]
+    pub const fn to_limbs(self) -> [u64; 4] {
+        [
+            self.low as u64,
+            (self.low >> 64) as u64,
+            self.high as u64,
+            (self.high >> 64) as u64,
+        ]
+    }
 }
 
 /// This trait converts a dalek scalar into a U256 integer
--- a/crates/proof-of-sql/src/base/encode/u256.rs
+++ b/crates/proof-of-sql/src/base/encode/u256.rs
@@ -41,17 +41,13 @@ impl U256 {
 /// This trait converts a dalek scalar into a U256 integer
 impl<T: MontConfig<4>> From<&MontScalar<T>> for U256 {
     fn from(val: &MontScalar<T>) -> Self {
-        let buf: [u64; 4] = val.into();
-        let low: u128 = u128::from(buf[0]) | (u128::from(buf[1]) << 64);
-        let high: u128 = u128::from(buf[2]) | (u128::from(buf[3]) << 64);
-        U256::from_words(low, high)
+        U256::from_limbs(val.into())
     }
 }
 
 /// This trait converts a U256 integer into a dalek scalar
 impl<T: MontConfig<4>> From<&U256> for MontScalar<T> {
     fn from(val: &U256) -> Self {
-        let bytes = [val.low.to_le_bytes(), val.high.to_le_bytes()].concat();
-        MontScalar::<T>::from_le_bytes_mod_order(&bytes)
+        MontScalar::<T>::from(val.to_limbs())
     }
 }
--- a/crates/proof-of-sql/src/base/encode/scalar_varint.rs
+++ b/crates/proof-of-sql/src/base/encode/scalar_varint.rs
@@ -1,8 +1,7 @@
 use crate::base::{
     encode::{ZigZag, U256},
-    scalar::MontScalar,
+    scalar::Scalar,
 };
-use ark_ff::MontConfig;
 use core::cmp::{max, Ordering};
 
 /// This function writes the input scalar x as a varint encoding to buf slice
@@ -14,7 +13,7 @@ use core::cmp::{max, Ordering};
 ///
 /// crash:
 /// - in case N is bigger than `buf.len()`
-pub fn write_scalar_varint<T: MontConfig<4>>(buf: &mut [u8], x: &MontScalar<T>) -> usize {
+pub fn write_scalar_varint<S: Scalar>(buf: &mut [u8], x: &S) -> usize {
     write_u256_varint(buf, x.zigzag())
 }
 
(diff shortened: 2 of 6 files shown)


crates/proof-of-sql/src/base/encode/u256.rs
```diff
@@ -16,6 +16,26 @@ impl U256 {
...
+    #[expect(clippy::cast_possible_truncation)]
+    #[inline]
+    pub const fn to_limbs(self) -> [u64; 4] {
+        [
+            self.low as u64,
+            (self.low >> 64) as u64,
+            self.high as u64,
+            (self.high >> 64) as u64,
+        ]
+    }
```
JayWhite2357, line 38: This is duplicate code.
Instead, either remove the `impl<T: MontConfig<4>> From<&MontScalar<T>> for U256` and `impl<T: MontConfig<4>> From<&U256> for MontScalar<T> ` blocks or reuse the duplicate code if needed.

JayWhite2357 (changes requested):

crates/proof-of-sql/src/base/encode/u256.rs
01luyicheng: Thanks for the catch, @JayWhite2357! You're right — the new `U256::from_limbs` / `U256::to_limbs` helpers duplicated the bodies of the two pre-existing `From` impls right below them.

I went with your "reuse the duplicate code" option: both `From` impls now delegate to the new helpers, so the limb-assembly / limb-split logic lives in exactly one place.

```rust
impl<T: MontConfig<4>> From<&MontScalar<T>> for U256 {
    fn from(val: &MontScalar<T>) -> Self {
        U256::from_limbs(val.into())
    }
}

impl<T: MontConfig<4>> From<&U256> for MontScalar<T> {
    fn from(val: &U256) -> Self {
        MontScalar::<T>::from(val.to_limbs())
    }
}
```

A few notes on the choice:

- I kept the `From` impls (rather than removing them outright) because `From<&U256> for MontScalar<T>` is still used by the test fixtures in `scalar_varint_test.rs` and `zigzag_test.rs` (the `let val: TestScalar = (&U256::from_words(...)).into();` pattern). Removing them would have churned 11 test call sites without changing any behavior. Happy to do that cleanup too if you'd prefer full removal.
- The reverse-direction impl previously used `from_le_bytes_mod_order`; the delegated form goes through `From<[u64; 4]>` (`Fp::new(BigInt(value))`). Both compute `value mod p`, and this is the same path the PR already routed `ZigZag<S> for U256` through in `zigzag.rs`. The existing `..._will_wrap_around_p` test covers the >p case.

Verification (all run locally on the new commit `1c933a6c`, parent `df7492af`):

- `cargo fmt --check` — clean (exit 0, no diff).
- `cargo check -p proof-of-sql --no-default-features` — exit 0 (only pre-existing dead-code warnings).
- `cargo check -p proof-of-sql --no-default-features --features=test` — exit 0.
- `cargo check -p proof-of-sql --no-default-features --features=arrow` — exit 0.
- `cargo test -p proof-of-sql --no-default-features --features=test --lib base::encode` — **53 passed; 0 failed; 0 ignored; 0 measured; 910 filtered out**. Includes `u256_limb_roundtrips_preserve_values`, all `scalar_varint_test::*`, and all `zigzag_test::*` cases that exercise the changed `From` impls.
- `cargo test -p proof-of-sql --no-default-features --features=test --lib base::scalar` — **12 passed; 0 failed; 0 ignored; 0 measured; 951 filtered out**. Includes all five `we_can_convert_*_with_wrapping*` tests (`...to_u256_with_wrapping`, `...to_u256_with_wrapping_of_negative_value`, `...u256_to_test_scalar_with_wrapping`, `...u256_to_test_scalar_with_wrapping_of_large_value`, `...u256_to_test_scalar_with_wrapping_of_random_u256`) that pin the `value mod p` behavior.
- `cargo clippy -p proof-of-sql --no-default-features --features=test -- -D warnings` — exits non-zero, but **all** flagged lints are pre-existing and in unrelated files (`dory/`, `dynamic_matrix_utils/`, `hyperkzg/`, `data_accessor_impl.rs`). I confirmed this by running clippy on the parent commit `df7492af` directly: the set of flagged files is byte-for-byte identical, and `u256.rs` does not appear in either run. No new warnings from this change.

Net diff: 1 file changed, 2 insertions(+), 6 deletions(-) in `crates/proof-of-sql/src/base/encode/u256.rs`.

JayWhite2357:
Closing all issues and external PRs due to suspected low-effort AI slop/spam. If you would like review on this, feel free to repoen.

PR #2659, opened 2026-07-22
feat: support `CASE WHEN` via a native CaseExpr proof primitive

rishitesh-snt:
# Rationale for this change

`CASE WHEN` wasn't supported by the planner. This adds it as a native proof
primitive, `CaseExpr`, rather than desugaring it into arithmetic. Because the
result is one of the branch values (not computed from them), it keeps the branch
type with no widening or overflow, and varchar branches work.

- `CaseExpr`: proves `CASE WHEN c_1 THEN v_1 ... ELSE v_else END` with
  first-match-wins. The prover commits a cheap boolean guard column per arm
  (after the first) and the result column, and proves two degree-2 identities:
  each guard is `c_i * (1 - sum of earlier guards)`, and the result is the
  guarded sum. Guards are 0/1 so only the single result is a full data column,
  regardless of arm count.
- `select_column`: a shared row-wise column selection helper (used by `CaseExpr`,
  reusable elsewhere) with its own unit tests.
- Planner lowers searched and simple CASE forms; CASE without ELSE is rejected
  (implicit NULL has no representation here).
- The new `DynProofExpr::Case` variant is appended last to keep existing
  serialized plans stable. The EVM serializer rejects it until the Solidity
  verifier has a matching routine.

# Are these changes tested?

Yes. End-to-end prove/verify for numeric, multi-arm (up to 4 conditions),
varchar, nested, and conditional-aggregation CASE, plus planning-time rejection
tests. `case_expr_test.rs` adds `try_new` validation tests and prove/verify
through `exercise_verification`, which tampers the result and proof to confirm
rejection. Note: `case_expr_test.rs` is blitzar-gated (like the sibling expr
tests) and runs in CI, not on Apple Silicon locally.

### DIFF
--- a/crates/proof-of-sql/src/base/database/column_select_operation.rs
+++ b/crates/proof-of-sql/src/base/database/column_select_operation.rs
@@ -1,0 +1,185 @@
+use super::{Column, ColumnType};
+use crate::base::scalar::Scalar;
+use alloc::vec::Vec;
+use bumpalo::Bump;
+
+/// View each source as a typed slice via `accessor`, then build a new slice of
+/// length `num_rows` that takes, for each row `r`, `sources[source_of_row(r)]`'s
+/// value at `r`.
+///
+/// `source_of_row` is evaluated lazily per row, so the caller need not materialize
+/// a per-row index column.
+///
+/// # Panics
+/// Panics if `accessor` returns `None` for any source (i.e. the sources are not all
+/// the same type) or if any `source_of_row(r)` is out of bounds.
+fn select<'a, S, T>(
+    alloc: &'a Bump,
+    sources: &[Column<'a, S>],
+    num_rows: usize,
+    source_of_row: impl Fn(usize) -> usize,
+    accessor: impl Fn(&Column<'a, S>) -> Option<&'a [T]>,
+) -> &'a [T]
+where
+    S: Scalar,
+    T: Copy,
+{
+    let typed: Vec<&'a [T]> = sources
+        .iter()
+        .map(|source| accessor(source).expect("sources must share a column type"))
+        .collect();
+    alloc.alloc_slice_fill_with(num_rows, |row_num| typed[source_of_row(row_num)][row_num])
+}
+
+/// Same as [`select`], for the `(values, scalars)` pair layout of `VarChar`/`VarBinary`.
+///
+/// # Panics
+/// Panics under the same conditions as [`select`].
+fn select_pair<'a, S, T>(
+    alloc: &'a Bump,
+    sources: &[Column<'a, S>],
+    num_rows: usize,
+    source_of_row: impl Fn(usize) -> usize,
+    accessor: impl Fn(&Column<'a, S>) -> Option<(&'a [T], &'a [S])>,
+) -> (&'a [T], &'a [S])
+where
+    S: Scalar,
+    T: Copy,
+{
+    let typed: Vec<(&'a [T], &'a [S])> = sources
+        .iter()
+        .map(|source| accessor(source).expect("sources must share a column type"))
+        .collect();
+    (
+        alloc.alloc_slice_fill_with(num_rows, |row_num| typed[source_of_row(row_num)].0[row_num]),
+        alloc.alloc_slice_fill_with(num_rows, |row_num| typed[source_of_row(row_num)].1[row_num]),
+    )
+}
+
+/// Build a column by taking, for each row `r`, the value at row `r` of
+/// `sources[row_source[r]]`.
+///
+/// This is the row-wise selection underlying `CASE` (and any other conditional
+/// pick): `source_of_row(r)` names which source column supplies row `r`. It is
+/// evaluated lazily per row, so no per-row index column is materialized.
+///
+/// All sources must share the column type, and the result has that type. Because
+/// the prover copies real values, every column type is supported, including
+/// `VarChar`/`VarBinary`.
+///
+/// # Panics
+/// Panics if the sources are not all of the same type, if `sources` is empty, or
+/// if any `source_of_row(r)` is out of bounds. Callers are expected to guarantee
+/// these (e.g. a `CaseExpr` validated at construction).
+pub(crate) fn select_column<'a, S: Scalar>(
+    alloc: &'a Bump,
+    sources: &[Column<'a, S>],
+    num_rows: usize,
+    source_of_row: impl Fn(usize) -> usize,
+) -> Column<'a, S> {
+    let column_type = sources
+        .first()
+        .expect("select_column requires at least one source")
+        .column_type();
+    let pick = &source_of_row;
+
+    match column_type {
+        ColumnType::Boolean => {
+            Column::Boolean(select(alloc, sources, num_rows, pick, Column::as_boolean))
+        }
+        ColumnType::Uint8 => {
+            Column::Uint8(select(alloc, sources, num_rows, pick, Column::as_uint8))
+        }
+        ColumnType::TinyInt => {
+            Column::TinyInt(select(alloc, sources, num_rows, pick, Column::as_tinyint))
+        }
+        ColumnType::SmallInt => {
+            Column::SmallInt(select(alloc, sources, num_rows, pick, Column::as_smallint))
+        }
+        ColumnType::Int => Column::Int(select(alloc, sources, num_rows, pick, Column::as_int)),
+        ColumnType::BigInt => {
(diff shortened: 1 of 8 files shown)


crates/proof-of-sql-planner/src/expr.rs
JayWhite2357, line 1: Let's put the planner changes in a second PR.

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
```diff
@@ -0,0 +1,284 @@
...
+    ///
+    /// Every condition must be boolean, every branch value (including ELSE) must
+    /// have exactly the same type, and there must be at least one WHEN arm.
+    pub fn try_new(
+        when_thens: Vec<(DynProofExpr, DynProofExpr)>,
+        else_expr: Box<DynProofExpr>,
+    ) -> AnalyzeResult<Self> {
+        let result_type = else_expr.data_type();
+        if when_thens.is_empty() {
+            return Err(AnalyzeError::InvalidDataType {
```
JayWhite2357: Having no `WHEN` arms is not an invalid data type error.

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// Create a CASE expression.
+    ///
+    /// Every condition must be boolean, every branch value (including ELSE) must
+    /// have exactly the same type, and there must be at least one WHEN arm.
+    pub fn try_new(
+        when_thens: Vec<(DynProofExpr, DynProofExpr)>,
+        else_expr: Box<DynProofExpr>,
+    ) -> AnalyzeResult<Self> {
+        let result_type = else_expr.data_type();
+        if when_thens.is_empty() {
```
JayWhite2357: I don't see any reason why the CaseExpr should prevent having no `WHEN` arms. It is a bit suboptimal, but that choice should be the decision of the planner, not the plan.

A decent chunk of the code you have assumes there is a "first" arm. But, you could rewrite it to not make that assumption, and the code would simplify a decent bit.

crates/proof-of-sql/src/base/database/column_select_operation.rs
```diff
@@ -0,0 +1,132 @@
...
+    sources: &[Column<'a, S>],
+    row_source: &[usize],
+) -> Column<'a, S> {
+    let column_type = sources
+        .first()
+        .expect("select_column requires at least one source")
+        .column_type();
+
+    /// Gather one typed slice per source, then pick per row.
+    macro_rules! select {
```
JayWhite2357: Avoid macros. There is nearly always a way to write code without them.

From the [Rust book](https://doc.rust-lang.org/book/ch20-05-macros.html)

>The downside... is that macro definitions are more complex than function... macro definitions are generally more difficult to read, understand, and maintain than function definitions.

In this case, you could convert this to a function that accepts `sources` and an `accessor` as parameters. (Although, I bet there's an even better way.)

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
```diff
@@ -0,0 +1,284 @@
...
+        // Sources for selection: one per arm, then the else branch last.
+        let mut sources = Vec::with_capacity(self.when_thens.len() + 1);
+        for (condition, value) in &self.when_thens {
+            condition_columns.push(condition.final_round_evaluate(builder, alloc, table, params)?);
+            sources.push(value.final_round_evaluate(builder, alloc, table, params)?);
+        }
+        sources.push(
+            self.else_expr
+                .final_round_evaluate(builder, alloc, table, params)?,
+        );
```
JayWhite2357: Prefer iterators over `push`.

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
```diff
@@ -0,0 +1,284 @@
...
+
+        let conditions: Vec<&[bool]> = condition_columns
+            .iter()
+            .map(|column| column.as_boolean().expect("condition is not boolean"))
+            .collect();
+        let else_source = self.when_thens.len();
+        let row_source = row_sources(&conditions, table.num_rows(), else_source);
```
JayWhite2357: `row_source` is an intermediate temporary column that takes up memory and shouldn't be needed.

JayWhite2357 (changes requested):

crates/proof-of-sql-planner/src/expr.rs
rishitesh-snt: Ok

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
rishitesh-snt: Changed the code

crates/proof-of-sql/src/base/database/column_select_operation.rs
rishitesh-snt: Removed the macros and replaced with functions.

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
rishitesh-snt: Changed the code

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
rishitesh-snt: Changed the code

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
rishitesh-snt: Changed the code.

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
```diff
@@ -0,0 +1,297 @@
...
+            })
+            .collect::<PlaceholderResult<Vec<_>>>()?
+            .into_iter()
+            .unzip();
+        // The else branch is the last source.
+        sources.push(self.else_expr.first_round_evaluate(alloc, table, params)?);
+        let else_source = self.when_thens.len();
+        Ok(select_column(alloc, &sources, table.num_rows(), |row| {
+            winning_source(&conditions, else_source, row)
+        }))
```
JayWhite2357, line 128: I feel like the optimal signature for the helper functions looks like this:


```suggestion
        let else_source = self.else_expr.first_round_evaluate(alloc, table, params)?;
        Ok(select_column(alloc, &conditions, &sources, &else_source))
```

You know that `table.num_rows()` is just `else_source.len()`.

As it is written, there are 3 functions: `select_column`, `winning_source`, and the closure.
I feel like all of those make it hard to read.

An added benefit is that we don't need `mut` on sources.

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
```diff
@@ -0,0 +1,297 @@
...
+                ))
+            })
+            .collect::<PlaceholderResult<Vec<_>>>()?
+            .into_iter()
+            .unzip();
+        // The else branch is the last source.
+        sources.push(
+            self.else_expr
+                .final_round_evaluate(builder, alloc, table, params)?,
+        );
```
JayWhite2357, line 159: I think the else is going to need to be handled specially, so it doesn't really make sense to push it onto the sources. This has the added benefit of not making sources mutable.

crates/proof-of-sql/src/sql/proof_exprs/case_expr.rs
```diff
@@ -0,0 +1,297 @@
...
+        let else_eval = self
+            .else_expr
+            .verifier_evaluate(builder, accessor, chi_eval, params)?;
+
+        // Mirror the prover's guard identities: each guard is read from the proof and
+        // checked against g_i - c_i + c_i*(sum earlier) = 0. The bare -c_i term needs
+        // no chi factor: every column is already 0 outside the table rows. `guard_sum`
+        // accumulates g_0 + ... + g_{i-1} as we go.
+        let mut guard_evals = Vec::with_capacity(condition_evals.len());
+        let mut guard_sum = S::ZERO;
```
JayWhite2357, line 259: Avoid `mut`. You should be able to use some iterator tools like `repeat` and `sum` to make this a bit cleaner.

JayWhite2357 (changes requested):

PR #2662, opened 2026-07-26
refactor!: make string to `Scalar` conversions explicit

vietlinhh02:
# Make string → `Scalar` conversions explicit

Closes the second checklist item of #228.

## What changed

`Scalar` no longer requires `From<&str>`, `From<String>`, or `From<&String>`. Those three bounds are replaced by a single trait method with a default implementation:

```rust
fn from_str_via_hash(val: &str) -> Self {
    <Self as ScalarExt>::from_byte_slice_via_hash(val.as_bytes())
}
```

Strings have no natural embedding into the field — they are hashed — so a silent `.into()` at the call site hid a Keccak invocation behind syntax that reads like a numeric conversion. Every generic call site now names the hash explicitly.

## Why the default implementation delegates

All three removed conversions already funnelled into the same place: `From<&str>`/`From<String>` called `x.as_bytes().into()`, which hit `From<&[u8]>`, which called `ScalarExt::from_byte_slice_via_hash`. `From<&String>` reached `From<String>` through the blanket `impl<F, T> From<&T> for MontScalar<F>`. The default implementation therefore delegates rather than re-deriving the hash, so the masking and limb-conversion logic stays in exactly one place.

**This refactor is behaviour-preserving at every call site** — the bytes hashed and the resulting field element are unchanged.

## Call sites updated

| File | Was |
| --- | --- |
| `base/arrow/arrow_array_to_column_conversion.rs` | `vals[i].into()` |
| `base/commitment/committable_column.rs` | `.map(Into::<S>::into)` |
| `base/database/column.rs` (×2) | `S::from(string)`, `.map(S::from)` |
| `base/database/literal_value.rs` | `str.into()` |
| `base/database/owned_column.rs` | `inner_product_ref_cast` |
| `base/database/owned_table_test_accessor.rs` | `(*s).into()` |
| `base/database/table_utility.rs` | `(*s).into()` |
| `sql/proof/provable_query_result.rs` | `decode_and_convert::<&str, S>` |
| `proof-of-sql-benches/src/utils/random_util.rs` (×3) | `Into::into(s)` |

Two of these follow patterns the codebase had already established for `VarBinary`, which does not have a natural embedding either:

- **`inner_product_with_strings`** mirrors the existing `inner_product_with_bytes`, including its comment (*"Cannot use blanket impls for `Vec<u8>` because bytes might have different embeddings as scalars"*) — the same reasoning applies verbatim to `String`.
- **`provable_query_result.rs`** now decodes `&str` and hashes it in a second step, exactly as the neighbouring `VarBinary` arm already decoded `&[u8]` and called `from_byte_slice_via_hash`.

`literal_value.rs::to_scalar` is a good illustration of the inconsistency this removes — `VarBinary` was already explicit while `VarChar` right above it was not:

```rust
 Self::VarChar(str) => S::from_str_via_hash(str),
 Self::VarBinary(bytes) => S::from_byte_slice_via_hash(bytes),
```

## Tests

Three tests added next to the existing hashing tests in `scalar_ext.rs`:

- `we_can_get_scalar_from_hashed_string` — agreement with `from_byte_slice_via_hash` on the same bytes, i.e. the delegation contract.
- `we_can_get_scalar_from_hashed_multibyte_string` — multi-byte UTF-8 hashes its encoded bytes.
- `we_can_get_zero_from_empty_string` — the empty string maps to `ZERO`, matching the empty byte slice.

I deliberately did not add a "different strings give different scalars" test, since `curve_25519_tests::two_different_strings_map_to_different_scalars` already covers that property.

## Verification

```
cargo check --workspace --all-features      # clean, no warnings
cargo test --all-features                   # CI's test command
cargo clippy --all-targets --all-features
cargo fmt --all --check
```

## One open question

I limited this to the trait bounds, which is what the checklist item asks for. The concrete `impl From<&str> for MontScalar` / `impl From<String> for MontScalar` still exist, so `MontScalar::from("…")` continues to compile even though generic code can no longer convert opaquely. Removing those two impls as well would finish the job for concrete types, at the cost of touching a fair number of call sites in tests. Happy to do that here or in a follow-up — whichever you prefer.

/claim #228

### DIFF
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -162,7 +162,7 @@ impl<'a, S: Scalar> From<&'a OwnedColumn<S>> for CommittableColumn<'a> {
             OwnedColumn::VarChar(strings) => CommittableColumn::VarChar(
                 strings
                     .iter()
-                    .map(Into::<S>::into)
+                    .map(|s| S::from_str_via_hash(s))
                     .map(Into::<[u64; 4]>::into)
                     .collect(),
             ),
--- a/crates/proof-of-sql/src/base/scalar/scalar.rs
+++ b/crates/proof-of-sql/src/base/scalar/scalar.rs
@@ -1,7 +1,10 @@
 #![expect(clippy::module_inception)]
 
-use crate::base::{encode::VarInt, ref_into::RefInto, scalar::ScalarConversionError};
-use alloc::string::String;
+use crate::base::{
+    encode::VarInt,
+    ref_into::RefInto,
+    scalar::{ScalarConversionError, ScalarExt},
+};
 use bnum::types::U256;
 use core::ops::Sub;
 use num_bigint::BigInt;
@@ -13,7 +16,6 @@ pub trait Scalar:
     + core::fmt::Display
     + PartialEq
     + Default
-    + for<'a> From<&'a str>
     + Sync
     + Send
     + num_traits::One
@@ -52,9 +54,7 @@ pub trait Scalar:
     + num_traits::Inv<Output = Option<Self>> // Note: `inv` should return `None` exactly when the element is zero.
     + core::ops::SubAssign
     + RefInto<[u64; 4]>
-    + for<'a> core::convert::From<&'a String>
     + VarInt
-    + core::convert::From<String>
     + core::convert::From<i128>
     + core::convert::From<i64>
     + core::convert::From<i32>
@@ -85,4 +85,18 @@ pub trait Scalar:
     const MAX_BITS: u8;
     /// A U256 representation of the largest signed value in the field.
     const MAX_SIGNED_U256: U256;
+
+    /// Converts a string to a `Scalar` by hashing its bytes.
+    ///
+    /// This is not a parsing method: the resulting field element bears no numeric
+    /// relationship to the contents of `val`. Strings have no natural embedding into
+    /// the field, so they are hashed down into the field's range and the resulting
+    /// bits are then converted through the limb representation.
+    ///
+    /// The empty string is the one exception: it maps to [`Scalar::ZERO`] directly,
+    /// without hashing.
+    #[must_use]
+    fn from_str_via_hash(val: &str) -> Self {
+        <Self as ScalarExt>::from_byte_slice_via_hash(val.as_bytes())
+    }
 }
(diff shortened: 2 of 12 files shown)


crates/proof-of-sql/src/base/scalar/scalar.rs
```diff
@@ -85,4 +85,18 @@ pub trait Scalar:
...
+
+    /// Converts a string to a `Scalar` by hashing its bytes.
+    ///
+    /// This is not a parsing method: the resulting field element bears no numeric
+    /// relationship to the contents of `val`. Strings have no natural embedding into
+    /// the field, so they are hashed down into the field's range and the resulting
+    /// bits are then converted through the limb representation.
+    ///
+    /// The empty string is the one exception: it maps to [`Scalar::ZERO`] directly,
+    /// without hashing.
```
JayWhite2357, line 97: Do not add superfluous details.

crates/proof-of-sql/src/base/commitment/committable_column.rs
```diff
@@ -162,7 +162,7 @@ impl<'a, S: Scalar> From<&'a OwnedColumn<S>> for CommittableColumn<'a> {
             OwnedColumn::VarChar(strings) => CommittableColumn::VarChar(
                 strings
                     .iter()
-                    .map(Into::<S>::into)
+                    .map(|s| S::from_str_via_hash(s))
```
JayWhite2357, line 165: I believe that you can do this:
```suggestion
                    .map(S::from_str_via_hash)
```
Other places in this PR can do the same thing, I believe.

crates/proof-of-sql/src/base/scalar/scalar.rs
```diff
@@ -85,4 +85,18 @@ pub trait Scalar:
...
+    #[must_use]
+    fn from_str_via_hash(val: &str) -> Self {
+        <Self as ScalarExt>::from_byte_slice_via_hash(val.as_bytes())
```
JayWhite2357, line 100: I believe that this causes the PR to be a breaking change because it changes the encoding used. Please ensure that the actual functionality doesn't change. This should be a pure refactor, and the only breaking change should be the API change to `Scalar`, but not the actual resulting commitments. Scalar.

JayWhite2357 (changes requested):

JayWhite2357:
Closing to clean up the PRs. Feel free to reopen if the problem are fixed.

PR #2663, opened 2026-08-06
refactor: replace opaque Scalar string conversions with from_str_via_hash

antheducation:
Closes the string-conversion checkbox of #228.

/claim #228

## What and why

`Scalar` required `for<'a> From<&'a str>`, `for<'a> From<&'a String>` and `From<String>`.
These are exactly the "opaque" conversions #228 is about: a string has **no natural
embedding** into the field, so the conversion hashes the UTF-8 bytes and masks the result.
Exposing that as `From` made a lossy, collision-resistant hash look like an ordinary
numeric conversion at every generic call site — `str.into()` reads like `1i64.into()` but
does something completely different.

Per the rule stated in the issue — `From` only for types with a natural embedding,
a trait method otherwise — this replaces the three bounds with an explicit method.

## Approach

`ScalarExt` already exists for precisely this ("blanket implementations for `Scalar` types …
to avoid cluttering the core `Scalar` implementation with default implementations") and
already houses the byte-slice sibling, `from_byte_slice_via_hash`. So:

- Added `ScalarExt::from_str_via_hash(val: &str) -> Self` with a **default implementation**
  delegating to `from_byte_slice_via_hash`. Since `impl<S: Scalar> ScalarExt for S {}` is a
  blanket impl, every `Scalar` gets it for free and no implementor has to write anything.
- Removed the three bounds from `Scalar` (and the now-unused `alloc::string::String` import).
- Updated the call sites that relied on those bounds to call the method explicitly. Each one
  now reads the same way as the `VarBinary` arm sitting next to it, which already used
  `from_byte_slice_via_hash`. For example, in `LiteralValue::to_scalar`:

  ```rust
  Self::VarChar(str)     => S::from_str_via_hash(str),
  Self::VarBinary(bytes) => S::from_byte_slice_via_hash(bytes),
  ```

- Added `slice_ops::inner_product_with_strings`, mirroring the existing
  `inner_product_with_bytes`, for the one site that needed it. Its doc comment gives the same
  rationale the bytes version already gives.

## Scope note (deliberate, happy to change)

The concrete `impl From<&str> for MontScalar` / `From<String> for MontScalar` impls are
**left in place**. Removing the *bounds* is what removes the opacity from all generic
`S: Scalar` code, which is the actual complaint. Removing the concrete impls as well would
touch **210+ call sites** (overwhelmingly `TestScalar::from("Cat")` in tests) and would bury
this change in churn. Happy to do that as a follow-up if you'd like it.

Also scoped to one checkbox, since the issue says each should be a separate PR. The limbs
and `VarInt` checkboxes are untouched here.

## Tests

Added five tests to `scalar_ext`:

- empty string hashes to `ZERO` (consistent with the empty byte slice)
- `from_str_via_hash("abc")` equals `from_byte_slice_via_hash(b"abc")`
- different strings produce different scalars
- multibyte/non-ASCII input hashes its UTF-8 encoding
- `from_str_via_hash("1") != ONE` — this is a hash, not a parse

## Verification

Run locally on Windows. `blitzar` is Linux-only, so I used the same feature set as your
`--no-default-features --features="arrow"` CI job:

- `cargo test -p proof-of-sql --no-default-features --features="arrow" --lib`
  → **1061 passed; 0 failed**
- `cargo fmt --all -- --check --config imports_granularity=Crate,group_imports=One` → clean
- `cargo clippy -p proof-of-sql --all-targets --no-default-features --features="arrow"`
  → no findings in any file this PR touches

I could not build the `blitzar`/`perf` feature paths locally, so those are relying on CI.

### DIFF
--- a/crates/proof-of-sql/src/base/arrow/arrow_array_to_column_conversion.rs
+++ b/crates/proof-of-sql/src/base/arrow/arrow_array_to_column_conversion.rs
@@ -279,7 +279,9 @@ impl ArrayRefExt for ArrayRef {
                     let scals = if let Some(scals) = precomputed_scals {
                         &scals[range.start..range.end]
                     } else {
-                        alloc.alloc_slice_fill_with(vals.len(), |i| -> S { vals[i].into() })
+                        alloc.alloc_slice_fill_with(vals.len(), |i| -> S {
+                            S::from_str_via_hash(vals[i])
+                        })
                     };
 
                     Ok(Column::VarChar((vals, scals)))
--- a/crates/proof-of-sql/src/base/commitment/committable_column.rs
+++ b/crates/proof-of-sql/src/base/commitment/committable_column.rs
@@ -162,7 +162,7 @@ impl<'a, S: Scalar> From<&'a OwnedColumn<S>> for CommittableColumn<'a> {
             OwnedColumn::VarChar(strings) => CommittableColumn::VarChar(
                 strings
                     .iter()
-                    .map(Into::<S>::into)
+                    .map(|s| S::from_str_via_hash(s))
                     .map(Into::<[u64; 4]>::into)
                     .collect(),
             ),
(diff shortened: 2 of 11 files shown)


JayWhite2357 (changes requested):
This is nearly identical to #2662.
See the comments on it.

JayWhite2357:
Closing to clean up the PRs. Feel free to reopen if the problems are fixed.

